PR #6, opened 2021-05-27
Rely on restricted node engine version to choose between esm or cjs

KnightNiwrem:


KnorpelSenf:
Tested with Node 12.20 and 14.13 and works. LGTM, thank you so much for investigating and fixing this!

@all-contributors add @KnightNiwrem for coding and finding the bug and also for the throttler plugin

PR #8, opened 2021-05-29
Retry `getUpdates` calls indefinitely when polling

KnorpelSenf:
Does not stop polling after one hour.

Also now retries upon all errors, not just API errors.

### DIFF
--- a/deno/src/bot.ts
+++ b/deno/src/bot.ts
@@ -276,7 +276,6 @@ export class Bot<C extends Context = Context> extends Composer<C> {
             // fetch updates
             const offset = this.lastTriedUpdateId + 1
             let updates: Update[] | undefined = undefined
-            let maxRetries = 1200 // 1 hour of continuously failing requests
             do {
                 try {
                     updates = await this.api.getUpdates(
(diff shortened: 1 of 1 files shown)


deno/src/bot.ts
```diff
@@ -276,28 +276,23 @@ export class Bot<C extends Context = Context> extends Composer<C> {
...
                     } else {
                         throw error
                     }
                 }
-            } while (
-                updates === undefined &&
-                this.pollingRunning &&
-                maxRetries-- > 0
-            )
+            } while (updates === undefined && this.pollingRunning)
```
wojpawlik, line 295: This condition could be further simplified to
```suggestion
            } while (this.pollingRunning)
```

if you add `break` right after `updates = await this.api.getUpdates(`

wojpawlik (approved):

deno/src/bot.ts
KnorpelSenf: That true, I just try to avoid `break` and `continue` as much as possible as they're basically `goto` …

KnorpelSenf:
Thanks for the prompt review :)

PR #13, opened 2021-07-01
Improve error handling

KnorpelSenf:
Fixes a bug in error boundaries.

Makes grammY ready for error handling in TypeScript 4.4 where errors in `catch` blocks will be of type `unknown` by default.

Improves the default error message that is printed if `BotError` is logged instead of properly inspecting the thrown error object.

Adds an option to print sensitive information to the logs, called `sensitiveLogs`. This will make the API client include the bot token in error messages, but as an advantage, more detailed system information become available when network requests fail. This option will be recommended to set to `true` for production scale bots in the deployment checklist.

### DIFF
--- a/deno/src/bot.ts
+++ b/deno/src/bot.ts
@@ -299,7 +299,13 @@ export class Bot<C extends Context = Context> extends Composer<C> {
                 try {
                     await this.handleUpdate(update)
                 } catch (err) {
-                    await this.errorHandler(err)
+                    // should always be true
+                    if (err instanceof BotError) {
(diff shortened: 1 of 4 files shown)


Tecardo1:
Why should sensitive data logged in production bots? Doesn’t make sense to me

KnorpelSenf:
Because they often log sensitive data anyway. User data, in particular. On the contrary, in a dev setup, this is rarely the case.

Also confer this: https://github.com/tdlib/telegram-bot-api/issues/138

PR #16, opened 2021-07-04
Check for Deno api property not exposed by Deno deploy

jokasimr:
This might not be the most 'gentle' way to check for permissions. :D

### DIFF
--- a/deno/src/platform.ts
+++ b/deno/src/platform.ts
@@ -11,12 +11,12 @@ const isDeno = typeof Deno !== 'undefined'
 if (isDeno) {
     debug.useColors = () => !Deno.noColor
     const env = { name: 'env', variable: 'DEBUG' } as const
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
Does this even compile on Deno Deploy? I thought that `Deno.permissions` is not available?

jokasimr:
I think it compiles, `Deno.permissions` just evaluates to `undefined` in that runtime.

jokasimr:
Here is a small test referencing the Dino.permissions property in the deploy runtime: 
 https://github.com/jokasimr/sf2-collective/blob/bot/bot/test.js

It runs here:
https://large-chicken-55.deno.dev/

And compiles :)

KnorpelSenf:
Amazing, that is certainly a change to the service then. This makes fixing the problem much easier. Could you check out #17?

PR #17, opened 2021-07-04
Support Deno Deploy

### LINKED ISSUE #15: Compatibility with Deno deploy
Hi, cool project!

I tried using grammY in a [Deno deploy](https://deno.com/deploy/docs) function and got the error 
`TypeError: cannot read property 'query' of undefined grammy Deno` in[ this line](https://github.com/grammyjs/grammY/blob/97a79f85b2ebc2e9ec626655e7c6b717238df695/deno/src/platform.ts#L14).

The line is `let res = await Deno.permissions.query(env)` and as I understand it the problem is that [Deno deploy does not expose the `Deno.permissions` API](https://deno.com/deploy/docs/runtime-api).

Not sure if this is something worth/possible doing anything about, but it would be very nice to be able to use this framework for deploying telegram bots on Deno deploy.

Thank you for the framework.

KnorpelSenf:
Title says it all.

Fixes #15 
Closes #16

### DIFF
--- a/deno/src/platform.ts
+++ b/deno/src/platform.ts
@@ -13,7 +13,9 @@ if (isDeno) {
     try {
         const val = Deno.env.get('DEBUG')
         if (val) debug.enable(val)
-    } catch (error) {}
+    } catch {
+        // cannot access env var, treat as if it is not set
+    }
 }
 
 import { iter } from 'https://deno.land/std@0.97.0/io/mod.ts'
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
Actually we can just as well drop the interactivity. There is not much a point to it since `--prompt`.

wojpawlik (approved):

PR #19, opened 2021-07-12
Support custom API types

KnorpelSenf:
Just as the context type can be adjusted, this PR introduces a type parameter for the API type.

It is used only for `bot.api` because `ctx.api` can already be adjusted via the context type. Also, we do not want to make a mess by having the context type depend on the API type.

This PR is required for a future implementation of a hydration plugin.

This PR is required by the code in the files plugin repository: https://github.com/grammyjs/files

### DIFF
--- a/deno/src/context.ts
+++ b/deno/src/context.ts
@@ -1,5 +1,6 @@
 // deno-lint-ignore-file camelcase
-import { Api, Other } from './core/api.ts'
+import { Api, Other as OtherApi } from './core/api.ts'
+import { Methods, RawApi } from './core/client.ts'
 import {
     Chat,
     ChatPermissions,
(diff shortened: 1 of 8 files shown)


deno/src/context.ts
```diff
@@ -1,5 +1,6 @@
 // deno-lint-ignore-file camelcase
-import { Api, Other } from './core/api.ts'
+import { Api, Other as OtherApi } from './core/api.ts'
```
EdJoPaTo, line 2: Maybe Other should get a better name. Something more expressive. `OtherArguments` maybe?

EdJoPaTo (commented):
I think I have a basic idea about this change but probably not enough to provide helpful feedback.

KnightNiwrem (approved):
Only went through the diffs (so if there is something that needs to be changed that is not in the diff, I've probably missed those). Otherwise, LGTM.

No major issues with running it with the minimal scaling example either.

PR #20, opened 2021-07-12
Make filter queries differentiate between chats and channels

KnorpelSenf:
Filter queries differ now for `message` updates and `channel_post` updates. For instance, `new_chat_member` can never be received in channels. On the other hand, `channel_chat_created` can never be received for `message` updates. This PR reflects that.

For reviewing, please confirm that the respective message types are correctly matched with messages/channel posts.


AndreoliBR (approved):
Everything is working as it should.

PR #24, opened 2021-08-04
Changed defaultGetSessionKey

Tecardo1:
To include ctx.myChatMember.chat as id reference.


KnorpelSenf:
We are going to change ctx.chat to respect my_chat_member updates for the 1.3 release. This PR will then be superfluous.

If you want to contribute this feature, consider fixing the context instead.

PR #26, opened 2021-08-04
Extend from

Tecardo1:

PR #27, opened 2021-08-04
Update context.ts

Tecardo1:


Tecardo1:
commit was for odd reasons not possible via vscode. Needed in addition to https://github.com/grammyjs/grammY/pull/26

KnorpelSenf:
This is now supported as of 8247773bc72f3e7821cfa0b3f6f0d35963d9200d.

PR #28, opened 2021-08-15
Add support for Web File on Deno

KnorpelSenf:
Allows `File` objects to be passed to `InputFile` (on Deno only). The file will be streamed via `file.stream()`. The filename will be taken from the supplied file object. Retrying API calls is possible. This change is non-breaking.

@wojpawlik is this possible to port to Node, preferably without raising the minimum Node version?
If no, this will simply be merged, and Node users will not get support for web files.

### DIFF
--- a/deno/src/platform.ts
+++ b/deno/src/platform.ts
@@ -51,7 +51,7 @@ export class InputFile {
     private [internalInputFileData]: ConstructorParameters<typeof InputFile>[0]
     public get [inputFileData]() {
         const file = this[internalInputFileData]
-        return file instanceof File ? file.stream() : file
+        return file instanceof Blob ? file.stream() : file
     }
     /**
      * Optional name of the constructed `InputFile` instance.
@@ -73,7 +73,7 @@ export class InputFile {
             | Uint8Array
             | ReadableStream<Uint8Array>
             | AsyncIterable<Uint8Array>
-            | File,
+            | Blob,
         filename?: string
     ) {
         this[internalInputFileData] = file
--- a/deno/src/platform.ts
+++ b/deno/src/platform.ts
@@ -37,6 +37,8 @@ export const baseFetchConfig = {}
 // === InputFile handling and File augmenting
 // Accessor for file data in `InputFile` instances
 export const inputFileData = Symbol('InputFile data')
+// Internal data storage for data in `InputFile` instances
+const internalInputFileData = Symbol('Internal InputFile data')
 
 /**
  * An `InputFile` wraps a number of different sources for [sending
@@ -46,7 +48,11 @@ export const inputFileData = Symbol('InputFile data')
  * Reference](https://core.telegram.org/bots/api#inputfile).
  */
 export class InputFile {
-    public readonly [inputFileData]: ConstructorParameters<typeof InputFile>[0]
+    private [internalInputFileData]: ConstructorParameters<typeof InputFile>[0]
+    public get [inputFileData]() {
+        const file = this[internalInputFileData]
+        return file instanceof File ? file.stream() : file
+    }
     /**
      * Optional name of the constructed `InputFile` instance.
      *
@@ -66,12 +72,14 @@ export class InputFile {
             | string
             | Uint8Array
             | ReadableStream<Uint8Array>
-            | AsyncIterable<Uint8Array>,
+            | AsyncIterable<Uint8Array>
+            | File,
         filename?: string
     ) {
-        this[inputFileData] = file
-        if (filename === undefined && typeof file === 'string')
-            filename = basename(file)
+        this[internalInputFileData] = file
+        if (filename === undefined)
+            if (typeof file === 'string') filename = basename(file)
+            else if (file instanceof File) filename = file.name
         this.filename = filename
     }
 }


deno/src/platform.ts
```diff
@@ -46,7 +48,11 @@ export const inputFileData = Symbol('InputFile data')
  * Reference](https://core.telegram.org/bots/api#inputfile).
  */
 export class InputFile {
-    public readonly [inputFileData]: ConstructorParameters<typeof InputFile>[0]
+    private [internalInputFileData]: ConstructorParameters<typeof InputFile>[0]
+    public get [inputFileData]() {
+        const file = this[internalInputFileData]
+        return file instanceof File ? file.stream() : file
```
wojpawlik, line 54:
```suggestion
        return file instanceof Blob ? file.stream() : file
```

I feel this getter can be simplified, but I'm not yet sure how...

deno/src/platform.ts
```diff
@@ -66,12 +72,14 @@ export class InputFile {
             | string
             | Uint8Array
             | ReadableStream<Uint8Array>
```
wojpawlik, line 74: Nitpick: `ReadableStream<Uint8Array>` is `AsyncIterable<Uint8Array>`, so it doesn't need to be listed separately.

wojpawlik (commented):
Node.js support is tricky.

Node 16 and `fetch-blob@2` have `Blob`, but no `File`. And simple `instanceof Blob` won't catch all `Blob` implementations.

I was thinking in terms of allowing `File` instead of `InputFile`, and adding `name` getter and `stream` method to `InputFile` to make it `File`-like.

deno/src/platform.ts
KnorpelSenf: I don't see any possible simplification either, especially because narrowing the type does not work with bracket access. Do let me know if you find something :)

KnorpelSenf:
> I was thinking in terms of allowing `File` instead of `InputFile`, and adding `name` getter and `stream` method to `InputFile` to make it `File`-like.

`InputFile` objects are only used with grammY, so there's no need for them to be `File`-like.

Also, it is a goal to be explicit about the file data that are passed to grammY methods, which is achieved by always relying on the `InputFile` constructor and nothing else. Permitting `File` objects contradicts this.

KnorpelSenf:
If this PR can be merged as-is, it will land in the 1.3 release, which is imminent.

wojpawlik:
I'd wait with this. Deno has no `fileFromPath` yet to take advantage of this. And I'll soonish tackle similar problems in https://github.com/telegraf/client and share my thoughts here.

deno/src/platform.ts
```diff
@@ -46,7 +48,11 @@ export const inputFileData = Symbol('InputFile data')
  * Reference](https://core.telegram.org/bots/api#inputfile).
  */
 export class InputFile {
-    public readonly [inputFileData]: ConstructorParameters<typeof InputFile>[0]
+    private [internalInputFileData]: ConstructorParameters<typeof InputFile>[0]
+    public get [inputFileData]() {
+        const file = this[internalInputFileData]
+        return file instanceof Blob ? file.stream() : file
+    }
```
wojpawlik, line 55: I know what I dislike about this change. This logic should be moved into `Client`.

KnorpelSenf: InputFile is semantically part of the client. The two cannot be separated. It just has to reside in a different file for cross-platform reasons.

In a way, the InputFile encapsulates the runtime-specific part of the client. Given that calling `stream()` only works on Deno, it needs to be part of InputFile.

That being said, I know this way to divide the code is mediocre at best. Maybe d2n will improve the situation.

wojpawlik: d2n should be able to shim `Blob`, so that it's possible to `if (Blob && foo instanceof Blob)`.

wojpawlik: Node 16.7.0 added `Blob::stream()`

deno/src/platform.ts
```diff
@@ -46,7 +48,11 @@ export const inputFileData = Symbol('InputFile data')
  * Reference](https://core.telegram.org/bots/api#inputfile).
  */
 export class InputFile {
-    public readonly [inputFileData]: ConstructorParameters<typeof InputFile>[0]
```
wojpawlik, line 49: For 2.0, `InputFile` should accept `Blob` instead of `path: string` (https://github.com/nodejs/node/issues/37340), and expose just `name: string` and a `.stream()` method.

KnorpelSenf: That sounds like a good idea, that would simply the abstraction 👍🏻

wojpawlik: Further, you could turn `InputFile` into an interface, and provide `class InputFileAsyncIterable` instead, perhaps some others. Zero type-switching in your code!

You could still easily detect `InputFile`s with

```ts
function isInputFile(value: any): value is InputFile {
  return value != null && typeof value.stream === 'function'
}
```

Note that web `File` already satisfies

```ts
interface InputFile {
  readonly name: string
  readonly stream: () => ReadableStream
}
```

KnorpelSenf: Again, the goal is to be explicit about time uploads by constructing an `InputFile` object. When bot code is read, it should be immediately obvious that a file upload will be performed. Permitting lots of things in the API directly contradicts this because it allows to circumvent the `InputFile` constructor.

That being said, I'm open to change completely how `InputFile` and client work together if there is a better solution.

wojpawlik: I'd argue that `new upload.Buffer(uint8array, name)` or `new upload.Iterable(streamOrIterable, name)` or `await Deno.getFile(path)` is more explicit than `new InputFile(oneOfFourTypes)`.

I don't think it makes sense to ever merge this PR. Better open a fresh one when working on 2.0.

wojpawlik: On 2nd thought, this could be made mergable:

1. Switch to `deno2node`,
2. Shim `Blob`,
3. Move `Blob` detection to `Client`.

But there doesn't seem to be any demand.

jimmywarting:
> Node 16 and fetch-blob@2 have Blob, but no File. And simple instanceof Blob won't catch all Blob implementations.

fyi, fetch-blob@3 have a File class, it also has a method to get a Blob/File wrapper out of a file path, so it dose not have to have everything in the memory.

Also it has a whatwg compatible `.stream()` method

PR #40, opened 2021-10-03
Check API root for trailing slash

KnorpelSenf:
This will throw an error upon bot creation of the `client.apiRoot` option contains a trailing slash. The kind of configuration is hard to debug, because Telegram simply returns a 404.

This PR also gives away more info in case Telegram returns errors that are hard to understand. It does this by printing sensitive logs (if enabled) also in case the error is returned by Telegram.

### DIFF
--- a/deno/src/core/client.ts
+++ b/deno/src/core/client.ts
@@ -206,6 +206,12 @@ class ApiClient<R extends RawApi> {
         private readonly webhookReplyEnvelope: WebhookReplyEnvelope = {}
     ) {
         this.options = { ...DEFAULT_OPTIONS, ...options }
+        if (this.options.apiRoot.endsWith('/'))
+            throw new Error(
+                `Remove the trailing '/' from the 'apiRoot' option! (Use '${this.options.apiRoot.substr(
+                    this.options.apiRoot.length - 1
+                )}' instead of '${this.options.apiRoot}'.)`
+            )
     }
 
     private call: ApiCallFn<R> = async (method, payload, signal) => {
--- a/deno/src/core/client.ts
+++ b/deno/src/core/client.ts
@@ -241,11 +241,10 @@ class ApiClient<R extends RawApi> {
                 })
             } catch (err) {
                 let msg = `Network request for '${method}' failed!`
-                if (isTelegramError(err)) {
+                if (isTelegramError(err))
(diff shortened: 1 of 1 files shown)


deno/src/core/client.ts
```diff
@@ -235,11 +241,10 @@ class ApiClient<R extends RawApi> {
                 })
             } catch (err) {
                 let msg = `Network request for '${method}' failed!`
-                if (isTelegramError(err)) {
+                if (isTelegramError(err))
                     msg += ` (${err.status}: ${err.statusText})`
-                } else if (this.options.sensitiveLogs && err instanceof Error) {
+                if (this.options.sensitiveLogs && err instanceof Error)
                     msg += ` ${err.message}`
-                }
```
EdJoPaTo, line 242: Personally I wouldn't use if without {} but thats probably personal taste.

deno/src/core/client.ts
```diff
@@ -206,6 +206,12 @@ class ApiClient<R extends RawApi> {
         private readonly webhookReplyEnvelope: WebhookReplyEnvelope = {}
     ) {
         this.options = { ...DEFAULT_OPTIONS, ...options }
+        if (this.options.apiRoot.endsWith('/'))
+            throw new Error(
+                `Remove the trailing '/' from the 'apiRoot' option! (Use '${this.options.apiRoot.substr(
+                    this.options.apiRoot.length - 1
+                )}' instead of '${this.options.apiRoot}'.)`
```
EdJoPaTo, line 213: If the logic already suggests what exactly should be done why not just do it directly without error?

EdJoPaTo (commented):
I dont think this directly closes the Issues #38 #39 as the their "undefined overrides default" behavior is still present and unrelated to this PR.

deno/src/core/client.ts
KnorpelSenf: 1. Being explicit in the user config is better than applying opaque transformations to it.
2. Consistency: https://github.com/grammyjs/grammY/blob/main/deno/src/composer.ts#L385-L392

deno/src/core/client.ts
KnorpelSenf: I do it sometimes … only if the code is still readable enough, however subjective that may be

EdJoPaTo (approved):

PR #41, opened 2021-10-04
Fix error message

KnorpelSenf:
Follow-up for #40


EdJoPaTo (approved):
without thinking about substr syntax it looked also good before :D

PR #42, opened 2021-10-08
Move `@grammyjs/types` to `devDependencies`

rojvv:


rojserbest:
Type definitions of this package? They will be different .d.ts files, they aren't related to `@grammyjs/types`.

KnorpelSenf:
The type definitions of @grammyjs/types need to be shipped to the user. They contain type definitions for the Bot API, which enables grammY to provide type safety across all API methods.

This PR effectively prevents the types from being shipped to library consumers. Hence my question how else you want to ship them, if you via a dependency.

PR #47, opened 2021-10-14
Improve error message for when undefined is passed as token

KnorpelSenf:
Title says it all.

Also adds a test case for this.


KnorpelSenf:
@all-contributors thanks to @amberlionk for a prompt review

PR #50, opened 2021-10-15
Add generic type for command name to Composer.command method

kospra:
When you have a command list that you use both in code and, for example, to automatically set your commands with WebHook in a different place - this typing could be useful.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -292,6 +292,16 @@
       "contributions": [
         "ideas"
       ]
+    },
+    {
+      "login": "kospra",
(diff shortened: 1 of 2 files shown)


KnorpelSenf:
Out of curiosity—the screenshots you posted show Telegraf code. Are you planning on migrating a bot?

kospra:
> Out of curiosity—the screenshots you posted show Telegraf code. Are you planning on migrating a bot?

Ye, looking into it, I like typings here more, but I still on telegraf for the moment.

PR #52, opened 2021-10-15
Check source code format in CI

KnorpelSenf:
Title says it all


wojpawlik:
I tend to first (quickly) lint, then (quickly) check formatting (auto fixable, so less severe than linting), then (potentially slowly) test (if tests don't have a separate job)

PR #53, opened 2021-10-15
Add worktop, AWS, and callback adapters & fix webhook reply

KnorpelSenf:
Previously, the method to be called was not included in webhook replies. This PR fixes the issue.


KnorpelSenf:
@all-contributors add @amberlionk for bug

PR #55, opened 2021-10-15
Automatically pick correct agent based on protocol

KnorpelSenf:
The default `Agent` is only possible to use on HTTPS. This is a problem if a local Bot API server is used, because then the API root can be accessed via HTTP. Prior to this PR, the `Agent` had to be set manually This was not just annoying, it is also hard to find out what the problem is if docs and examples are ignored.

This PR introduces a check for the protocol. From now on, an HTTPS agent is used on HTTPS, an HTTP agent is used on HTTP, and no agent at all is used if for some reason the protocol is something else.


KnorpelSenf:
Converted to draft to await a conclusion of the discussion started by https://t.me/grammyjs/18808

PR #56, opened 2021-10-15
Move to deno2node

KnorpelSenf:
Drops the custom `backport.sh` script and moves to using https://github.com/wojpawlik/deno2node which emerged out of grammY.

Enables GitHub installs of the package by running `npm i grammyjs/grammY`. This will build the project from source.


KnorpelSenf:
```sh
# works:
npm i grammyjs/grammy#move-to-deno2node
```

PR #57, opened 2021-10-16
Support arrays of predicates in filter

KnorpelSenf:
Adds support for the following syntax:

```ts
bot.filter([predicate0, predicate1, predicate2], handler)
```
This is functionally equivalent to:
```
bot.filter(predicate0).filter(predicate1).filter(predicate2, handler)
```


sayem314:
Here is a real-world scenario. I hope helper function names are self-explanatory. `.onlyAdmin` here query database which is an expensive call, it should be called only if `.onlyGroup` is truthy. Now there might be some additional filters that are gonna query the database and parallel execution is gonna be just be non-efficient in this case.

```ts
.filter([helpers.onlyGroup, helpers.onlyAdmin, helpers.hasReplyTo])
```

KnorpelSenf:
I see.

I just noticed that this PR has a poor design choice anyway.

In `bot.on`, you can
- chain calls for AND concatenation, and
- pass an array of filter queries for their OR concatenation.

With `bot.filter`, you can already
- chain calls for AND concatenation, and
- this PR allows you to pass an array—shouldn't it do OR? I.e. check if any of the predicates is true, instead of providing a second way to combine via AND?

Right now, PR is actually completely inconsistent and counterintuitive.

I know that this PR was done in the first place to make AND chaining easier, but now I'm not convinced it's the right thing to do anymore. In fact, I'm inclined to change the behaviour of array-based filtering to logical OR.

Can you think of a different way how grammY can simplify your use case?

Also, note that you can always do
```ts
const sequential = [onlyGroup, hasReply, isAdmin].reduce((pred, p) =>
  async (ctx: MyContext) => await pred(ctx) && await p(ctx))

bot.command('pin').filter(sequential, ctx => { /* ... */ })
```

KnorpelSenf:
Closing as a conclusion of the discussion above.

PR #60, opened 2021-10-18
Add `method` property to webhook reply response

kraftwerk28:
According to the [doc](https://core.telegram.org/bots/api#making-requests-when-getting-updates), when responding to webhook call, the `method` field must be specified in the json.


KnorpelSenf:
In case anyone is wondering: closed as duplicate of #53

PR #61, opened 2021-10-18
Rework framework adapters, and add new ones

kraftwerk28:


KnorpelSenf:
@kraftwerk28 awaiting your review.

src/convenience/frameworks.deno.ts
```diff
@@ -1,38 +1,47 @@
...
 
-export const defaultFramework: SupportedFrameworks = "oak";
+export const stdHttp: AdapterFactory<{ secretPath: string }> = (
+    { secretPath },
+) => (req: Request) => {
+    let resolveResponse: (res: Response) => void;
+    const reqUrl = new URL(req.url);
+    return {
+        update: (req.method !== "POST" || reqUrl.pathname !== secretPath)
+            ? Promise.resolve(undefined)
```
KnorpelSenf, line 16: Why resolving `undefined`? I'd expect a rejection here.

package.json
```diff
@@ -23,12 +23,14 @@
         "@grammyjs/types": "^2.3.1",
         "abort-controller": "^3.0.0",
         "debug": "^4.3.2",
-        "node-fetch": "^2.6.5"
+        "node-fetch": "^2.6.5",
+        "tsscmp": "^1.0.6"
```
KnorpelSenf, line 27: Isn't this only necessary before Node 6?

src/convenience/frameworks.node.ts
```diff
@@ -1,90 +1,85 @@
 // deno-lint-ignore-file no-explicit-any
-import type { IncomingMessage, ServerResponse } from "http";
+import { IncomingMessage, ServerResponse } from "http";
```
KnorpelSenf, line 2: Prefer `import type`. This will allows to not import the module here at runtime.

src/convenience/frameworks.node.ts
```diff
@@ -1,90 +1,85 @@
...
-    }),
-    http: (req: IncomingMessage, res: ServerResponse) => ({
-        update: new Promise<any>((resolve) => {
+export const http: AdapterFactory<{ secretPath: string }> = ({ secretPath }) =>
+    (req: IncomingMessage, res: ServerResponse) => ({
+        update: new Promise((resolve) => {
+            if (
+                req.method !== "POST" || !safeStrCmp(secretPath, req.url ?? "")
+            ) {
+                resolve(undefined);
```
KnorpelSenf, line 17: Ditto—why not rejecting?

src/convenience/frameworks.node.ts
```diff
@@ -1,90 +1,85 @@
...
-                    resolve(JSON.parse(raw));
-                });
-        }),
+                .end(json),
+    });
+
+export const https = http;
+
+export const express: AdapterFactory = () =>
+    (req: any, res: any) => ({
```
KnorpelSenf, line 37: I'm wondering now if we should start importing the types of express as a development dependency in order to have type safety here … what's your opinion?

tsconfig.json
```diff
@@ -13,7 +13,8 @@
         "module": "commonjs",
         "outDir": "./out/",
         "skipLibCheck": true,
-        "target": "es2019"
+        "target": "es2019",
+        "esModuleInterop": true
```
KnorpelSenf, line 17: Isn't this a breaking change?

Either way, I believe it can be reverted when we drop the newly introduced dependency.

src/convenience/webhook.ts
```diff
@@ -119,21 +66,25 @@ const frameworkAdapters: Record<SupportedFrameworks, FrameworkAdapter> = {
...
+    A extends FrameworkAdapter,
+    C extends Context = Context,
+>(
     bot: Bot<C>,
-    framework: SupportedFrameworks = "express",
-    onTimeout: "throw" | "return" | ((...args: any[]) => unknown) = "throw",
-    timeoutMilliseconds = 10_000,
-) {
-    const server = frameworkAdapters[framework] ?? standard;
+    adapter?: A,
```
KnorpelSenf, line 78: I missed this. This is a breaking change so we either have to wait with merging until 2.0 (some time later next year maybe?), or it has to be reverted.

KnorpelSenf: Alternatively, we could introduce a union type. This would still permit string identifiers (non-breaking) but optionally allow people to pass their own adapter. May make it hard to find good types, not sure.

kraftwerk28: Sure, this should be in the next major. Thought I feel that giving a choice among predefined frameworks has less flexibility than allowing to put either a custom one or import from grammY's code. I'll add a temporary union

KnorpelSenf: I don't see why users would have to access the predefined objects. How about not exposing all of them, and simply allowing string identifiers as well as custom objects? One could pass a string to use a predefined one, or an object to define a custom one.

src/convenience/frameworks.node.ts
kraftwerk28: Then we need to have `@types/*` for every framework used there, which is a bit overkill for a few lines of bridging code imo. If you're ok with that, I'll add that too.

KnorpelSenf: That's true, I agree that it's overkill. I'm fine with not having type safety here.

KnorpelSenf:
Why the force-push? Did anything else change?

src/convenience/webhook.ts
KnorpelSenf: Is there a reason for exposing them?

kraftwerk28:
I did a pull-rebase from `main` branch

kraftwerk28:
Is it OK to disable `no-explicit-any` lint rule for `frameworks.*.ts`?

KnorpelSenf:
I see. Now I'll need to go through all old commits again. Merge commits are usually much easier to review. Now that we're squash-merging PRs, so having a few merges on the feature branches isn't problematic.

KnorpelSenf:
> Is it OK to disable `no-explicit-any` lint rule for `frameworks.*.ts`?

Yes.

src/convenience/webhook.ts
rojserbest: A parameter like `adapter: BuiltInAdapters | AdapterObject` is ideal in my opinion.

rojserbest: image
This way it's ideal

KnorpelSenf: That's also what I have in mind. Plus, letting the adapters become implementation details that aren't exported.

src/convenience/frameworks.node.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+
+export const http: AdapterFactory<{ secretPath: string }> = ({ secretPath }) =>
+    (req: IncomingMessage, res: ServerResponse) => ({
+        update: new Promise((resolve, reject) => {
+            if (
+                req.method !== "POST" ||
+                !safeStringCompare(secretPath, req.url ?? "")
+            ) {
+                reject();
```
KnorpelSenf, line 18: Please reject with an error that has a helpful error message

src/convenience/frameworks.node.ts
```diff
@@ -0,0 +1,88 @@
...
+                return;
+            }
+            const chunks: Buffer[] = [];
+            req
+                .on("data", (chunk) => chunks.push(chunk))
+                .once("end", () => {
+                    const raw = Buffer.concat(chunks).toString("utf-8");
+                    resolve(JSON.parse(raw));
+                })
+                .once("error", (e) => reject(e));
```
KnorpelSenf, line 28: nit:
```suggestion
                .once("error", reject);
```

KnorpelSenf (changes requested):
IIRC we agreed on proceeding like this.

1. Do not support patch matching for now, introduce this with a breaking change in the next major version.
2. Do not expose all config objects. Instead, let them be identified via strings. It is still allowed to pass your own config object, which effectively allows you to do path matching manually.
3. The current organisation of files (separating Deno/Node) can stay that way, as long as 2 is applied.

Make sure the identifier of the adapters do not change. This is currently the case for at least `awsLambda`.

src/convenience/webhook.ts
```diff
@@ -149,14 +97,27 @@ export function webhookCallback<C extends Context = Context>(
...
+                bot.handleUpdate(await update, webhookReplyEnvelope),
+                typeof onTimeout === "function"
+                    ? () => onTimeout(...args)
+                    : onTimeout,
+                timeoutMilliseconds,
+            );
+            if (end !== undefined && !usedWebhookReply) end();
+        } catch {
+            // Adapter rejected processing an update
+        }
```
KnorpelSenf, line 119: Also, we should not suppress error, but throw them instead.

kraftwerk28: Does it mean an error occured if the adapter just rejected the request from Telegram (e.g. `requrst.url !== secretPath`)? Or we can create an `class UpdateRejected extends Error` and throw iif `!(error instanceof UpdateRejected)`.

KnorpelSenf: We are not going to support secret path matching in this iteration. If an adapter throws, something is wrong.

If we later implement adapters that choose not to process some requests, they must integrate with their respective web frameworks, and e.g. call `next()`.

KnorpelSenf (approved):
LGTM now.

As stated in the conversation above, we need to change the call signature of `webhookCallback` in order to support more complex options objects in a usable and convenient way. This is a breaking change that will be introduced in the next major release.

Until then, users must route their updates manually, and perform necessary path matching themselves if they decide to go for a low-level framework that does not support this out of the box.

KnorpelSenf:
Thank you for this contribution, and my apologies that it took so long to get this through.

@all-contributors add @kraftwerk28 for code, ideas, and review!

PR #73, opened 2021-11-03
Permit strings for callback and pre-checkout queries

KnorpelSenf:
Title says it all. Discussed in https://t.me/grammyjs/21243

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -1266,12 +1266,12 @@ export class Context implements RenamedUpdate {
      * **Official reference:** https://core.telegram.org/bots/api#answercallbackquery
      */
     answerCallbackQuery(
-        other?: Other<"answerCallbackQuery", "callback_query_id">,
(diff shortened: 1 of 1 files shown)


Loskir (approved):

EdJoPaTo (approved):
That would have made a lot of Telegraf → grammY migrations a lot simpler :D

KnorpelSenf:
> That would have made a lot of Telegraf → grammY migrations a lot simpler :D

You can drop a comment about these things if they annoy you. I know large parts of the API surface by heart, but it's not 100 % :)

EdJoPaTo:
> You can drop a comment about these things if they annoy you.

Guess its always a balance between "more strict" and "easier to use". Both are understandable goals. Migrating to a stricter interface also helps with improving code quality so I saw it partly as a feature and not as "very annoying".

PR #76, opened 2021-11-06
Retry network calls during initialization

KnorpelSenf:
Certain network issues cannot be handled automatically by the library consumer, so it does not make sense to throw these errors. Instead, grammY will now print some debug output and retry.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -3,7 +3,7 @@ import { BotError, Composer, run } from "./composer.ts";
 import { Context } from "./context.ts";
 import { Api } from "./core/api.ts";
 import { ApiClientOptions, WebhookReplyEnvelope } from "./core/client.ts";
-import { GrammyError } from "./core/error.ts";
+import { GrammyError, HttpError } from "./core/error.ts";
 import { debug as d, Update, UserFromGetMe } from "./platform.deno.ts";
 const debug = d("grammy:bot");
 const debugErr = d("grammy:error");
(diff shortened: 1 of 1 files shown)


darvesh:
Is retry with no backoff a good idea though? Especially when it returns 429

KnorpelSenf:
Good point, we should at least wait the recommended number of seconds for 429 errors.

I agree that it's a little weird to wait 3 seconds during polling, but not during initialisation. However, I'd rather go with removing the delay for polling rather than introducing one here. If retrying is the recovery strategy from flaky network issues, I don't see how trying a few more times could cause any harm.

KnightNiwrem (approved):
Any plans to bound the maximum number of retries?

KnorpelSenf:
What would happen after the last retry? Crashing? The point is that we acknowledge these kinds of errors to be temporary and unpredictable. We cannot recover from them or do anything about it. Retrying reliably and infinitely rather than stopping at an arbitrary point is the logical step here.

This PR builds upon advice from the TDLib creator himself. The conversation starts at https://t.me/tdlibchat/22922

src/bot.ts
```diff
@@ -448,3 +450,29 @@ you can circumvent this protection against memory leaks.`);
         this.errorHandler = errorHandler;
     }
 }
+
+async function withRetries(task: () => Promise<unknown>) {
+    let success = false;
+    while (!success) {
+        try {
+            await task();
+            success = true;
```
dzek69, line 459: i'd return the result of `task()` here, so the function is more useful for the future :P

KnorpelSenf: Good point

KnorpelSenf: Good point!

KnorpelSenf: This will be done in #182.

PR #78, opened 2021-11-06
Catch throwing iterators

KnorpelSenf:
If an iterator throws, the corresponding stream causes an unhandled promise rejection. This is an ugly specification, and this PR work around it by wrapping the iterator and cancelling the request upon error.

Closes #34.

Uses ideas from https://github.com/denoland/deno_std/issues/1214.

Issue #34, opened 2021-09-25
Weird reaction on InputFile with path that does not exist


dzek69:
When I pass inexistent path to `InputFile` grammY probably still tries to read it. This simple code:

```javascript
    bot.on("message", (ctx) => {
        console.log("start", new Date().toISOString());
        ctx.replyWithVideo(new InputFile("/tmp/this.does.not.exist")).catch(e => {
            console.error(e);
            console.error("end", new Date().toISOString());
        });
    });
```

Results in:
```
node             | start 2021-09-25T10:27:57.567Z
node             | HttpError: Network request for 'sendVideo' failed!
node             |     at ApiClient.call (/home/node/app/node_modules/grammy/out/core/client.js:55:27)
node             |     at processTicksAndRejections (node:internal/process/task_queues:96:5)
node             |     at ApiClient.callApi (/home/node/app/node_modules/grammy/out/core/client.js:68:22) {
node             |   error: FetchError: request to https://api.telegram.org/xxxxxxxxxxxxxxxxxxxx/sendVideo failed, reason: socket hang up
node             |       at ClientRequest.<anonymous> (/home/node/app/node_modules/node-fetch/lib/index.js:1461:11)
node             |       at ClientRequest.emit (node:events:402:35)
node             |       at ClientRequest.emit (node:domain:475:12)
node             |       at TLSSocket.socketOnEnd (node:_http_client:471:9)
node             |       at TLSSocket.emit (node:events:402:35)
node             |       at TLSSocket.emit (node:domain:475:12)
node             |       at endReadableNT (node:internal/streams/readable:1343:12)
node             |       at processTicksAndRejections (node:internal/process/task_queues:83:21) {
node             |     type: 'system',
node             |     errno: 'ECONNRESET',
node             |     code: 'ECONNRESET'
node             |   }
node             | }
node             | end 2021-09-25T10:28:57.618Z
```

KnorpelSenf:
Needs to fixed when the workaround for https://github.com/denoland/deno_std/issues/1214 is implmeneted

KnorpelSenf:
@dzek69 please review the PR that fixes this.

PR #79, opened 2021-11-06
Request timeouts

KnorpelSenf:
Adds a timeout of 500 seconds to every API request.

Note that calls to `getUpdates` can never take more than 50 seconds.

Builds on the changes introduced by #78 so it should only be reviewed and merged afterwards.


dzek69:
@KnorpelSenf can you update this PR so we don't see the changes introduced in #78 in the diff view?

KnorpelSenf:
@dzek69 please check again

src/core/client.ts
```diff
@@ -252,21 +252,28 @@ class ApiClient<R extends RawApi> {
 
             const res = await new Promise<ApiResponse<ApiCallResult<M, R>>>(
                 (resolve, reject) => {
-                    function onStreamError(err: unknown) {
+                    const onHttpError = toHttpError(method, sensLogs, reject);
+                    function onOtherError(err: unknown) {
                         abort();
```
dzek69, line 257: won't calling abort() cause `reject` to be called through `onHttpError`?

repeated calls to `reject` are ignored, but this is a code smell usually, with Node.js you can even track something like that: https://nodejs.org/api/process.html#event-multipleresolves

KnorpelSenf: True, it could cause several rejections because `fetch` will eventually also throw when the request is cancelled. We could introduce a flag `rejected` that prevents this, but I don't like that because I'm already unhappy with how hacky this gets. Do you have any suggestions?

dzek69: I guess we can simply remove this `reject(err)` after abort? Aborting request should immediately throw and won't slow anything down or something

KnorpelSenf: We cannot remove it because that would change which error is thrown. It will no longer be clear the the request timed out. Instead, it only looks like the request was cancelled, which is confusing.

KnorpelSenf: Update: we can now do it! The web spec was changed in https://github.com/whatwg/dom/pull/1027 and Deno 1.16 has already shipped support. I will update this PR accordingly as soon as I know how to achieve this on Node.

KnorpelSenf: Well, it does not look like we're going to get support for this anytime soon. The `abort-controller` package that is used under Node.js had its last publish 3 years ago, and the repo is pretty dead.

We now have several options.
1. Implement a hacky version with flags and checks to prevent repeated rejections.
2. Implement a new abstraction that works similarly to `deferred` from https://deno.land/std and that catches and suppressed subsequent rejections when already rejected.
3. Ignore the problem and simply reject multiple times. This is not good, but it only really happens in rare cases and where there are errors anyway.
4. Accept that the error message does not tell you what a request failed. Can be mitigated by manipulating the error message on the fly (hacky) or by adding a note into the docs ([here](https://grammy.dev/guide/deployment-types.html#ending-webhook-requests-in-time)).
5. Not merging.

Here is how much I like those options on a scale of “I'm fine with that” to “Not really an option”: 4-2-1-3-5

@dzek69 what is your opinion?

@KnightNiwrem @trgwii @wojpawlik @EdJoPaTo feel free to leave a comment here, too, if you have time.

dzek69: Well. Multiple rejections will probably never cause any real issue, won't crash Node/other engines or anything.

1. That's why 1 seem pointless if the code would stay very similar to what we currently have.
2. again is pointless. The code smell will persist, only internally it won't execute, but it does not matter.
3. I'd keep that. I personally leave not urgent stuff like that unfixed with a @TODO comment. It may get fixed when I'm bored or when better solution will become available. Sometimes it may never be fixed. Or after 10 years when it actually start causing real trouble (like JS engines starting to crash in such situation). I know people may not like having tens of @TODOs in the project but I think that's a sane compromise sometimes. I don't know what do you think about it.
4. We can as well fix only for Deno and leave Node with less meaningful messages 🤷🏻‍♂️ 
5. I don't even know what happens without that change on requests that takes more time. Can you share some info?
6. I don't know if node-fetch will allow different implementation of abort-controller on node. I had some troubles with that long time ago. Not sure if that's still an issue, but it sounds like it's worthy to give it a try. Not like I know any other implementation of it. But maybe forking it is actually a good idea.

KnorpelSenf: 3. Fine by me.
4. That's hard, because then we'd have to extract the core part of the client into the platform-specific files, so we can write two clients. I'd rather not.
5. Requests may never terminate without timeouts. When using the default polling, this will make the bot hang indefinitely.
6. I think it's possible as long as it follows the specs. At least that's what `node-fetch` promises. Still, it is a lot of work.

KnorpelSenf: The problem with implementing 4 and ignoring the error message is that we then end up with #34 again.

dzek69: Ok, so at first I wanted to say to merge it but one got me thinking...

Regarding `5.` - in the code there is `// 500 seconds timeout, same as in Bot API` -- where you did found that information?

I took a look at the docs and found nothing, so I went through Telegram Bot Api code and found this:
- https://github.com/tdlib/telegram-bot-api/blob/e9587a0083c8b9e9c058ec71aec94055f1cd595d/telegram-bot-api/HttpServer.h#L69 - seems like there is 500 here, but I verified this and
- https://github.com/tdlib/td/blob/377bd18909b7a602b40e8bb074cdeb2830baef91/tdnet/td/net/HttpInboundConnection.cpp#L18 this is called `idle_timeout`. `idle` got me thinking again because I remember (but my memory sometimes are making jokes on me) having a huge upload on my personal (slow to upload) network (with self-hosted bot api). It took around 20 minutes IIRC but uploaded just fine. So I dug deeper and
- https://github.com/tdlib/td/blob/377bd18909b7a602b40e8bb074cdeb2830baef91/tdnet/td/net/HttpConnectionBase.cpp#L42 this timeout seems to be reset as long something is happening with the connection: https://github.com/tdlib/td/blob/377bd18909b7a602b40e8bb074cdeb2830baef91/tdnet/td/net/HttpConnectionBase.cpp#L141 (this last part I'm not sure I understand correctly, but it seems just a reading loop to me)

So, to sum up:
- having a timeout is nice, I'd personally want a low timeout of ~10-30 seconds for most requests, but it seems that hardcoded value of 500 seconds is just wrong and it should be configurable. Either globally but I'd more prefer to have it per request

dzek69: I don't get what #34 has common with this issue. You mean jus the "weird reaction" part from the title?

src/core/client.ts
```diff
@@ -227,57 +227,67 @@ class ApiClient<R extends RawApi> {
...
+                    if (rejected) return;
+                    rejected = true;
+                    reject(err);
+                    abort();
+                }
+                // Create fetch config
+                const config = formDataRequired
+                    ? createFormDataPayload(p, onOtherError)
+                    : createJsonPayload(p);
+                // Use same timeouts for requests as Bot API
```
dzek69, line 278:
```suggestion
                // Use same timeouts for requests as Bot API
                if (rejected) return; // don't start the request if createFormDataPayload already rejected
```

src/core/client.ts
KnorpelSenf: Source of 500 seconds: https://t.me/tdlibchat/23377

If you pass a file path that doesn't exist, the stream that makes up the request body will throw, and fetch fails the request with a cryptic message. We need to catch the error from the iterator before it reaches fetch.

src/core/client.ts
KnorpelSenf: You're technically right if we regard the payload creation as a black box. Practically, this can never happen. The error callback can only be invoked when the body is consumed, which only happens as soon as it's passed to `fetch`

dzek69: oh, right, i missed that

src/core/client.ts
KnorpelSenf: I'm generally against any kind of configuration unless there's a very good reason to introduce it. For example, shorter timeouts are easily added via API transformers. If you wanted longer timeouts, that would be something to think about.

Is there a good reason why people should concern themselves with a timeout value of each API call?

src/core/client.ts
```diff
@@ -227,50 +227,67 @@ class ApiClient<R extends RawApi> {
...
+                const timeoutHandle = setTimeout(() => {
+                    const msg =
+                        `Request to '${method}' timed out after 500 seconds`;
+                    onOtherError(new Error(msg));
+                }, 500_000); // 500 seconds
+                // Perform request
+                fetch(url, { ...options, ...config }).then((res) => res.json())
+                    .then(resolve)
+                    .catch(onHttpError)
+                    .finally(() => clearTimeout(timeoutHandle));
```
KnightNiwrem, line 288: This is roughly what I have in mind from the comment in telegram: https://replit.com/join/mmkxuuwfqm-knightniwrem

KnorpelSenf: I see. In other words, we need to check `rejected` before we `resolve` the result of `json`? I don't think we can cancel the parsing somehow.

I think the code already works, but we're calling resolve after reject, which is a bad practice.

KnightNiwrem: Hmm.. I'm less concerned about resolving after rejecting, than I am about how the behavior looks like to the developers and users of the bot.

Presumably, if `fetch(...)` completes (e.g. sendMessage), then that means the user would see the message in the chat. However, if timeout intercepts this, it should attempt to abort an already-completed fetch operation, and then reject (i.e. throw the api call). From the developer's perspective, they would assume that the sendMessage failed due to timeout, but in reality, the message is present in the chat.

src/core/client.ts
dzek69: You are right - there is no need to make people concerned - so we should provide a sane default.

But for example I personally have a good reason to want this configured - because I host my bot api server on slow network and as I said - I had uploads taking 20 minutes (so 500 seconds hard limit isn't high enough for me).

Also for high-traffic bots If I'm sending just text messages or commands - I expect it to take max 10 seconds. If it takes longer there is 99% chance it's stuck and won't finish even in those 500 seconds. I'd rather want it to fail fast and let me (or auto-retry) retry the request or consider it failed and done, so I can free the queue for other users of my bot.

But when sending media this 10 seconds timeout just doesn't fit. I need it higher for requests that I expect to take longer.

src/core/client.ts
dzek69: We can clear timeout inside first `then`

src/core/client.ts
EdJoPaTo: The upload speed is an interesting thought here: With 2GB as the current filelimit 500 seconds would require 4MB/s or higher upload to avoid the timeout.

KnorpelSenf: The 500 seconds do indeed assume that very large file transfers (note that a request can potentially be up to 20 GB) are performed locally. That's btw also a reasonable thing to do, but while I'd recommend you to not push everything over the network _twice_, I still want to build a library that works in all environments. That means that grammY shouldn't limit you from doing it, no matter how inefficient it is.

There's no fixed sane default in your setup. You'd need to dynamically adjust timeouts based on file size, which is a parameter that grammY cannot know. It supports streams with no known length.

What do you think about a flag that optionally disables timeouts entirely? That way, you can keep your architecture, and we still have our sane default.

src/core/client.ts
KnorpelSenf: Now I understand what you mean, good point! I did not think of that. I'll fix it :)

src/core/client.ts
dzek69: > If the selfhosted server is hosted at the same site the upload to the selfhosted Server is fast → no problem.

@EdJoPaTo actually not. API server must successfully upload the data to the Telegram servers to be able to respond with success result. So even if API server would read the data as fast as possible it would still wait with no transfer between the bot before sending a reply. But it actually reads the data from the bot not faster than it sends it out to Telegram servers to prevent keeping slow but huge uploads in the memory (I guess, but this sounds reasonable). At least that how it looks when I upload stuff on my slow network.

dzek69: > request can potentially be up to 20 GB
Isn't the limit around 5GB?

Globally disabling timeouts seems ok, but only ok. I don't want GrammY to guess the timeout - that sounds like out of the scope of a library. So I'm still all for having this configurable but to prevent having more parameters withing the api methods we could add something like `buildUrl` but for timeout. I don't fully understand API transformers yet but reading your previous comment I understand this is no good for fully controlling the timeout.

KnightNiwrem: If 1-reject-per-promise is desirable (which I'm not too bothered about either), I'd imagine that a promise-only implementation would be something like (also no longer returns a giant promise):

```ts
const waitFetch = new Promise(async (resolve, reject) => {
    try {
        const res = await fetch(url, { ...options, ...config });
        resolve(res);
    } catch (err) {
        let msg = `Network request for '${method}' failed!`;
        if (isTelegramError(err)) {
            msg += ` (${err.status}: ${err.statusText})`;
        }
        if (opts.sensitiveLogs && err instanceof Error) {
            msg += ` ${err.message}`;
        }
        reject(new HttpError(msg, err));
    }
});

let onStreamError: Function;
const waitStreamError = new Promise((_, reject) => {
    onStreamError = (err: unknown) => {
        abort();
        reject(err);
    }
});

const config = formDataRequired
    ? createFormDataPayload(p, onStreamError)
    : createJsonPayload(p);

let timeoutHandle: any;
const waitTimeoutError = new Promise((_, reject) => {
    timeoutHandle = setTimeout(() => {
        const msg = `Request to '${method}' timed out after 500 seconds`;
        abort();
        reject(new Error(msg));
    }, 500_000);
});

// Perform request
try {
    const res = await Promise.race([
        waitFetch,
        waitStreamError,
        waitTimeoutError,
    ]);
    clearTimeout(timeoutHandle);
    const resJSON = await res.json();
    return resJSON;
} catch (err) {
    clearTimeout(timeoutHandle);
    throw err;
}
```

KnightNiwrem (approved):

dzek69:
@KnorpelSenf so the 500 seconds hardcoded timeout stays for now?

KnorpelSenf:
There's one problem that remains. Calls to `getFile` cause the Bot API server to download the file from Telegram. This won't complete in time on slow connections, so we'll have to rethink this PR before merging.

KnorpelSenf:
We can add extra config for timeouts that only applies for file transfers to local Bot API servers. Thereby we keep the amount of configuration low for most users, and still allow the use with slow servers that need a very long time to perform file transfers.

src/core/client.ts
KnorpelSenf: This implementation is cool, encapsulating the multiple rejections behind a `Promise.race` is smart. I think we should do it that way

dzek69:
File transfers may be slow as well when the bot is hosted at home server and "global" bot api server is used. So being able to set some extra config only when apiRoot is non-default isn't that good too.

IMO still the best option is to let user decide about timeout value by using some method similiar to `buildUrl` that will get method name and will return a timeout value to use. It's pretty simple and should match all use cases, slow and fast connections, etc.

KnorpelSenf:
What upload speeds are you calculating? The default assumes you don't go below 1024 kbps on average per request. Are you hosting bots on slower connections?

dzek69:
I was thinking about `getFile` actually, but I forgot global bot api has limit for downloads as well (and it's actually lower than uploads for some reason). So yeah, exceeding default limit should be possible only on local bot api.

KnorpelSenf:
Reading the source code of the Bot API server, it looks like the 500 seconds per request are hard-coded and cannot be changed, also not locally. We might still want to add a timeout in order to support third-party API servers, but can you first confirm that you are indeed able to run the Bot API server locally and perform requests that exceed 8 minutes and 20 seconds?

dzek69:
I will test it again right now, give me around 1 hour and I will report back with detailed results.

dzek69:
Still experimenting. I want to make sure that my conclusions are valid :)

KnorpelSenf:
Take your time. I won't be able to review our commit anything in the next 12 hours.

dzek69:
Ok, I've tested this several times and got pretty different results.

Quick summary:
- My longest living (but finally errored) connection between bot and the local bot api server was 12 minutes 28 seconds
- My longes successful living connection between bot and the local bot api server was 8 minutes and 37 seconds
- All (except one) the requests actually got response from bot api server. The response was sometimes invalid/unexpected, but it was there.

TEST NUMBER #1:

- My bot started uploading 1,7 GB file on 15:02:33:
image
This message indicates upload start. Ignore the `burze.dzis.net` link - it's just an url trigger - i hardcoded the returned files for testing purposes.

- When the upload from bot to api server was started I immediately got this on bot api server: `[ 3][t 0][1636812153.529000520][HttpConnectionBase.cpp:122][!HttpInboundConnection]     Slow HTTP connection: migrate to 4` - it's weird to consider the connection slow in the first second, but whatever. The timestamp given here translates to: Sat Nov 13 2021 15:02:33 GMT+0100
- Then there was no useful info about the file transfer between bot and bot api server - just the usual cpu usage and getUpdates calls. This means bot api server does not try to upload the file while it's being downloaded from the bot.
- on 15:06:40 (4 minutes, 13 seconds) bot api server completed receiving the file from the bot and started uploading it to Telegram servers:
```
[ 3][t 0][1636812400.771632909][Query.cpp:46][!HttpConnection]  QUERY: create [ptr:0x7f12e0e6d330][botBOT_ID_WAS_HERE:BOT_KEY_WAS_HERE ][time:  6269.7ns][method:           sendvideo]{[chat_id;443314775], [video;attach://zjtujmcdook86w8y], [type;video], [reply_to_message_id;4606]}{[name:output.mp4][size:1761371445]}
[ 3][t 0][1636812400.771727323][Td.cpp:3184][#BOT_ID_WAS_HERE][!Td][&td_requests]     Receive request 34359802368: getMessage {
  chat_id = 443314775
  message_id = 4829741056
}
// some lines omitted here
[ 3][t 0][1636812400.803952932][Td.cpp:3184][#BOT_ID_WAS_HERE][!Td][&td_requests]     Receive request 30064835840: sendMessage {
  chat_id = 443314775
  message_thread_id = 0
  reply_to_message_id = 4829741056
  options = messageSendOptions {
    disable_notification = false
    from_background = false
    scheduling_state = null
  }
  reply_markup = null
  input_message_content = inputMessageVideo {
    video = inputFileLocal {
      path = "/tmp/output.mp4"
    }
    thumbnail = null
    added_sticker_file_ids = vector[0] {
    }
    duration = 0
    width = 0
    height = 0
    supports_streaming = false
    caption = formattedText {
      text = ""
      entities = vector[0] {
      }
    }
    ttl = 0
  }
}
[ 3][t 0][1636812400.803993463][MessagesManager.cpp:23343][#BOT_ID_WAS_HERE][!Td]     Begin to send message to chat 443314775 in reply to server message 4606
[ 3][t 0][1636812400.804008483][MessagesManager.cpp:31895][#BOT_ID_WAS_HERE][!Td]     Search for server message 4606 in chat 443314775 found 0x7f12e1162490
[ 3][t 0][1636812400.804087877][MessageContent.cpp:1916][#BOT_ID_WAS_HERE][!Td]       Get input message content from inputMessageVideo {
  video = inputFileLocal {
    path = "/tmp/output.mp4"
  }
  thumbnail = null
  added_sticker_file_ids = vector[0] {
  }
  duration = 0
  width = 0
  height = 0
  supports_streaming = false
  caption = formattedText {
    text = ""
    entities = vector[0] {
    }
  }
  ttl = 0
}
[ 3][t 0][1636812400.804137468][FileManager.cpp:1195][#BOT_ID_WAS_HERE][!Td]  Register file data [[remote_name:] chat 443314775 [size:1761371445][expected_size:0] NoKey{} local [full local location of videos] at "/tmp/output.mp4"{}] as 998(0) from register_local
[ 3][t 0][1636812400.804167509][FileManager.cpp:3036][#BOT_ID_WAS_HERE][!Td]  Checking file 998(0) of type videos/videos
[ 3][t 0][1636812400.804177045][FileManager.cpp:1055][#BOT_ID_WAS_HERE][!Td]  Dup file 998(0) to 999(0)
[ 3][t 0][1636812400.804192543][MessageContent.cpp:1607][#BOT_ID_WAS_HERE][!Td]       Create InputMessageContent with file 999(0) and thumbnail 0(0)
[ 3][t 0][1636812400.805367231][VideosManager.cpp:52][#BOT_ID_WAS_HERE][!Td]  Receive video 999(0)
[ 3][t 0][1636812400.805408954][FileManager.cpp:1055][#BOT_ID_WAS_HERE][!Td]  Dup file 999(0) to 1000(0)
[ 3][t 0][1636812400.805426120][MessagesManager.cpp:31895][#BOT_ID_WAS_HERE][!Td]     Search for server message 4606 in chat 443314775 found 0x7f12e1162490
[ 3][t 0][1636812400.805442571][MessagesManager.cpp:32099][#BOT_ID_WAS_HERE][!Td]     Adding yet unsent message 4607.1 of type Video to chat 443314775 from send message. Last new is server message 4607, last is server message 4607, from_update = true, have_previous = true, have_next = true
[ 3][t 0][1636812400.805457115][MessagesManager.cpp:31895][#BOT_ID_WAS_HERE][!Td]     Search for yet unsent message 4607.1 in chat 443314775 found 0
[ 3][t 0][1636812400.805514812][MessagesManager.cpp:32453][#BOT_ID_WAS_HERE][!Td]     Adding not found yet unsent message 4607.1 to chat 443314775 from send message
[ 3][t 0][1636812400.805570602][MessagesManager.cpp:32526][#BOT_ID_WAS_HERE][!Td]     Attach yet unsent message 4607.1 to the previous server message 4607 in chat 443314775
[ 3][t 0][1636812400.805585861][MessagesManager.cpp:13635][#BOT_ID_WAS_HERE][!Td]     Set chat 443314775 last message to yet unsent message 4607.1 from add_message_to_dialog
[ 3][t 0][1636812400.805603027][MessagesManager.cpp:23659][#BOT_ID_WAS_HERE][!Td]     Do send yet unsent message 4607.1 in chat 443314775
[ 3][t 0][1636812400.805613994][MessagesManager.cpp:23709][#BOT_ID_WAS_HERE][!Td]     Ask to upload file 1000(0) with bad parts {}
[ 3][t 0][1636812400.805623054][FileManager.cpp:2503][#BOT_ID_WAS_HERE][!Td]  Resume upload of file 1000(0) with priority 1 and force = false
[ 3][t 0][1636812400.805651187][FileManager.cpp:2548][#BOT_ID_WAS_HERE][!Td]  Change upload priority of file 1000(0) to 1
[ 3][t 0][1636812400.805662393][FileManager.cpp:2640][#BOT_ID_WAS_HERE][!Td]  Skip run_generate, because file 998(0) can't be generated
[ 3][t 0][1636812400.805670261][FileManager.cpp:388][#BOT_ID_WAS_HERE][!Td][&update_file]     File 998(0) has changed upload priority to 1
[ 3][t 0][1636812400.805689334][FileManager.cpp:2829][#BOT_ID_WAS_HERE][!Td]  File 1000(0) upload request has sent to FileLoadManager
[ 3][t 0][1636812400.805724143][FileManager.cpp:2994][#BOT_ID_WAS_HERE][!Td][&update_file]    Send file 1000(0) as 998(0) and update send_updates_flag_ for file 998(0)
[ 3][t 3][1636812400.805725336][FileUploader.cpp:132][#BOT_ID_WAS_HERE][!Uploader]    In FileUploader::on_update_local_location with [full local location of videos] at "/tmp/output.mp4". Have path = "/tmp/output.mp4", local_size = -1, local_is_ready = true and file type = videos
[ 3][t 0][1636812400.805746793][FileManager.cpp:2994][#BOT_ID_WAS_HERE][!Td][&update_file]    Send file 1000(0) as 998(0) and update send_updates_flag_ for file 998(0)
[ 3][t 3][1636812400.805754423][FileUploader.cpp:173][#BOT_ID_WAS_HERE][!Uploader]    Set file local_size to 1761371445
[ 3][t 0][1636812400.805771112][Td.cpp:4365][#BOT_ID_WAS_HERE][!Td][&td_requests]     Sending update: updateNewMessage {
  message = message {
    id = 4830789633
    sender = messageSenderUser {
      user_id = BOT_ID_WAS_HERE
    }
    chat_id = 443314775
    sending_state = messageSendingStatePending {
    }
    scheduling_state = null
    is_outgoing = true
    is_pinned = false
    can_be_edited = false
    can_be_forwarded = true
    can_be_deleted_only_for_self = false
    can_be_deleted_for_all_users = true
    can_get_statistics = false
    can_get_message_thread = false
    can_get_media_timestamp_links = false
    has_timestamped_media = false
    is_channel_post = false
    contains_unread_mention = false
    date = 1636812400
    edit_date = 0
    forward_info = null
    interaction_info = null
    reply_in_chat_id = 443314775
    reply_to_message_id = 4829741056
    message_thread_id = 0
    ttl = 0
    ttl_expires_in = 0.000000
    via_bot_user_id = 0
    author_signature = ""
    media_album_id = 0
    restriction_reason = ""
    content = messageVideo {
      video = video {
        duration = 0
        width = 0
        height = 0
        file_name = "output.mp4"
        mime_type = "video/mp4"
        has_stickers = false
        supports_streaming = false
        minithumbnail = null
        thumbnail = null
        video = file {
          id = 998
          size = 1761371445
          expected_size = 1761371445
          local = localFile {
            path = "/tmp/output.mp4"
            can_be_downloaded = false
            can_be_deleted = false
            is_downloading_active = false
            is_downloading_completed = true
            download_offset = 0
            downloaded_prefix_size = 1761371445
            downloaded_size = 1761371445
          }
          remote = remoteFile {
            id = ""
            unique_id = ""
            is_uploading_active = true
            is_uploading_completed = false
            uploaded_size = 0
          }
        }
      }
      caption = formattedText {
        text = ""
        entities = vector[0] {
        }
      }
      is_secret = false
    }
    reply_markup = null
  }
}
```
- Now the transfer between bot and bot api server stopped - bot waits for api server to respond with something
- Console messages in bot api server were repeating with `uploaded_size` value updated.
- On 15:15:01 (8 minutes 21 seconds) Bot api server responsed with unexpected response: `502 Bad Gateway\n` which resulted in node-fetch complaining about not being able to decode json:
```
FetchError: invalid json response body at http://srv.lan:8081/botBOT_ID_WAS_HERE:BOT_KEY_WAS_HERE/sendVideo reason: Unexpected token B in JSON at position 4 at /home/node/app/node_modules/@dzek69/grammy/node_modules/node-fetch/lib/index.js:273:32 at runMicrotasks (<anonymous>) at processTicksAndRejections (node:internal/process/task_queues:96:5) at value (/home/node/app/node_modules/@dzek69/grammy/out/core/client.js:80:28) at /home/node/app/node_modules/@grammyjs/auto-retry/out/index.js:26:22
```
image
- Bot api continues to upload file to TG servers and finally on 15:26:16 the message arrives:
image

TEST NUMBER #5:

- 16:38:49 bot started uploading 755 MB file
- ~16:41:47 (2 minutes 58 seconds) bot finishes and now bot api server started uploading
- 16:50:52 (12 minutes 3 seconds in total, 9 minutes and 5 seconds since communication stopped) I received the message
- I did not get a response from bot api server at all! This is where having a timeout is actually a valid need! I may got false information (error when the request actually was quite successful) but having *some* information will be better than no information at all (user could for example block his tasks queue and my bot would never start working on any message from this user)

TEST NUMBER #7:

- 17:19:51 bot started uploading 630 MB file
- 17:21:04 (1 minute, 13 seconds) bot api srv started uploading
- 17:28:28 (I received the message and got success response from bot api server, 8 minutes 37 seconds in total, 7 minutes and 24 seconds since communication between bot and api server stopped)
image
- This is where having a 500 second timeout counted just from the beginning would cause me to get timeout but the request would end ok without it.

Conclusion:

- While the file is transferred from the bot to bot api server - it counts as active connection. The 500 seconds hardcoded limit as I discovered earlier and described [here](https://github.com/grammyjs/grammY/pull/79#discussion_r745842653) only applies to connections when no data is being transferred. During file transfer this timeout gets reset until data is no longer exchanged.
- Also it is possible for someone to modify the hardcoded 500 seconds timeout and built their own version of bot api server
- Ideally - we should replicate the same thing from bot perspective. I mean - while data is being exchanged it should be counted as active connection. No data transfer should start the countdown to Timeout error. I believe this is not possible with node-fetch and would require stepping away from it to something more raw. 
- So in general the point we should focus around is that a valid the connection between bot and bot api server may take longer than 500 seconds
- Current implementation is okay to merge, it can be improved later but I'd prefer to have it configurable, so I don't have to use a fork of grammy for my own needs.

Also some more comment: I know my environment is pretty specific and requires pretty specific file to make timeout in current implementation to backfire at ime, but let's consider another environment example:

- Bot api server is hosted on a dedicated machine
- Bot itself is hosted at home on slower network

This could result in:
1) 1000 seconds upload from bot to bot api server - this is slow part - but bot api server would not end connection because it is valid from its perspective (as well as from node app perspective)
2) 20 seconds upload from bot api server to tg servers (because dedicated machine has ultra fast upload)
3) bot api server repsonds with success message in total of 1020 seconds

BTW: This example what I actually planned to do initially, I wanted to move bot api server to dedicated machine but keep the bot at home. I wanted to allow some friends to have their bots at their homes but use "my" "local but external" bot api server to work around default bot limits.

There are other valid use cases for that as well.

I hope that's helpful for your following considerations what to do about proposed changes :) it took me a while

KnorpelSenf:
Thank you so much for the thorough testing!

Having timeouts for requests is both very important and not urgent at all. This means that I view it much more important to get this _right_ than to get this _soon_.

I am going to take some time to study the Bot API server more closely in order to explain your observations. I don't think I have any other comments before that—except maybe this one:

> I believe this is not possible with node-fetch and would require stepping away from it to something more raw.

Quite the contrary, it's pretty simple with grammY. I just somehow didn't think it was a good idea, but if the Bot API server does it that way, we can learn from that, and adapt.

dzek69:
> I am going to take some time to study the Bot API server more closely in order to explain your observations.

What exactly is still unclear? Maybe I will try to explain more precisely.

> Quite the contrary, it's pretty simple with grammY.

Can you explain how? I think you could somehow "simulate" it with streams (I'm not a stream pro), but how about Buffers? Are you internally converting Buffers into streams as well?

KnorpelSenf:
> Maybe I will try to explain more precisely.

Your explanations are good. They just don't all match with what I've been assuming so far, so need to have a deeper understanding of why the Bot API server does what you observe, and which code is responsible for that.

> Can you explain how?

grammY has a custom multipart/form-data implementation that builds upon a generator of buffer objects. We can tap into this byte stream and postpone the timeout every time a chunk is sent. Check out the implementation, it's surprisingly straightforward.

dzek69:
> We can tap into this byte stream and postpone the timeout every time a chunk is sent.

It's good to hear we can do that! That's basically the same thing that bot api server is doing.

If any more questions arise just ping me.

KnorpelSenf:
I still have not looked into the Bot API server to determine what we have to do about longer timeouts. Obviously, if the conclusion of that investigation is that requests may take longer than 500 seconds and the Bot API server does not terminate them, then the code needs to be adjusted further.

Otherwise, this implementation would be complete. I applied @KnightNiwrem's suggestion and optimised it further.

KnorpelSenf:
I did not understand from the C++ code how HTTP connections are able to live 17 seconds longer than 500 seconds. Maybe I did not delve too deep enough into it, or something fishy is going on, but either way I do not have the motivation to investigate this further.

Instead, going to back to first-principle thinking, these are the things we all agree on (I hope):
- We must have a timeout for all requests.
- The timeout must be in the range of minutes, rather than hours, in order to be useful for the majority of bots.
- We must have support for optional configuration of the timeout, since people could use their own Bot API server in a controlled environment where connections can live for hours and still be working correctly. This is probably really bad user experience, but grammY should not dictate anything here.

My only concern about adding config was that it's not necessary for, well, basically everyone else except those mentioned in the last point. I have now put into the documentation that most people should not care about the option, and elaborated why that is the case. Merging the timeout config with the Bot API server config couples two originally orthogonal concepts, so it is now ideal, either.

I am now okay with merging this.

@dzek69 can you confirm that the suggested changes work well for you?

KnorpelSenf:
Regarding the idea to postpone the timeout for every sent chunk: this does not work if the Bot API server is hosted locally on a slow connection when `getFile` calls are performed. In other words, it only works if the connection to the Bot API server is slow, but fails in cases where the Bot API server has a slow connection to Telegram. Hence, this only solves part of the problem. Consequently, it is not the right choice to add this complexity to the form-data creation.

dzek69:
@KnorpelSenf I will take another look at this on the weekend

KnorpelSenf:
@dzek69 any updates?

dzek69:
@KnorpelSenf i was too busy, sorry. Maybe this weeked as this change is actually blocking me from releasing my bot, so I definitely want to review it.

dzek69:
Ok, I spent some time on this again and I'm sad you don't consider to postpone timeout with every chunk, because it hits me personally (I will actually host my bot api server on the VPS and the bot code at home). It may help with part of the problem or not with every possible use case, but it's also the most complete solution possible to implement. If the complexity isn't "too complex" I'd go with it. Just because we can't do impossible doesn't mean we should do nothing :)

I do understand that user shouldn't be bothered with timeouts until really needed, I have also no idea what would be "sane" default timeout - maximum 500 seconds sounds good if you don't want to have a lot of `Grammy can't send files` issues open, but it's pretty useless when sending just text messages IMO. I'd go with 500 seconds default (as it is), but if that's al least hackable with Bot API transformers (is it? I haven't try that yet) then it's fine. If not - I'd implement a way to change the timeout per request before considering this issue solved.

dzek69:
I will try my best to reply ASAP in this thread from now on

KnorpelSenf:
Postponing per chunk is only a solution for sending files via form-data uploads. It's not a solution for other uploads. It's not a solution for receiving files. This would be a very leaky abstraction.

Are you aware that this PR allows you to set your own timeout values or even disable them completely? This will work for everyone who wants to tolerate extremely poor connection speeds in production. There are no caveats to this—in contrast to the other idea.

dzek69:
ok, let's skip the postponing then

i am aware that i can change/disable the timeout, but it's a global setting.

Is there any way of being able to set different value per call? ie via bot api transformers?

KnorpelSenf:
Unfortunately, the transformers aren't flexible enough for this yet. However, we're going to rewrite this part of grammY completely for a distant 2.0 release. It will basically be changed to using `(prev, cctx) => result` and then we can easily add this functionality per call.

I would like to avoid postponing this. There are no timeouts now, and by disabling them you'll get the same behaviour in future versions. You can then opt in again after we have request-specific timeouts in 2.0.

roj1512:
Yep.

dzek69:
Okay, let's merge it.

PR #80, opened 2021-11-06
Support creating InputFile instances from URLs

### LINKED ISSUE #33: Feature request: ability to re-upload a file from given url
I want to use for example `sendMediaGroup` with an URL but I don't want grammy to pass the URL to Telegram but instead in the background download the file and upload it as a new file

why?

Telegram fetches the media information from URL and caches it. Small enough MP4 files with no audio are considered Animations. MP4 with audio will be a Video.

I want to send an album with two mp4 files, one has audio and one has not, so they are Video + Animation. And Telegram does not allow me to send them as media group as Media Groups cannot contain Animations.

If i upload files from disk with InputFile then it's ok:
```javascript
ctx.replyWithMediaGroup([
    {
        type: "video",
        media: new InputFile("/tmp/evv/no-audio.mp4"),
    },
    {
        type: "video",
        media: new InputFile("/tmp/evv/yes-audio.mp4"),
    },
])
```
Telegram won't use it's cache on uploads and will consider my mp4 files a Videos.

If I use URL: 
```javascript
ctx.replyWithMediaGroup([
    {
        type: "video",
        media: "https://dev.nitra.pl/no-audio.mp4",
    },
    {
        type: "video",
        media: "https://dev.nitra.pl/yes-audio.mp4",
    },
])
```

Telegram will return:
```json
{
  "ok": false,
  "error_code": 400,
  "description": "Bad Request: wrong file identifier/HTTP URL specified"
}
```

Proposed change could be useful in other cases as well - like my bot has access to some url, but telegram servers don't - useful with local http-based microservices.

I can of course download the file manually but it would be a much smoother experience to use something like that:
```javascript
ctx.replyWithMediaGroup([
    {
        type: "video",
        media: new ReuploadFile("https://dev.nitra.pl/no-audio.mp4"),
    },
    {
        type: "video",
        media: new ReuploadFile("https://dev.nitra.pl/yes-audio.mp4"),
    },
])
```

KnorpelSenf:
Adds the following syntax for constructing `InputFile`s based on remote data.

```ts
const url = 'https://raw.githubusercontent.com/grammyjs/website/main/logos/grammY.png'

// both work
new InputFile({ url })
new InputFile(new URL(url))
```

The data will be fetched lazily from the specified URL as soon as the `InputFile` is sent. The file contents are streamed, so only a small portion of the data is kept in memory at all times. When several files are sent in an album, they are piped though one by one, such that only a single outgoing request is open at all times.

Currently does not support more advanced configuration options. Note that it's already possible to do this:

```ts
const { body } = await fetch(url, opts)
if (body === null) throw new Error('No data')
const inputFile = new InputFile(body)
```

This allows you to pass any configuration to `fetch`.

Currently does not support caching files on disk. This is very hard to get right, and it is questionable if it has any advantages. Confer the discussion at #33 for more information about caching.

Based off of #77. Please review and merge #77 first.

Closes #33.

### DIFF
--- a/src/platform.deno.ts
+++ b/src/platform.deno.ts
@@ -95,16 +95,7 @@ export class InputFile {
         if (
             typeof data === "object" && ("url" in data || data instanceof URL)
         ) {
-            data = (async function* () {
-                let url = data instanceof URL ? data : data.url;
-                const { body } = await fetch(url);
-                if (body === null) {
-                    throw new Error(
-                        `Download failed, no response body from '${url}'`,
-                    );
-                }
-                yield* body;
-            })();
+            data = fetchFile(data instanceof URL ? data : data.url);
         } else if (
             typeof data !== "string" && (!(data instanceof Uint8Array))
         ) {
@@ -112,6 +103,16 @@ export class InputFile {
         }
         return data;
     }
+}
+
+async function* fetchFile(url: string | URL): AsyncIterable<Uint8Array> {
+    const { body } = await fetch(url);
+    if (body === null) {
+        throw new Error(
+            `Download failed, no response body from '${url}'`,
+        );
+    }
+    yield* body;
 }
 
 // === Export InputFile types
--- a/src/platform.deno.ts
+++ b/src/platform.deno.ts
@@ -34,6 +34,15 @@ export const streamFile = isDeno
 // === Base configuration for `fetch` calls
 export const baseFetchConfig = {};
 
+/** Something that looks like a URL. */
+interface URLLike {
+    /**
+     * Identifier of the resouce. Must be in a format that can be parsed by the
+     * URL constructor.
+     */
+    url: string;
+}
+
 // === InputFile handling and File augmenting
 // Accessor for file data in `InputFile` instances
 export const inputFileData = Symbol("InputFile data");
@@ -65,6 +74,8 @@ export class InputFile {
     constructor(
         file:
             | string
+            | URL
+            | URLLike
             | Uint8Array
             | ReadableStream<Uint8Array>
             | AsyncIterable<Uint8Array>,
@@ -80,8 +91,23 @@ export class InputFile {
         if (this.consumed) {
             throw new Error("Cannot reuse InputFile data source!");
         }
-        const data = this.fileData;
-        if (typeof data !== "string" && (!(data instanceof Uint8Array))) {
+        let data = this.fileData;
+        if (
+            typeof data === "object" && ("url" in data || data instanceof URL)
+        ) {
+            data = (async function* () {
+                let url = data instanceof URL ? data : data.url;
+                const { body } = await fetch(url);
+                if (body === null) {
+                    throw new Error(
+                        `Download failed, no response body from '${url}'`,
+                    );
+                }
+                yield* body;
+            })();
+        } else if (
+            typeof data !== "string" && (!(data instanceof Uint8Array))
+        ) {
             this.consumed = false;
         }
         return data;
--- a/src/platform.deno.ts
+++ b/src/platform.deno.ts
@@ -46,7 +46,8 @@ export const inputFileData = Symbol("InputFile data");
  * Reference](https://core.telegram.org/bots/api#inputfile).
  */
 export class InputFile {
-    public readonly [inputFileData]: ConstructorParameters<typeof InputFile>[0];
+    private consumed = false;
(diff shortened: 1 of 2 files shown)


Loskir:
What's the benefit of using `AsyncIterable` compared to `Promise`?

KnorpelSenf:
Those are two very different things. What are you referring to?

src/platform.deno.ts
```diff
@@ -64,17 +74,45 @@ export class InputFile {
...
+
+async function* fetchFile(url: string | URL): AsyncIterable<Uint8Array> {
+    const { body } = await fetch(url);
+    if (body === null) {
+        throw new Error(
+            `Download failed, no response body from '${url}'`,
+        );
+    }
+    yield* body;
 }
```
Loskir, line 116: I mean why use `AsyncIterator` here instead of `Promise`?

But I see, in node it's actually used to yield chunks one by one instead of the whole thing

KnorpelSenf: I fail to see how a `Promise` relates to handling data streams

KnorpelSenf: Can you provide example code that streams the request body using a `Promise` instead of an iterable?

Loskir: Nevermind, I'm just used to `fetch` returning a promise where `body` is just a value, not a stream

src/platform.node.ts
```diff
@@ -52,15 +62,54 @@ export class InputFile {
...
+            typeof data !== "string" && (!(data instanceof Uint8Array))
+        ) {
+            this.consumed = false;
+        }
+        return data;
+    }
+}
+
+async function* fetchFile(url: string | URL): AsyncIterable<Uint8Array> {
+    const { body } = await fetch(url);
```
dzek69, line 99: is `fetch` global? i don't see where it come from

KnorpelSenf: Only on Deno and the web. On Node, it is shimmed by deno2node using node-fetch.

KnorpelSenf: Maybe it should be improved in deno2node that the shimming isn't performed for Node-specific source files.

dzek69: I don't know anything about deno2node. Just looking at the file as a whole from Node.js dev perspective i see `fetch` that is not imported from anywhere so it will crash until `fetch` is added to globals

KnorpelSenf: I can only do that after the changes to deno2node. Right now, the shimming will produce clashing imports, hence causing the compilation to fail.

KnorpelSenf: Ah I see. No, shimming in deno2node works by injecting import statements into the modules that actually need the shim. It does not touch global definitions.

I wrote a new README in https://github.com/wojpawlik/deno2node/pull/14 a long time ago but it has not been merged yet. I improved this README now with respect to this misunderstanding. I hope we have a chance to get this through soon

cc @wojpawlik

src/platform.deno.ts
```diff
@@ -64,17 +74,45 @@ export class InputFile {
...
     ) {
-        this[inputFileData] = file;
+        this.fileData = file;
         if (filename === undefined && typeof file === "string") {
             filename = basename(file);
         }
         this.filename = filename;
     }
+    get [inputFileData]() {
+        if (this.consumed) {
```
dzek69, line 91: I looked at this PR again and:
- won't this break `auto-retry`? I mean the plugin won't be able to actually retry the request, because InputFile will be always consumed. 
- I don't see `this.consumed = true` anywhere

KnorpelSenf: > won't this break `auto-retry`?

No, because the `consumed` flag is only meant to be set when iterators or other non-reusable data sources are used.

> I don't see `this.consumed = true` anywhere

True. This was fixed in [`23dcf2a` (#77)](https://github.com/grammyjs/grammY/pull/77/commits/23dcf2affda5934e0d93bf2e055308a259612ea3) but it looks like some merge has reverted the change. There is a lot going on in the same files right now, this get a little difficult. Rest assured that it will work after #77.

KnorpelSenf: I think the problem was that we should not have merged this before #77, as I stated in the original description.

dzek69: Doesn't GitHub have something to block merges before something else is merged?

dzek69: "PR dependencies" or something like that. I'm not a heavy user of GitHub, so I don't really know

PR #81, opened 2021-11-07
Dynamically determine the type of ctx.match

KnightNiwrem:
```ts
bot.hears('a string', ctx => {
  const matches = ctx.match;
  console.log(matches?.toLowerCase()); // ok
  console.log(matches?.groups); // error
}).use(ctx => {
  const matches = ctx.match;
  console.log(matches?.toLowerCase()); // ok
  console.log(matches?.groups); // error
});

bot.hears(['a string'], ctx => {
  const matches = ctx.match;
  console.log(matches?.toLowerCase()); // ok
  console.log(matches?.groups); // error
});

bot.hears(/a regexp/, ctx => {
  const matches = ctx.match;
  console.log(matches?.toLowerCase()); // error
  console.log(matches?.groups); // ok
});

bot.hears([/a regexp/], ctx => {
  const matches = ctx.match;
  console.log(matches?.toLowerCase()); // error
  console.log(matches?.groups); // ok
});

bot.hears(['a mixed array', /of string and regexp/], ctx => {
  const matches = ctx.match;
  console.log(matches?.toLowerCase()); // error
  console.log(matches?.groups); // error

  if (typeof matches === "string") {
    console.log(matches?.toLowerCase()); // ok
    console.log(matches?.groups); // error
  } else {
    console.log(matches?.toLowerCase()); // error
    console.log(matches?.groups); // ok
  }
});
```


src/composer.ts
```diff
@@ -303,18 +316,95 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
...
+                    string | RegExpMatchArray,
+                    ":text" | ":caption"
+                > => {
+                    const msg = ctx.message ?? ctx.channelPost;
+                    const txt = msg.text ?? msg.caption;
+                    return match(ctx, txt, triggerFns);
+                },
+                ...middleware,
+            );
+        }
```
KnorpelSenf, line 407: I don't think we should replicate the same code so many times. If TypeScript sucks at unifying things here, then we should type cast the arguments explicitly.

src/composer.ts
```diff
@@ -464,14 +554,87 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
...
+        trigger: MaybeNonEmptyArray<string | RegExp>,
+        ...middleware: Array<
+            Middleware<
+                FilteredMatchContext<
+                    C,
+                    string | RegExpMatchArray,
+                    "callback_query:data"
+                >
+            >
+        >
```
KnorpelSenf, line 600: Is there a way to make this less verbose? I'm as uncomfortable as you probably are with half of the code in this file being type annotations of function overloads.

src/composer.ts
```diff
@@ -820,28 +1132,42 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
...
+type RegExpExecFn = (content: string) => RegExpExecArray | null;
+
+function isNonEmptyStringArray(xs: any): xs is NonEmptyArray<string> {
+    return Array.isArray(xs) && xs.length > 0 &&
+        xs.every((x) => typeof x === "string");
+}
+function isNonEmptyRegExpArray(xs: any): xs is NonEmptyArray<RegExp> {
+    return Array.isArray(xs) && xs.length > 0 &&
+        xs.every((x) => x instanceof RegExp);
+}
```
KnorpelSenf, line 1145: Why is this distinction even necessary? It worked well before, it is just for type annotations?

src/composer.ts
```diff
@@ -820,28 +1132,42 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
...
 }
 
-type HearsContext<C extends Context> = Filter<
-    C & { match: string | RegExpMatchArray },
-    ":text" | ":caption"
->;
 type CommandContext<C extends Context> = Filter<
     C & { match: string },
     ":entities:bot_command"
 >;
```
KnorpelSenf, line 1159: Can be expressed via `FilteredMatchContext` now.

src/context.ts
```diff
@@ -74,7 +74,7 @@ export class Context implements RenamedUpdate {
      * Used by some middleware to store information about how a certain string
      * or regular expression was matched.
      */
-    public match?: string | RegExpMatchArray | null;
+    public match?: string | RegExpMatchArray;
```
KnorpelSenf, line 77: Nice catch!

src/filter.ts
```diff
@@ -358,7 +358,7 @@ type CollapseL2<
     : never
         : never);
 // All queries
-type AllValidFilterQueries = InjectShortcuts;
+export type AllValidFilterQueries = InjectShortcuts;
```
KnorpelSenf, line 361: Use `FilterQuery` instead of `AllValidFilterQueries`. This helper type was introduced to prevent type alias expansion in editor tooling, and to make the API reference more readable. `FilterQuery` also has docs.

KnorpelSenf (commented):
You're calling it the verbose edition—is there a second PR that is less verbose, or what does this name imply?

Generally it looks like a good way to tackle this problem. However, I don't like at all how verbose things get because of this. I'm against merging unless we manage to cut it down.

KnightNiwrem:
> You're calling it the verbose edition—is there a second PR that is less verbose, or what does this name imply?
> 
> Generally it looks like a good way to tackle this problem. However, I don't like at all how verbose things get because of this. I'm against merging unless we manage to cut it down.

Agreed. There isn't a second PR that is less verbose.. or at least there _wasn't_. But I've managed to find a way to avoid code duplication for now, at the expense of longer type parameters

src/composer.ts
KnightNiwrem: Agreed. Trying to avoid casting right now though.

src/composer.ts
KnightNiwrem: Yup. It's no longer required, and deleted.

src/composer.ts
```diff
@@ -1,8 +1,20 @@
 import { Context } from "./context.ts";
-import { Filter, FilterQuery, matchFilter } from "./filter.ts";
+import {
```
KnorpelSenf: Please revert

src/composer.ts
```diff
@@ -302,16 +314,45 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
...
      * @param middleware The middleware to register
      */
-    hears(
-        trigger: MaybeArray<string | RegExp>,
-        ...middleware: Array<Middleware<HearsContext<C>>>
-    ): Composer<HearsContext<C>> {
-        const trg = triggerFn(trigger);
+    hears<
+        T extends string = string,
+        U extends FilteredMatchContext<C, T, ":text" | ":caption"> =
```
KnorpelSenf, line 315: I don't think this is correct. Now, you can pass arbitrary context types that fulfill the `extends` constraint. The result would promise to generate the necessary predicate for that, which it does not.

src/composer.ts
```diff
@@ -302,16 +314,45 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
...
+            FilteredMatchContext<C, T, ":text" | ":caption">,
+    >(
+        trigger: MaybeNonEmptyArray<string | RegExp>,
+        ...middleware: Array<Middleware<U>>
+    ) {
+        const triggers = toNonEmptyArray(trigger);
+        const triggerFns = triggers.map((t) => triggerFn(t));
         return this.on([":text", ":caption"]).filter(
-            (ctx): ctx is HearsContext<C> => {
+            (ctx): ctx is U => {
```
KnorpelSenf, line 348: Here we are promising to guarantee anything that ensures `U`, which isn't true.

KnorpelSenf (commented):
Less lines, but now we're having wrong type annotations.

KnorpelSenf:
I need some more time to investigate this. Kinda want to see this merged, but still trying to make it more readable. Novice TS users need to read the type annotations and get at least an idea of what this does.

KnorpelSenf:
I appreciate your effort and I still want to see this problem solved, but the type annotations that are necessary to make this happen are getting out of hand. The annotations are not straightforward, and if we as developers of grammY think that, then it's already way to complicated for users, especially for the beginners among them. I do not support this change.

In addition, #100 is going to complicate the implementation further. Do we then need four different kinds of context types, four different kinds of middleware types, and all of that for all affected methods?

I am open to split `hears` and its sibling into two methods, such as `hears`/`hearsRegex`, or `hearsExact`/`hears`, where the former matches strings, and the latter matches regular expressions, respectively.

Alternatively, we may look into disallowing mixing strings and regex in array on the type level. If anyone needs this, they could type-case explicitly to make it work, as the runtime code can continue supporting it.

Either way, the change would need to be backed by the community. For example, if people mind having multiple almost identical methods more than minding the small inaccuracy in types, then we obviously will not touch the current code. Moreover, both suggestions are breaking changes, so they would have to wait until the next major release.

Either way, I am going to close this PR for now. We may come back to it if anyone can think of a novel way to solve this, or if the tooling changes (who knows what TS does next).

Thank you again for this attempt!

PR #82, opened 2021-11-07
Replace stateful regex.exec(str) with str.match(regex)

darvesh:


KnorpelSenf (approved):
Amazing catch, thanks!

KnorpelSenf:
@all-contributors add @darvesh for bug and code

PR #84, opened 2021-11-09
Added missing pre, text_link, text_mention entity types to filters

dcdunkan:
There was 3 message entity types missing (https://core.telegram.org/bots/api#messageentity) in src/filters.ts. `pre`, `text_link`, `text_mention`. Added those.


KnorpelSenf (approved):
Thanks!

KnorpelSenf:
@all-contributors add dcdunkan for bug and code

PR #87, opened 2021-11-09
Support file uploads via Deno.open()

KnorpelSenf:
Adds support for `Blob` under Deno.

Adds support for `Deno.File` under Deno. This allows for:
```ts
new InputFile(await Deno.open('/path/to/file'))
```

Also adds support for passing a `Promise` of any value under both Deno and Node.js. This allows for:
```ts
new InputFile(Deno.open('/path/to/file'))
```
under Deno, as well as
```ts
import { promises as fs } from "fs";
new InputFile(fs.readFile("asdf"));
```
under Node.

Note that this change will be integrated into #77 in order to prevent conflicts.

### DIFF
--- a/src/platform.deno.ts
+++ b/src/platform.deno.ts
@@ -69,6 +69,7 @@ export class InputFile {
     constructor(
         file:
             | string
+            | Blob
             | URL
             | URLLike
             | Uint8Array
@@ -87,7 +88,7 @@ export class InputFile {
             throw new Error("Cannot reuse InputFile data source!");
         }
         const data = this.fileData;
-        // Handle file paths
+        // Handle local files
         if (typeof data === "string") {
             if (!isDeno) {
                 throw new Error(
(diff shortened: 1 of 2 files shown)


KnorpelSenf:
@KnightNiwrem maybe you should also take a look at this. I'm not sure if there's something wrong about supporting `Promise`s for everything.

src/platform.deno.ts
```diff
@@ -116,6 +122,9 @@ async function* fetchFile(url: string | URL): AsyncIterable<Uint8Array> {
     }
     yield* body;
 }
+function isDenoFile(data: unknown): data is Deno.File {
+    return isDeno && data instanceof Deno.File;
```
wojpawlik, line 126: This is file is Deno-specific, why the runtime check?
```suggestion
    return data instanceof Deno.File;
```

wojpawlik (approved):
> ```ts
> import { promises as fs } from "fs";
> new InputFile(fs.readFile("asdf"));
> ```
>
> under Node.

`new InputFile(fs.createReadStream("asdf"))` is preferable, unless the `InputFile` instance will be reused.

src/platform.deno.ts
KnorpelSenf: This file is web-specific. This does coincide with Deno (hence the name), but it will also be bundled for browsers.

wojpawlik:
Relevant: https://deno.com/blog/v1.16#fetch-now-supports-fetching-file-urls

PR #94, opened 2021-11-21
Allow URLs as strings for InputFile

KnorpelSenf:
Changes the behaviour of `new InputFile(string)`. Tries now to guess if the given value is a URL or a string, and handles the resources accordingly. The guessing is performed by attempting to construct `new URL(string)`. If this succeeds and either we are running on Deno or the resulting URL object does not have `file:` as protocol, the URL object is passed to `fetch`.

Adds `new InputFile({ path: '/path/to/file' })` which will always interpret the given string as a local file path.


KnorpelSenf:
Not that this is another PR this will be integrated into #77.

KnorpelSenf:
Suggestion: instead of introducing ambiguity, always print to debug which resource type is attempted to be consumed. For example,
```ts
debug('Uploading by string, taking it as local file path')
```
This avoids the ambiguity while still being transparent about why passing a URL as string isn't doing what people may think it does.

PR #111, opened 2021-11-25
ES6 bundling

KnorpelSenf:
https://get.grammy.dev/es6@dev.js but somehow it is not compatible with Deno yet.

@trgwii do you know how to fix the build script?

I'd like this to become possible in both browsers and Deno env:
```ts
import { Bot } from 'https://get.grammy.dev/es6@dev.js'
// or
// import { Bot } from 'https://get.grammy.dev/es6@v1.4.2.js'

const bot = new Bot("");
// etc
```
Uses https://github.com/denoland/deno/issues/9638#issuecomment-790282221


KnorpelSenf:
You may push to the branch and inspect the Netlify logs, but **please take care**! Build minutes are not infinite. It is important that you develop and test locally, and only push to this branch when you're confident that the solution works.

KnorpelSenf:
We need to figure out how we can bundle things on GitHub and then publish on Netlify. It is financially not feasible to perform the bundling on Netlify's infra for every CI build.

KnorpelSenf:
I stopped the builds for now. However, all currently bundled versions remain accessible on `get.grammy.dev`.

KnorpelSenf:
Waiting for feeback on https://github.com/nwtgck/actions-netlify/issues/689

bundling/netlify.toml
```diff
@@ -0,0 +1,12 @@
+[[headers]]
+for = "/*"
+[headers.values]
+Access-Control-Allow-Origin = "*"
+
+[[redirects]]
+force = true # ensure that we always redirect
+from = "/"
```
nwtgck, line 8: I came from https://github.com/nwtgck/actions-netlify/issues/689. 

Could you try `from = "/*"` instead?

image
https://docs.netlify.com/configure-builds/file-based-configuration/

KnorpelSenf: See 5b58b8897979249e937b994f641a15c52ddf3954. Same issue.

KnorpelSenf: Now that I think about it, that's actually not the desired configuration. I would like all assets to be served from the URLs that I upload them to.

The redirect was simply added because if anyone visits https://get.grammy.dev with a browser (instead of loading the bundled assets from code), they should be taken to https://grammy.dev.

If there's a chance that the redirects can cause any problems, then we can just drop them for now. This is what I did in fc29e32424252516adefd859856bcac653546ca8 but the problem persists.

nwtgck:
We still have the error with simple netlify.toml...

KnorpelSenf:
It is only necessary when downloading the scripts from the CDN later. I can easily remove the file until we can find out how to make the deploys work.

KnorpelSenf:
Same story …

nwtgck:
Could you confirm the `secrets.NETLIFY_SITE_ID` and `secrets.NETLIFY_AUTH_TOKEN` do not contain white spaces or other unescaped characters? I guess they might be used in query parameter and cause the error.

KnorpelSenf:
There was some whitespace in the site identifier. Thank you so much for your help!

nwtgck:
Good!

KnorpelSenf:
Check out https://61a6300e419e94174861e1b7--distracted-varahamihira-adb7a7.netlify.app/es6@v1.4.2.js

KnorpelSenf:
It looks like Deno does not output ES6 code. The `?.` operator was introduced in ES2020 and should be transpiled. I will investigate this.

KnorpelSenf:
What a hassle. Now it's finally working. Let's hope the next Deno version does not break it again.

It would be good to have a solution where already bundled modules are never touched again. The current solutoin may corrupt old files on the CDN that people rely on. This may be acceptable at this experimental stage, but it would be good to have that fixed in the long run. Contributions and ideas are welcome.

PR #112, opened 2021-11-28
Support synchronous iterables

KnorpelSenf:
This is already working, all we need to do is to permit it in the type annotations.

What's more, we're already expecting this to work in the docs: https://grammy.dev/guide/files.html#uploading-your-own-files


KnightNiwrem (approved):

KnorpelSenf:
Fastest PR in the world

PR #116, opened 2021-11-30
Support more bundling options

KnorpelSenf:
Adds support for ES3, ES5, and ESNext bundling


KnorpelSenf:
Available formats are now following `https://get.grammy.dev/es<ES version>@<grammY version>.js`.

Examples:

1.0 release in ES3: https://get.grammy.dev/es3@v1.0.0.js
Current stable in ES6: https://get.grammy.dev/es6@v1.4.2.js
Current HEAD in ESNext: https://get.grammy.dev/esnext@dev.js

KnorpelSenf:
Actually, ES3 and ES5 bundling does not make any sense, and these versions have never worked for anyone. Internally, grammY uses a `Proxy` instance for every single API request, which only exists since ES6. It makes no sense at all to provide those old versions.

PR #120, opened 2021-12-03
[chore]: Edit name of contributor

habemuscode:
Small PR in order to edit my contributors name


KnorpelSenf (approved):
The tool picked your name because that's what you published on your GitHub profile. If you prefer not to have it public, you may also want to consider changing it there, too.

habemuscode:
@KnorpelSenf Thanks, changed in my profile

PR #126, opened 2021-12-07
Add support for Bot API 5.5

KnorpelSenf:
Adds support for Bot API 5.5.

The only _interesting_ grammy-specific change is that we can now do:
```ts
bot.on('message:is_automatic_forward')
```
and it will trigger for all messages in a group that were posted in the attached channel.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -957,7 +957,7 @@ export class Context implements RenamedUpdate {
      */
     banChatSenderChat(
         sender_chat_id: number,
-        other?: Other<"banChatSenderChat", "user_id">,
+        other?: Other<"banChatSenderChat", "sender_chat_id">,
         signal?: AbortSignal,
     ) {
         return this.api.banChatSenderChat(
--- a/src/context.ts
+++ b/src/context.ts
@@ -950,17 +950,20 @@ export class Context implements RenamedUpdate {
      * Context-aware alias for `api.banChatSenderChat`. Use this method to ban a channel chat in a supergroup or a channel. The owner of the chat will not be able to send messages and join live streams on behalf of the chat, unless it is unbanned first. The bot must be an administrator in the supergroup or channel for this to work and must have the appropriate administrator rights. Returns True on success.
      *
      * @param sender_chat_id Unique identifier of the target sender chat
+     * @param other Optional remaining parameters, confer the official reference below
      * @param signal Optional `AbortSignal` to cancel the request
      *
      * **Official reference:** https://core.telegram.org/bots/api#banchatsenderchat
      */
     banChatSenderChat(
         sender_chat_id: number,
+        other?: Other<"promoteChatMember", "user_id">,
         signal?: AbortSignal,
     ) {
         return this.api.banChatSenderChat(
             orThrow(this.chat, "banChatSenderChat").id,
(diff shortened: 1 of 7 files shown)


habemuscode (approved):
LGTM

src/context.ts
```diff
@@ -946,6 +946,44 @@ export class Context implements RenamedUpdate {
...
 
+    /**
+     * Context-aware alias for `api.banChatSenderChat`. Use this method to ban a channel chat in a supergroup or a channel. The owner of the chat will not be able to send messages and join live streams on behalf of the chat, unless it is unbanned first. The bot must be an administrator in the supergroup or channel for this to work and must have the appropriate administrator rights. Returns True on success.
+     *
+     * @param sender_chat_id Unique identifier of the target sender chat
+     * @param signal Optional `AbortSignal` to cancel the request
+     *
+     * **Official reference:** https://core.telegram.org/bots/api#banchatsenderchat
+     */
+    banChatSenderChat(
```
KnightNiwrem, line 957: This api should come with until_date options

src/core/api.ts
```diff
@@ -884,6 +884,43 @@ export class Api<R extends RawApi = RawApi> {
...
+    /**
+     * Use this method to ban a channel chat in a supergroup or a channel. The owner of the chat will not be able to send messages and join live streams on behalf of the chat, unless it is unbanned first. The bot must be an administrator in the supergroup or channel for this to work and must have the appropriate administrator rights. Returns True on success.
+     *
+     * @param chat_id Unique identifier for the target chat or username of the target channel (in the format @channelusername)
+     * @param sender_chat_id Unique identifier of the target sender chat
+     * @param signal Optional `AbortSignal` to cancel the request
+     *
+     * **Official reference:** https://core.telegram.org/bots/api#banchatsenderchat
+     */
+    banChatSenderChat(
```
KnightNiwrem, line 896: Same here

KnightNiwrem (changes requested):

src/context.ts
KnorpelSenf: Thanks, nice catches!

src/context.ts
```diff
@@ -946,6 +946,47 @@ export class Context implements RenamedUpdate {
...
+     *
+     * @param sender_chat_id Unique identifier of the target sender chat
+     * @param other Optional remaining parameters, confer the official reference below
+     * @param signal Optional `AbortSignal` to cancel the request
+     *
+     * **Official reference:** https://core.telegram.org/bots/api#banchatsenderchat
+     */
+    banChatSenderChat(
+        sender_chat_id: number,
+        other?: Other<"banChatSenderChat", "user_id">,
```
KnightNiwrem, line 960: `user_id` is not a parameter for `banChatSenderChat`, so the other object would simply be all the parameters of `banChatSenderChat` (probably).

KnightNiwrem (changes requested):

src/context.ts
KnorpelSenf: Ohhhh now I get you. How can I struggle so much copying and pasting stuff? Thanks!

KnightNiwrem (approved):

PR #129, opened 2021-12-09
Fix throwing error upon `bot.stop()`

KnorpelSenf:
Fixes errors reported in https://t.me/grammyjs/16197 and https://t.me/grammyjs/28217, namely that `bot.stop()` caused the built-in polling to throw an error.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -425,14 +425,10 @@ you can circumvent this protection against memory leaks.`);
     async stop() {
         if (this.pollingRunning) {
             debug("Stopping bot, saving update offset");
-            debug("+++++++++++++++++++++ aborting ");
             this.pollingAbortController?.abort();
             this.pollingRunning = false;
-            debug("+++++++++++++++++++++ incrementing ");
             const offset = this.lastTriedUpdateId + 1;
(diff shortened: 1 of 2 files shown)


KnorpelSenf:
It is very easy to see that this PR fixes the issue. If you run
```ts
import { Bot } from 'https://deno.land/x/grammy@1.5.2/mod.ts'

const bot = new Bot(Deno.env.get('BOT_TOKEN') ?? '')

bot.on('message:text', ctx => ctx.reply(ctx.msg.text))

bot.start()

await new Promise(r => setTimeout(r, 5000))

await bot.stop()
```
then it will crash, but if you run
```ts
import { Bot } from 'https://raw.githubusercontent.com/grammyjs/grammY/4e7bb8b1a362e8f228ce22e3ca5efad208e804a1/src/mod.ts'

const bot = new Bot(Deno.env.get('BOT_TOKEN') ?? '')

bot.on('message:text', ctx => ctx.reply(ctx.msg.text))

bot.start()

await new Promise(r => setTimeout(r, 5000))

await bot.stop()
```
then everything works fine.

KnorpelSenf:
Taking https://t.me/grammyjs/28297 as an accepting review.

PR #130, opened 2021-12-10
Add `readAll` and internal data access to memory session

KnorpelSenf:
Title says it all. Requested in https://t.me/grammyjs_ru/101

### DIFF
--- a/src/convenience/session.ts
+++ b/src/convenience/session.ts
@@ -361,6 +361,17 @@ export class MemorySessionStorage<S> implements StorageAdapter<S> {
         return value.session;
     }
 
+    /**
+     * Reads the values for all keys of the session storage, and returns them as
+     * an array.
+     */
+    readAll() {
+        return Array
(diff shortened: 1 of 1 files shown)


Satont:
It looks cool.

In the same discussion, we had the idea of adding a method to get all the sessions. Can we add this in the same PR?

KnorpelSenf:
Yes, we can add that.

I'm also thinking about allowing

```ts
for await (const data of ctx.session) {
  // handle session data of user
}
```

which would effectively allow you to enumerate all sessions.

This would be optional to implement for storage adapters. Hence, it is a non-breaking change. Storage adapters that do not support this will simply throw an error when this is attempted.

KnorpelSenf:
@Satont please review.

Satont:
Looking good, thank you. 👍

KnorpelSenf:
Taking that as an accepted review, merging.

PR #133, opened 2021-12-13
Fix error for command with a leading slash

glukki:
With `composer.command('/start')`, the expected error message is:
```Do not include '/' when registering command handlers (use 'start' not '/start')```
but instead I received
```Do not include '/' when registering command handlers (use '/' not '/start')```


KnorpelSenf (approved):
Nice catch, thanks for fixing it!

KnorpelSenf:
@all-contributors add @glukki for bug and code

PR #140, opened 2021-12-19
Skip typechecks while bundling old versions

KnorpelSenf:
When Deno (and thus TS) upgrades, this can make already released versions stop compiling. That just happened with TS 4.5. Such a case breaks our bundling script. This PR applies `--no-check` to all bundling operations, and it does so with the following reasons:
- All versions typecheck when released. This is enough safety.
- Bundles do not contain types, and since tsc transpilation JS output is stable over versions due to regression tests, we can safely ignore the types in the code.
- It takes a lot of time to check the same stuff over and over again, which slows down CI by a large factor.
- Bundling is broken and there are no other solutions that take this little effort and work that well.


KnorpelSenf:
Taking https://t.me/grammyjs/30528 as an approval.

PR #145, opened 2021-12-28
Do not send undefined values in form data

### LINKED ISSUE #144: Bug: undefined and unspecified have different behavior
I migrated yet another bot to grammy and found out that there is something fishy here.
I do not use transformers or something, its so far a plain Telegraf → grammY migration.

Technically `telegraf-inline-menu` → `grammy-inline-menu` is used which sets undefined to some properties instead of omitting them. I recreated the issue with plain grammY calls but haven't looked into it what exactly might cause this issue.

The goal is to send a photo without caption.

```typescript
await context.replyWithPhoto(media, {caption: undefined});
```

Ends up creating a photo message with a caption which is literally saying "undefined".

```typescript
await context.replyWithPhoto(media, {caption: undefined, parse_mode: undefined});
```

Fails with "Call to 'sendPhoto' failed! (400: Bad Request: unsupported parse_mode)"
Part of the error contains this, which indicated the error still thinks the `undefined` is `undefined` and not the string `"undefined"`:
```javascript
{
  method: 'sendPhoto',
  payload: {
    chat_id: 2956631,
    photo: 'attach://hyyujfn4azjcick6',
    caption: undefined,
    parse_mode: undefined
  },
  ok: false,
  error_code: 400,
  description: 'Bad Request: unsupported parse_mode',
  parameters: {}
}
```

I did not create a minimal bot to check these calls in a simple environment but as the bot is not that complex I dont think there are side effects to it currently. But I might be wrong here.

KnorpelSenf:
Closes #144


src/core/payload.ts
```diff
@@ -113,6 +113,7 @@ async function* payloadToMultipartItr(
     const separator = enc.encode(`\r\n--${boundary}\r\n`);
     let first = true;
     for (const [key, value] of Object.entries(payload)) {
+        if (value == null) continue;
```
EdJoPaTo, line 116: Without trying this out… `parse_mode: undefined` shouldnt match `null`?

Maybe I overlooked something but is this working recursive into objects?

KnorpelSenf: > `parse_mode: undefined` shouldnt match `null`?

```
< undefined === null
> false
< undefined == null
> true
```

Welcome to JS.

> Maybe I overlooked something but is this working recursive into objects?

https://github.com/grammyjs/grammY/blob/446a02471594572ff253a43a01e174b4971a6d37/src/core/payload.ts#L117 calls https://github.com/grammyjs/grammY/blob/446a02471594572ff253a43a01e174b4971a6d37/src/core/payload.ts#L30 which already replaces `null` values by `undefined`, which will be removed during serialisation.

EdJoPaTo: > Welcome to JS.

Readable code is fun :D

KnorpelSenf: It's consistent with the rest of the ecosystem. It's also less verbose than `value === undefined || value === null` but I do see that point that it could be more readable to use the long version.

EdJoPaTo (approved):
Just tested, this fix works :sparkles:

PR #146, opened 2021-12-28
Update deno2node

KnorpelSenf:
Follow-up for #143

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
         "@types/node": "^12.20.36",
         "@types/node-fetch": "^2.5.12",
         "all-contributors-cli": "^6.20.0",
-        "deno2node": "^1.0.0"
+        "deno2node": "^1.1.0"
     },
     "files": [
         "out/"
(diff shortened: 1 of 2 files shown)


KnorpelSenf:
@wojpawlik could you investigate why CI fails? I do not understand how version 1.0.0 and 1.1.0 differ in a way that prevents grammY from compiling. Does it somehow use updated Node.js types or so? It looks like the `fetch` types are updated.

wojpawlik:
I can only imagine TypeScript upgrade breaking `@types/node-fetch`, but I don't see it reported at DefinitelyTyped. https://github.com/fromdeno/deno2node/compare/v1.0.0...v1.1.0

KnorpelSenf:
@wojpawlik thanks for the link, that helped. Please review again.

We removed shims from Node-specific files. That's why TS assumed DOM types, rather than using fetch from `node-fetch`. It was fixed by importing `fetch` explicitly. It is much better now.

KnightNiwrem (approved):

Issue #108, opened 2021-11-23
plugin: BotCommandScope builder


KnorpelSenf:
Create a new repo with a plugin for advanced command filtering that goes beyond what people usually need. This not only allows users to do cooler things, but it also helps the core package stay focused on other things.

Desired features:
- should work like menu plugin, but for command filtering
- argument matching
- `ctx.command` property that exposes the invoked command on the context object (useful for what we currently do with `bot.command(array)` where it is not possible to find out which command was invoked without inspecting the message text)
- custom command prefixes, such as `!` instead of `/`
- optional case insensitivity
- plugin should have the name `commands`

EdJoPaTo:
It might be interesting to export the commands to [setMyCommands](https://core.telegram.org/bots/api#setmycommands).
Not sure if that's easily possible as they also require a description.

Some kind of check would be great to ensure only commands are set which are actually implemented in the bot.

KnorpelSenf:
I was thinking to have a method `syncCommands` which internally calls `setMyCommands` with all the right scopes set. That would be the point of the command builder, to unify setting the commands and registering listeners for them. This is analogous to the menu plugin in the way that the latter combines building inline keyboard buttons and handlers, and the commands plugin would combine defining commands and their handlers.

AndreoliBR:
If I well undestood, this can use the following syntax...?

```ts
bot.command
    .name("start")
    .description("Welcomes first-time users")
    .filter(/* Do filtering stuff */)
    .defaultDo((ctx) => { /* Everyone */ })
    .onPrivateChat((ctx) => { /* Only on PM */ })
    .onGroupAdmins((ctx) => { /* Only group admins */ })
```

KnorpelSenf:
I do not have a concrete plan for this yet. You're more than welcome to make suggestions!

Telegram often has matching concepts that are at different places in the API, where it is convenient to bring things together. Inline keyboards are sent, and callback queries are received. The menu plugin brings this together by letting you define handlers directly alongside your inline keyboard.

The commands plugin could bring together two other things, namely setting command scopes and handling command messages. That's as far as I have planned, the rest are good ideas that may or may not fit into the cohesive vision that we need for this plugin.

dcdunkan:
Feature suggestion: "Did you mean ...?" for misspelled commands. It could also have a configurable error-help message like we usually see when subcommands are misspelled in CLI applications.

waptik:
[...]

PR #147, opened 2021-12-29
Do not handle commands from channels

KnorpelSenf:
They don't exist anyway (the server does not parse the command entities in channels) so we might as well drop this.

#108 will add command handling that does not rely on parsed entities.

This PR has the advantage that `ctx.from` will now always be present after `bot.command`.


src/composer.ts
```diff
@@ -384,24 +384,22 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
...
             }
             const set = cmd.indexOf("@") === -1 ? noAtCommands : atCommands;
             set.add(cmd);
         });
-        return this.on(":entities:bot_command").filter(
+        return this.on("message:entities:bot_command").filter(
             (ctx): ctx is CommandContext<C> => {
-                const msg = ctx.message ?? ctx.channelPost;
-                const txt = msg.text ?? msg.caption;
-                const entities = msg.entities ?? msg.caption_entities;
```
EdJoPaTo, line 398: This not only removes support for commands in channels but also for commands on media messages with captions?

Not sure if that even works normally, never tried that.

KnorpelSenf: grammY used to support commands in caption in its early days, but a lot of people in the community complained that it is not intuitive for command handlers to trigger when e.g. an image caption contains a command. Hence, the support was remove by changing the filter query from `::bot_command` to `:entities:bot_command`.

The code that was removed here was essentially dead code.

joseanoxp: You can mix it, it's perfect!

KnorpelSenf: Mix? Do you mean merge?

joseanoxp (approved):
Get in gear!

roj1512 (approved):
The changes are doing what they are meant to. Commands aren't handled in channels, anymore.

Tecardo1 (approved):

KnorpelSenf:
> The changes are doing what they are meant to do. Commands aren't handled in channels, anymore.

Were commands handled in channels before?

KnorpelSenf:
Oh well. I tested this once again, and interestingly enough the entities are not rendered in any client, but bots receive them. Hence, the assumption that _the server does not parse the command entities in channels_ is wrong. As a result, we cannot drop this feature without doing a breaking change. I will close this unmerged.

Note that if you still want `ctx.from` to be present, you can do
```ts
bot.on('message').command('start', ctx => {
  const name = ctx.from.first_name // `ctx.from` is present now!
})
```

EdJoPaTo:
The dead code you mentioned about the commands on captions should still be removed I guess?

KnorpelSenf:
https://github.com/grammyjs/grammY/commit/e322af1c91b8ec94d2959201bf094a53b3b0bad2

Issue #136, opened 2021-12-16
grammY conversations (Why Scenes Are Bad)


KnorpelSenf:
Admittedly, that title was clickbait. It seems like you are interested in scenes or wizards, and other programming patters that allow you to define conversational interfaces like they were a finite-state machine (FSM, [wiki](https://en.wikipedia.org/wiki/Finite-state_machine)).

> THIS IS A  A LOT OF TEXT. Please still read everything before you comment. Let's try to keep the signal-to-noise ratio up high on this one :)

# What Is This Issue

One of the features that get requested most often are scenes. This issue shall:

1. Bring everyone on the same page what people mean when they say scenes.
2. Why there is not going to be a traditional implementation of scenes for grammY, and how we're trying to do better.
3. Update you about the current progress.
4. Introduce two novel competing concepts that could both turn out as better than scenes.
5. Serve as a forum to discuss where we want to take this library regarding grammY conversations/scenes.

# 1. What are scenes?

A chat is a conversational interface. This means that the chat between the user and the bot evolves over time. Old messages stay relevant when processing current ones, as they provide the context of the conversation that determines how to interpret messages.

```
< /start
>>> How old are you?
< 42
>>> Cool, how old is your mother?
< 70
>>> Alright, she was 28 when you were born!
```

Note how the user sends two messages, and both are numbers. We only know that those two numbers mean two different things because we can follow the flow of the conversation. The two age numbers are following up two different questions. Hence, in order to provide a natural conversational flow, we must store the history of the chat, and take it into account when interpreting messages.

> Note that Telegram does not store the chat history for bots, so you should store them yourself. This is often done via sessions, but you can also use your own database.

In fact, we often don't need to know the entire chat history. The few most recent messages are often enough to remember, as we likely don't have to care about what the user sent back in 2018. It is therefore common to construct _state_, i.e. a small bit of data that stores where in the conversation where are. In our example, we would only need to store if the last question was about the age of the user, or about the age of their mother.

Scenes are a way to express this conversational style by allowing you to define a finite-state machine. Please [google](https://www.google.com/search?q=finite-state+machine) what this is, it is essential for the following discussion. The state is usually stored in the session data. They achieve this by isolating a part of the middleware into a block that can be entered and left.

Different bot frameworks have different syntax for this, but it typically works roughly like this (explanatory code, do not try to run):

```ts
// Define a separate part of the middleware handling.
const scene = new Scene('my-scene')
[...]

PR #152, opened 2022-01-03
PROOF OF CONCEPT; DO NOT MERGE: Implement else

KnorpelSenf:
Explanation for https://github.com/grammyjs/grammY/issues/136#issuecomment-1004013705


KnorpelSenf:
If we ever figure that we do indeed want to merge this, we would need to add a protection against doing
```ts
bot.filter()
bot.use()
bot.else()
```
because that will be super confusing, as the `else` call and the `use` call are actually going to be inverted.

KnorpelSenf:
This only works with `filter` and `on`. It does not work with `command` or `hears` or any other composer method that uses multi-level filtering.

PR #158, opened 2022-01-17
Implement cross-level twins

### LINKED ISSUE #157: fix: Cross-level twin typing
The filter queries have a feature called twin properties, which will make some properties guaranteed if other ones are present. For example, animation messages always have the document property specified, and the filter queries account for that.

So far, they only work on L2. However, there are fields such as `from` (only absent in channel post messages) or `edit_date` (only absent for new messages) which need to inspect L1 to in order to make an assumption about L2 properties. In other words, we cannot simply assume that the twin properties can be read from a flat list. Instead, we need to to build up nested objects structures that can be merged (intersected) into the result type in order to support twins that have dependencies across several levels.

KnorpelSenf:
Fixes #157.

Please review very thoroughly. This is a fragile part of the library, and it is nearly impossible to write automatic tests for it. Just throw it onto a large grammY code base and check that everything still compiles. I tried out a bunch of toy examples for testing purposes, but it would be nice to have an independent confirmation that this works.

### DIFF
--- a/src/filter.ts
+++ b/src/filter.ts
@@ -312,21 +312,23 @@ const UPDATE_KEYS = {
 // === Build up all possible filter queries from the above validation structure
 type KeyOf<T> = string & keyof T; // Emulate `keyofStringsOnly`
 
+// Suggestion building base structure
 type S = typeof UPDATE_KEYS;
 
-// E.g. 'message'
-type L1 = KeyOf<S>;
-// E.g. 'message:entities'
-type L2<K extends L1 = L1> = K extends unknown ? `${K}:${KeyOf<S[K]>}` : never;
-// E.g. 'message:entities:url'
-type L3<K0 extends L1 = L1> = K0 extends unknown ? L3_<K0> : never;
-type L3_<
-    K0 extends L1,
-    K1 extends KeyOf<S[K0]> = KeyOf<S[K0]>,
-> = K1 extends unknown ? `${K0}:${K1}:${KeyOf<S[K0][K1]>}` : never;
-// All three combined
-type L123 = L1 | L2 | L3;
-// E.g. 'message::url'
+// E.g. 'message' suggestions
+type L1S = KeyOf<S>;
+// E.g. 'message:entities' suggestions
+type L2S<L1 extends L1S = L1S> = L1 extends unknown ? `${L1}:${KeyOf<S[L1]>}`
+    : never;
+// E.g. 'message:entities:url' suggestions
+type L3S<L1 extends L1S = L1S> = L1 extends unknown ? L3S_<L1> : never;
+type L3S_<
+    L1 extends L1S,
+    L2 extends KeyOf<S[L1]> = KeyOf<S[L1]>,
+> = L2 extends unknown ? `${L1}:${L2}:${KeyOf<S[L1][L2]>}` : never;
+// Suggestions for all three combined
+type L123 = L1S | L2S | L3S;
+// E.g. 'message::url' generation
 type InjectShortcuts<Q extends L123 = L123> = Q extends
     `${infer R}:${infer S}:${infer T}`
     ? `${CollapseL1<R, L1Shortcuts>}:${CollapseL2<S, L2Shortcuts>}:${T}`
@@ -390,13 +392,13 @@ type NotUndefined = string | number | boolean | SomeObject;
 type RunQuery<Q extends string> = L1Discriminator<Q, L1Parts<Q>>;
 
 // gets all L1 query snippets
-type L1Parts<Q extends string> = Q extends `${infer U}:${string}` ? U : Q;
+type L1Parts<Q extends string> = Q extends `${infer L1}:${string}` ? L1 : Q;
 // gets all L2 query snippets for the given L1 part, or `never`
 type L2Parts<
     Q extends string,
-    P extends string,
-> = Q extends `${P}:${infer U}:${string}` ? U
-    : Q extends `${P}:${infer U}` ? U
+    L1 extends string,
+> = Q extends `${L1}:${infer L2}:${string}` ? L2
+    : Q extends `${L1}:${infer L2}` ? L2
     : never;
 
 // build up all combinations of all L1 fields
@@ -521,8 +523,9 @@ type L1Shortcuts = KeyOf<typeof L1_SHORTCUTS>;
 type L2Shortcuts = KeyOf<typeof L2_SHORTCUTS>;
 
 type ExpandShortcuts<Q extends string> = Q extends
-    `${infer R}:${infer S}:${infer T}` ? `${ExpandL1<R>}:${ExpandL2<S>}:${T}`
-    : Q extends `${infer R}:${infer S}` ? `${ExpandL1<R>}:${ExpandL2<S>}`
+    `${infer L1}:${infer L2}:${infer L3}`
+    ? `${ExpandL1<L1>}:${ExpandL2<L2>}:${L3}`
+    : Q extends `${infer L1}:${infer L2}` ? `${ExpandL1<L1>}:${ExpandL2<L2>}`
     : ExpandL1<Q>;
 type ExpandL1<S extends string> = S extends L1Shortcuts
     ? typeof L1_SHORTCUTS[S][number]
--- a/src/filter.ts
+++ b/src/filter.ts
@@ -387,30 +387,7 @@ type NotUndefined = string | number | boolean | SomeObject;
  * Given a FilterQuery, returns an object that, when intersected with an Update,
  * marks those properties as required that are guaranteed to exist.
  */
-type RunQuery<Q extends string> = L1Combinations<Q, L1Parts<Q>>;
-
-// build up all combinations of all L1 fields
-type L1Combinations<Q extends string, L1 extends string> = Combine<
-    L1Fields<Q, L1>,
-    L1
->;
-// maps each L1 part of the filter query to an object
-type L1Fields<Q extends string, L1 extends string> = L1 extends unknown
-    ? Record<L1, L2Combinations<L2Parts<Q, L1>>>
-    : never;
-
-// build up all combinations of all L2 fields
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
One approval in https://t.me/grammyjs/34660, would like to see at least one more

KnorpelSenf:
Additionally needs to be applied to `ctx.update`.

src/filter.ts
```diff
@@ -387,41 +387,47 @@ type NotUndefined = string | number | boolean | SomeObject;
  * Given a FilterQuery, returns an object that, when intersected with an Update,
  * marks those properties as required that are guaranteed to exist.
  */
-type RunQuery<Q extends string> = L1Combinations<Q, L1Parts<Q>>;
+type RunQuery<Q extends string> = L1Discriminator<Q, L1Parts<Q>>;
```
KnorpelSenf, line 390: Combinators were renamed to discriminators, because they are used to narrow down properties. The fact that they can be combined is less speaking.

src/filter.ts
```diff
@@ -387,41 +387,47 @@ type NotUndefined = string | number | boolean | SomeObject;
...
+    Q extends string,
+    P extends string,
+> = Q extends `${P}:${infer U}:${string}` ? U
+    : Q extends `${P}:${infer U}` ? U
+    : never;
 
 // build up all combinations of all L1 fields
-type L1Combinations<Q extends string, L1 extends string> = Combine<
-    L1Fields<Q, L1>,
+type L1Discriminator<Q extends string, L1 extends string> = Combine<
```
KnorpelSenf, line 403: Combinators were renamed to discriminators.

src/filter.ts
```diff
@@ -387,41 +387,47 @@ type NotUndefined = string | number | boolean | SomeObject;
...
+    L1Fragment<Q, L1>,
```
KnorpelSenf, line 404: Fields were renamed to fragments, because they make up the fragments of the type to be created. The fact that they are objects which have fields is less speaking.

src/filter.ts
```diff
@@ -387,41 +387,47 @@ type NotUndefined = string | number | boolean | SomeObject;
...
+type L1Fragment<Q extends string, L1 extends string> = L1 extends unknown
+    ? Record<L1, L2Discriminator<L1, L2Parts<Q, L1>>>
     : never;
 
 // build up all combinations of all L2 fields
-type L2Combinations<L2 extends string> = [L2] extends [never] ? NotUndefined // short-circuit L1 queries (L2 is never)
-    : Combine<L2Fields<L2>, L2>;
+type L2Discriminator<L1 extends string, L2 extends string> = [L2] extends
+    [never] ? L2ShallowFragment<L1> // short-circuit L1 queries (L2 is never), only add twins
+    : Combine<L2Fragment<L1, L2>, L2>;
```
KnorpelSenf, line 415: If no L2 filter query is given, we used to simply yield `NotUndefined`. However, now that we have twin properties that cross levels (go from L1 to L2), we additionally have to ensure that sub-types on L2 are present, even though only an L1 filter is given. Hence, we must build up the correct fragment for this.

The fragment can be shallow, it does not actually compute any L2 filter. All it does is to apply the correct cross-level twins.

src/filter.ts
```diff
@@ -387,41 +387,47 @@ type NotUndefined = string | number | boolean | SomeObject;
...
 // maps each L2 part of the filter query to an object and handles siblings
-type L2Fields<L2 extends string> = L2 extends unknown
-    ? Record<L2 | Twins<L2>, NotUndefined>
+type L2Fragment<L1 extends string, L2 extends string> = L2 extends unknown
+    ? Record<L2 | AddTwins<L1, L2>, NotUndefined>
```
KnorpelSenf, line 418: `Twins` was renamed to `AddTwins`, don't ask me why I did that.

src/filter.ts
```diff
@@ -387,41 +387,47 @@ type NotUndefined = string | number | boolean | SomeObject;
...
     : never;
+// does the same as L1Fragment but without combining L2 properties
+type L2ShallowFragment<L1 extends string> = Record<
+    AddTwins<L1, never>,
+    NotUndefined
+>;
```
KnorpelSenf, line 424: Effectively only applied cross-level twins, could be implemented directly via `TwinsFromL1` but this is more readable and equivalent.

src/filter.ts
```diff
@@ -387,41 +387,47 @@ type NotUndefined = string | number | boolean | SomeObject;
...
-// gets all L1 query snippets
-type L1Parts<Q extends string> = Q extends `${infer U}:${string}` ? U : Q;
-// gets all L2 query snippets for the given L1 part, or `never`
-type L2Parts<
-    Q extends string,
-    P extends string,
-> = Q extends `${P}:${infer U}:${string}` ? U
-    : Q extends `${P}:${infer U}` ? U
-    : never;
-
```
KnorpelSenf, line 424: Copied up from here.

src/filter.ts
```diff
@@ -387,41 +387,47 @@ type NotUndefined = string | number | boolean | SomeObject;
...
+
+// gets all L1 query snippets
+type L1Parts<Q extends string> = Q extends `${infer U}:${string}` ? U : Q;
+// gets all L2 query snippets for the given L1 part, or `never`
+type L2Parts<
+    Q extends string,
+    P extends string,
+> = Q extends `${P}:${infer U}:${string}` ? U
+    : Q extends `${P}:${infer U}` ? U
+    : never;
```
KnorpelSenf, line 400: Copied up from below.

src/filter.ts
```diff
@@ -466,7 +472,9 @@ interface Shortcuts<U extends Update> {
         : [U["chat_join_request"]] extends [SomeObject]
             ? U["chat_join_request"]["chat"]
         : undefined;
-    // senderChat: disregarded here because always optional on 'Message'
+    senderChat: [Shortcuts<U>["msg"]] extends [SomeObject]
+        ? Shortcuts<U>["msg"]["sender_chat"]
+        : undefined;
```
KnorpelSenf, line 477: Applies the correct type for the shortcut `ctx.senderChat`.

src/filter.ts
```diff
@@ -524,15 +532,40 @@ type ExpandL2<S extends string> = S extends L2Shortcuts
...
 
 // === Define some helpers for when one property implies the existence of others
-type Twins<V extends string> = V extends KeyOf<Equivalents> ? Equivalents[V]
-    : V;
-type Equivalents = {
+
+// merges twins based on L1 with those based on L1 and L2
+type AddTwins<L1 extends string, L2 extends string> =
+    | TwinsFromL1<L1, L2>
+    | TwinsFromL2<L1, L2>;
```
KnorpelSenf, line 539: `Twins` was renamed to `AddTwins`, and cross-level twins were added to the type definition.

src/filter.ts
```diff
@@ -524,15 +532,40 @@ type ExpandL2<S extends string> = S extends L2Shortcuts
...
+
+// yields twins based on a given L1 property
+type TwinsFromL1<L1 extends string, L2 extends string> = L1 extends
+    KeyOf<L1Equivalents> ? L1Equivalents[L1]
+    : L2;
```
KnorpelSenf, line 544: Cross-level twin implementation. Yields the correct fields on L2 for given keys on L1.

src/filter.ts
```diff
@@ -524,15 +532,40 @@ type ExpandL2<S extends string> = S extends L2Shortcuts
...
+type L1Equivalents = {
+    message: "from";
+    edited_message: "from" | "edit_date";
+    channel_post: "sender_chat";
+    edited_channel_post: "sender_chat" | "edit_date";
+};
```
KnorpelSenf, line 550: Definition of all cross-level twins in the API as of version 5.6.

src/filter.ts
```diff
@@ -524,15 +532,40 @@ type ExpandL2<S extends string> = S extends L2Shortcuts
...
+
+// yields twins based on given L1 and L2 properties
+type TwinsFromL2<L1 extends string, L2 extends string> = L1 extends
+    KeyOf<L2Equivalents>
+    ? L2 extends KeyOf<L2Equivalents[L1]> ? L2Equivalents[L1][L2] : L2
+    : L2;
```
KnorpelSenf, line 556: Old `Twins` implementation, but now takes into account that twins (same-level) are only applied behind the right L1 properties.

For example, before we would also set `document` when given `animation` inside `callback_query` updates, and the only reason this was not a problem is because those clashes coincidentally do not appear.

Now, we make sure that twins are only applied behind the correct update types. This is also a lot more flexible for future API changes.

src/filter.ts
```diff
@@ -524,15 +532,40 @@ type ExpandL2<S extends string> = S extends L2Shortcuts
...
+type L2Equivalents = {
+    message: MessageEquivalents;
+    edited_message: MessageEquivalents;
+    channel_post: MessageEquivalents;
+    edited_channel_post: MessageEquivalents;
+};
+type MessageEquivalents = {
     animation: "document";
-    entities: TextMessages;
+    entities: "text";
```
KnorpelSenf, line 565: Inlined type definition.

src/filter.ts
```diff
@@ -524,15 +532,40 @@ type ExpandL2<S extends string> = S extends L2Shortcuts
...
     caption: CaptionMessages;
     caption_entities: CaptionMessages;
 };
-type TextMessages = "text";
```
KnorpelSenf, line 535: Was inlined.

KnorpelSenf (commented):
This is a documentation review in order to explain the changes better.

EdJoPaTo (approved):
> Just throw it onto a large grammY code base and check that everything still compiles.

```bash
npm i grammyjs/grammy#cross-level-twins && npm test
```

Throwing that into 7 repos of mine created no error. I do not use the feature in question much as most are Telegraf migrations. Still it means nothing seems to be completely out of place.

I had not a look into the code changes tho.

KnorpelSenf:
Thank you for your work! That's perfectly the kind of testing I was hoping for. I'll merge this soon and publish a patch release.

PR #159, opened 2022-01-18
Preliminary attempt to fix complex types with cross-level-twins

MKRhere:
This PQ still needs work.

It works well in the following cases:

```TypeScript
bot.on(["message:text"], ctx => {
	// user: User
	const user = ctx.from;
});

bot.on(["message:text", "channel_post:text"], ctx => {
	// user: User | undefined
	const user = ctx.from;
});
```

And not very well in this case:

```TypeScript
bot.on(["channel_post", ":forward_date"])
	.on(":text")
	.on(["::url", "::hashtag", "::cashtag"], ctx => {
		// Property 'id' does not exist on type 'User | Chat | undefined'. deno-ts(2339)
		const { id } = ctx.from ?? ctx.senderChat;
	});
```

It's still up for visibility now, I'll attempt to resolve this as well, although it is definitely a bit complex.

### DIFF
--- a/src/filter.ts
+++ b/src/filter.ts
@@ -383,13 +383,16 @@ export type FilterQuery = AllValidFilterQueries;
  */
 // deno-lint-ignore ban-types
 type SomeObject = object;
-type NotUndefined = string | number | boolean | SomeObject;
 
 /**
  * Given a FilterQuery, returns an object that, when intersected with an Update,
  * marks those properties as required that are guaranteed to exist.
  */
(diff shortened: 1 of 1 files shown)


KnorpelSenf (commented):
I know this is not done, and I'll review more thoroughly when it is, but here are two comments already. They are actually rather a comment and a question:

KnorpelSenf:
Are you still planning on working on this? :)

MKRhere:
Closing, since I don't think I'll be able to tackle it for now. If I look at it again, I'll open a new PQ.

PR #162, opened 2022-01-22
Refactor polling and handle 429 errors

KnorpelSenf:
Refactors the built-in polling to be a lot more readable by extracting the different parts into methods.

Correctly handles 429 errors thrown during polling. These are never thrown during normal operation, but can happen while the Bot API server is closing. In that case, the returned `retry_after` value should be respected, which is what this PQ is doing now.

This must be reviewed from Deno 1.17.1 or from Node.js because https://github.com/denoland/deno/issues/13469 prevents grammY from working on Deno>=1.17.2 at the moment.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -244,6 +244,31 @@ export class Bot<
     }
 
     /**
+     * Internal. Do not call. Handles an update batch sequentially by supplying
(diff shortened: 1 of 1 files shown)


roj1512 (approved):
LGTM

Issue #109, opened 2021-11-23
Roadmap 1.10: Conversational Interfaces


KnorpelSenf:
This is a list of tasks that should be completed before the 1.10 release. All of them focus on improving how well the core library is suited to be used in [conversations](https://github.com/grammyjs/conversations).

- #100
- #103
- #105
- #127
- #131
- #157
- #176
- #188
- #247
- #201

This list is tentative. Things may be added or removed if we figure that that makes sense.

There is no ETA—it will be released once everything is done. It goes faster if you contribute.

If you want to help out but you don't know how to decide, please leave a comment here. If you want to work on some issue but you don't know how to begin, please leave a comment there. In any case, you can join [the community chat](https://t.me/grammyjs) and ask.

KnorpelSenf:
The discussion on #136 hints at a number of new core primitives that are required to make conversational interfaces work properly. It may make sense to postpone this update, and to only include changes related to grammY conversations in that first update. I think it'll include #100, #105, and #127, plus the changes needed for conversations.

KnorpelSenf:
The approach I mentioned above did not work out. The new approach the was started in https://github.com/grammyjs/conversations does not require any modifications of the core library. Hence, after grammY conversations are released, this update can be completed in one go.

KnorpelSenf:
Now, a number of features look like they will drastically improve conversations, even though they may technically work without an update to the core lib. Also, there would be a lot of changes in this release.

Consequently, for those two reasons, the 1.9 release will be divided into two releases. I edited this roadmap and removed everything that will be postponed. Check out the roadmap for the next release: #202

PR #164, opened 2022-01-29
Add support for commands in messages with media

martiliones:
Add support for commands in messages with photo, files etc

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -391,11 +391,12 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
             const set = cmd.indexOf("@") === -1 ? noAtCommands : atCommands;
             set.add(cmd);
         });
-        return this.on(":entities:bot_command").filter(
+        return this.on([":entities:bot_command", ":caption_entities:bot_command"]).filter(
             (ctx): ctx is CommandContext<C> => {
                 const msg = ctx.message ?? ctx.channelPost;
-                const txt = msg.text;
-                return msg.entities.some((e) => {
+                const txt = msg.text ?? msg.caption;
+                const entities = msg.entities ?? msg.caption_entities;
+                return entities.some((e) => {
                     if (e.type !== "bot_command") return false;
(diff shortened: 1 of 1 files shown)


src/composer.ts
```diff
@@ -391,11 +391,12 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
             const set = cmd.indexOf("@") === -1 ? noAtCommands : atCommands;
             set.add(cmd);
         });
-        return this.on(":entities:bot_command").filter(
+        return this.on([":entities:bot_command", ":caption_entities:bot_command"]).filter(
```
KnorpelSenf, line 394:
```suggestion
        return this.on("::bot_command").filter(
```

KnorpelSenf (changes requested):
Thanks for the contribution!

This is a breaking change:
```ts
bot.command('start', ctx => {
  // The PR breaks this:
  const text = ctx.msg.text
})
```
I am happy to merge this for a 2.0 release. However, there are a lot of things to do before a 2.0 release makes sense, so this PR will have to remain open for months.

That being said, this only takes so much time because I am the only one actively working on the core package. For example, if someone else could take responsibility for #136, I could put more time into this. Naturally, I also appreciate a lot if anyone wants to help out with #109!

If the composer code was to change, you do not need to update this PR. I will ping you when this becomes relevant again, i.e. when work on 2.0 begins.

src/composer.ts
KnorpelSenf: This suggestion is actually not faster. grammY's internal filter query optimisation makes both things equivalent. It is only shorter.

KnorpelSenf:
Backed by the community in https://t.me/grammyjs/36850

roziscoding:
This functionality is available via the [Commands Plugin](https://github.com/grammyjs/commands), if anyone needs to use it now :)

KnorpelSenf:
Why close it? I was still planning on merging this for 2.0.

PR #165, opened 2022-01-31
Update to Bot API 5.7

KnorpelSenf:
Updates grammY to Bot API 5.7.

Blocked by https://github.com/grammyjs/types/pull/5.

Pending tasks:
- [x] Update dependencies

### DIFF
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -1499,7 +1499,7 @@ export class Api<R extends RawApi = RawApi> {
     }
 
     /**
-     * Use this method to send static .WEBP or animated .TGS stickers. On success, the sent Message is returned.
+     * Use this method to send static .WEBP, animated .TGS, or video .WEBM stickers. On success, the sent Message is returned.
      *
      * @param chat_id Unique identifier for the target chat or username of the target channel (in the format @channelusername)
      * @param sticker Sticker to send. Pass a file_id as String to send a file that exists on the Telegram servers (recommended), pass an HTTP URL as a String for Telegram to get a .WEBP file from the Internet, or upload a new one using multipart/form-data.
@@ -1547,7 +1547,7 @@ export class Api<R extends RawApi = RawApi> {
     }
 
     /**
-     * Use this method to create a new sticker set owned by a user. The bot will be able to edit the sticker set thus created. You must use exactly one of the fields png_sticker or tgs_sticker. Returns True on success.
+     * Use this method to create a new sticker set owned by a user. The bot will be able to edit the sticker set thus created. You must use exactly one of the fields png_sticker, tgs_sticker, or webm_sticker. Returns True on success.
      *
      * @param user_id User identifier of created sticker set owner
      * @param name Short name of sticker set, to be used in t.me/addstickers/ URLs (e.g., animals). Can contain only english letters, digits and underscores. Must begin with a letter, can't contain consecutive underscores and must end in “_by_<bot username>”. <bot_username> is case insensitive. 1-64 characters.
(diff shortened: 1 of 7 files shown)


MKRhere (approved):

EdJoPaTo (approved):

src/core/api.ts
```diff
@@ -1630,19 +1630,19 @@ export class Api<R extends RawApi = RawApi> {
...
+     * @param thumb A PNG image with the thumbnail, must be up to 128 kilobytes in size and have width and height exactly 100px, or a TGS animation with the thumbnail up to 32 kilobytes in size; see https://core.telegram.org/stickers#animated-sticker-requirements for animated sticker technical requirements, or a WEBM video with the thumbnail up to 32 kilobytes in size; see https://core.telegram.org/stickers#video-sticker-requirements for video sticker technical requirements. Pass a file_id as a String to send a file that already exists on the Telegram servers, pass an HTTP URL as a String for Telegram to get a file from the Internet, or upload a new one using multipart/form-data. More info on Sending Files ». Animated sticker set thumbnails can't be uploaded via HTTP URL.
      * @param signal Optional `AbortSignal` to cancel the request
      *
      * **Official reference:** https://core.telegram.org/bots/api#setstickersetthumb
      */
     setStickerSetThumb(
         name: string,
         user_id: number,
-        thumb: InputFile | string,
+        thumb?: InputFile | string,
```
EdJoPaTo, line 1645: oh, maybe I have overlooked something.

setThumb which has an optional thumb? Method feels a bit strange this way?

EdJoPaTo: Its the same in the official docs but I don't understand why that is. The text doesn't describe why it should be optional.

If it is able to remove a thumbnail that should be more explicit to not forget the argument:

```suggestion
        thumb: InputFile | string | undefined,
```

KnorpelSenf: Yeah I thought about this. On the other hand, `Api` is about making it convenient. I would find it odd to have to pass `{ thumb: new InputFile() }` at this point.

There are other methods with only a single optional argument which do the same thing. It just does not make much sense.

The only real case where this can look awkward is when removing the thumbnail but wanting to pass a signal to abort the request, in which case it becomes `setStickerSetThumb(name, id, undefined, signal)`. I still find that readable for what it does.

KnorpelSenf: Especially considering that the alternative would be to pass `{}` instead of `undefined`.

Can I release?

EdJoPaTo: Specifying `undefined` over omitting it makes it clear that this is the wanted behaviour and not something that was forgotten to set.

Guess I am fine either way, but it should be consistent. This should be changed then: https://github.com/grammyjs/grammY/blob/d5731cbe53d9a5ad753656bbcb9226bb8283b956/src/core/api.ts#L1056

KnorpelSenf: > Its the same in the official docs but I don't understand why that is. The text doesn't describe why it should be optional.
> 
> If it is able to remove a thumbnail that should be more explicit to not forget the argument:

I find it weird to interdict something at compile-time that works at runtime. It is perfectly okay not to pass the argument.

This is also the case if we were to introduce an options object, which can always be omitted.

KnorpelSenf: > This should be changed then:
> 
> https://github.com/grammyjs/grammY/blob/d5731cbe53d9a5ad753656bbcb9226bb8283b956/src/core/api.ts#L1056

Good point! I'll fix it and then release.

PR #166, opened 2022-02-01
feat: export Bot API types

### LINKED ISSUE #127: feat: Re-export Bot API types
They currently are exported from `@grammyjs/types`, so users have to have two explicit dependencies. This causes some trouble:

- deps can get out of sync
- especially bad on Deno because it always checks the entire project
- not intuitive to install a second package, while the main one already uses the types
- the `InputFileProxy` type is hard to understand

Implementing this has some disadvantages, too:

- adds ~170 type exports to grammY (can be mitigated with a namespace), but still makes API surface much larger
- API reference will now mix grammY's definitions with those of the Bot API

KnorpelSenf:
As described in the discussion around https://t.me/grammyjs/25957

Closes #127.


slavafomin:
It looks like there is some duplication:

Screenshot_2022-02-02_22-57-42

They are:

`export namespace Chat { ... }`

and

```
export type Chat =
  | Chat.PrivateChat
  | Chat.GroupChat
  | Chat.SupergroupChat
  | Chat.ChannelChat;
```

Is it expected?

KnorpelSenf:
Is this related to the current issue? Could you please check if the same happens with `@grammyjs/types`?

slavafomin:
> Is this related to the current issue? Could you please check if the same happens with `@grammyjs/types`?

Yep, the `@grammyjs/types` has the same two overlapping exports. However, it doesn't look like a problem, it seems that TypeScript can distinguish between the actual type being used and the namespace. It's IDE that is getting confused.

Regarding the original issue, I've confirmed that the fix is working good for my use case, however, you should know that it's pretty limited in scope. I'm only importing the `Chat` type right now.

src/types.ts
```diff
@@ -0,0 +1 @@
+export * from "./types.deno.ts";
```
wojpawlik, line 1: This file is unnecessary.

types.d.ts
```diff
@@ -0,0 +1 @@
+export * from './out/types'
```
wojpawlik, line 1: I'd rather this file wasn't necessary...

wojpawlik (approved):

src/types.ts
KnorpelSenf: From what point of view?

types.d.ts
KnorpelSenf: Can you elaborate? Importing types from `grammy/types` on Node does not work for me without this file.

KnorpelSenf:
> > Is this related to the current issue? Could you please check if the same happens with `@grammyjs/types`?
> 
> Yep, the `@grammyjs/types` has the same two overlapping exports. However, it doesn't look like a problem, it seems that TypeScript can distinguish between the actual type being used and the namespace. It's IDE that is getting confused.

Thanks for confirming. It works with VSCode. I recommend opening an issue at the repo of the respective editor.

> Regarding the original issue, I've confirmed that the fix is working good for my use case, however, you should know that it's pretty limited in scope. I'm only importing the `Chat` type right now.

This change is rather straightforward. It is reasonable to assume that it works with all types if it works with one of them.

types.d.ts
wojpawlik: https://github.com/colinhacks/zod/issues/805

KnorpelSenf: What do you mean? A cryptic link is not a good elaboration, are you somehow suggesting this can be fixed in `package.json` or something?

wojpawlik: Indeed:

```jsonc
  // @filename: package.json
  "exports": {
    ".": {
      "require": "./out/mod.js",
      "types": "./out/mod.d.ts"
    },
    "./types": {
      "types": "./out/types.d.ts"
    },
 }
```

src/types.ts
wojpawlik: Deno users can import directly from `src/types.deno.ts`.

types.d.ts
wojpawlik:
```suggestion
export * from './out/types.node.d.ts'
```

src/types.ts
KnorpelSenf: Wouldn't you consider that to be a leaky abstraction over the platform independency?

types.d.ts
KnorpelSenf: > Indeed:
> 
> ```js
>   // @filename: package.json
>   "exports": {
>     ".": {
>       "require": "./out/mod.js",
>       "types": "./out/mod.d.ts"
>     },
>     "./types": {
>       "types": "./out/types.node.d.ts"
>     },
>   }
> ```

This was contained in the first commit of this PQ. Why would I have removed it? 🤔

KnorpelSenf:
It is desirable to be able to import from `./types.mod` next to `./mod.ts` as well as to import from `"grammy/types"` next to `"grammy"`. This can be achieved by adding a small file that abstracts away from the Deno/Node hybrid nature of the project.

We could expose this difference to users by making them import from `types.(deno|node).ts`, but I would much rather avoid that userland code is exposed to artefacts or our build setup.

src/types.ts
KnorpelSenf: See https://github.com/grammyjs/grammY/pull/166#issuecomment-1126811724 below

KnorpelSenf:
I'll test this a bit more, but unless I find something, this will be merged as-is.

EdJoPaTo:
This does not happen with `grammy` or `grammyjs/grammy#composer-chat-type`:

```
> npm i grammyjs/grammy#export-api-types

> tsc --sourceMap && node --enable-source-maps dist/index.js

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in …/node_modules/grammy/package.json imported from …/dist/bot/index.js
    at __node_internal_captureLargerStackTrace (node:internal/errors:466:5)
    at new NodeError (node:internal/errors:377:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:440:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:664:7)
    at packageResolve (node:internal/modules/esm/resolve:893:14)
    at moduleResolve (node:internal/modules/esm/resolve:959:20)
    at defaultResolve (node:internal/modules/esm/resolve:1174:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:605:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:318:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:80:40)
    at link (node:internal/modules/esm/module_job:78:36) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}

Node.js v18.2.0
```

KnorpelSenf:
Hmm then something is probably still wrong with the config. This needs to be fixed. I know why I did not merge this yet … thanks for testing!

Loskir (approved):
lgtm (haven't tried)

roziscoding (approved):

EdJoPaTo:
Is there a reason why `@grammyjs/types` is a dependency before this PR rather than a devDependency? Normally you would only need types at dev time, not required at runtime.
With this PR its bundled so its needed to be a dependency, I just wonder why its already that way.

EdJoPaTo (approved):
I tested `npm i grammyjs/grammy#export-api-types` in both a CommonJS and an ESM style bot and both worked fine.
I also moved imports from `@grammyjs/types` towards `grammy/types` which also worked.

I havnt looked into the code that deeply but from my user point of view this change seems to be working as intended.

KnorpelSenf:
> Is there a reason why `@grammyjs/types` is a dependency before this PR rather than a devDependency?

grammY provides type safety when you work with it. That's why users must download the types when they download grammY. If it was a dev dep, people would not get the types upon `npm i grammy`.

PR #169, opened 2022-02-27
Modify match type evaluation only

KnightNiwrem:


KnorpelSenf:
What was the purpose of this again?

src/composer.ts
```diff
@@ -827,11 +827,17 @@ function triggerFn(trigger: MaybeArray<string | RegExp>) {
...
+            ? Extract<C[K], string | RegExpMatchArray>
+            : C[K];
+    },
     ":text" | ":caption"
 >;
 type CommandContext<C extends Context> = Filter<
-    C & { match: string },
+    {
+        [K in keyof C]: K extends "match" ? Extract<C[K], string> : C[K];
+    },
```
KnorpelSenf, line 840: Does this work?
```suggestion
    Omit<C, 'match'> & { match: Extract<C['match'], string> },
```

KnorpelSenf (changes requested):
I have one idea how this could be simplified, can you confirm that it works?

src/composer.ts
KnightNiwrem: Verified and accepted

KnorpelSenf: Tremendous

src/context.ts
```diff
@@ -74,7 +74,7 @@ export class Context implements RenamedUpdate {
      * Used by some middleware to store information about how a certain string
      * or regular expression was matched.
      */
-    public match?: string | RegExpMatchArray;
+    public match: string | RegExpMatchArray | undefined;
```
KnorpelSenf, line 77: If I'm not mistaken, this can be reverted now.

KnorpelSenf (changes requested):

src/context.ts
KnorpelSenf: Because we no longer rely on a mapped type which required the field to always be present (even though it isn't).

KnorpelSenf:
```suggestion
    public match?: string | RegExpMatchArray;
```

KnightNiwrem: This is correct

KnorpelSenf (approved):
Thank you so much!

PR #173, opened 2022-03-27
fix: fetch type for Deno 1.20.3

darvesh:
Fixes #172

PR #174, opened 2022-03-27
Migrate from `Deno.File` to `Deno.FsFile` and merge #173

KnorpelSenf:
This is breaking, but only for Deno users who do an update of Deno which deprecates `Deno.File` and makes their linting fail.

Fixes CI.

Includes #173.


darvesh:
LGTM

KnorpelSenf:
@all-contributors add @darvesh for review

KnightNiwrem (approved):

PR #187, opened 2022-04-16
Update to Bot API 6.0

KnorpelSenf:
Updates grammY to Bot API 6.0.

https://core.telegram.org/bots/api#april-16-2022


KnorpelSenf:
For some reason, skypack fails to build `@grammyjs/types@v2.7.0` but it succeeds for `@grammyjs/types@v2.6.0`. I could not figure out what broke it, so we're using https://esm.sh instead now.

dcdunkan (approved):
Looks good to me.

HeeroML (approved):

PR #191, opened 2022-04-21
Improve error message for undefined session key

KnorpelSenf:
As discussed in https://t.me/grammyjs/44217

### DIFF
--- a/src/convenience/session.ts
+++ b/src/convenience/session.ts
@@ -324,7 +324,7 @@ function undef<C extends Context>(
     const lazy = opts.lazy ?? false;
     const reason = getSessionKey === defaultGetSessionKey
         ? "this update does not belong to a chat, so the session key is undefined"
-        : "your custom `getSessionKey` function returned undefined for this update";
+        : "the custom `getSessionKey` function returned undefined for this update";
     return `Cannot ${op} ${lazy ? "lazy " : ""}session data because ${reason}!`;
 }
 
--- a/src/convenience/session.ts
+++ b/src/convenience/session.ts
@@ -189,17 +189,15 @@ export function session<S, C extends Context>(
         Object.defineProperty(ctx, "session", {
             get() {
                 if (key === undefined) {
-                    throw new Error(
-                        "Cannot access session data because the session key was undefined!",
-                    );
+                    const msg = undef("access", getSessionKey);
+                    throw new Error(msg);
                 }
                 return value;
             },
             set(v) {
                 if (key === undefined) {
-                    throw new Error(
-                        "Cannot assign session data because the session key was undefined!",
-                    );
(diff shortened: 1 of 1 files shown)


src/convenience/session.ts
```diff
@@ -314,10 +310,24 @@ export function lazySession<S, C extends Context>(
...
+/** Returns a useful error message for when the session key is undefined */
+function undef<C extends Context>(
+    op: "access" | "assign",
+    getSessionKey: SessionOptions<C>["getSessionKey"],
+    opts: { lazy?: boolean } = {},
+) {
+    const lazy = opts.lazy ?? false;
+    const reason = getSessionKey === defaultGetSessionKey
+        ? "this update does not belong to a chat, so the session key is undefined"
+        : "your custom `getSessionKey` function returned undefined for this update";
```
EdJoPaTo, line 327: in some cases I send the `error.message` to the user and print the full error on the console. This way the user can have some idea what went wrong in some cases. Is it my fault or is it something wrong in the bot. I even got a „this is your problem, fix Line n“ as a PM once because of this. Using „your“ in this case can be confusing.

```suggestion
        : "the custom `getSessionKey` function returned undefined for this update";
```

src/convenience/session.ts
```diff
@@ -314,10 +310,24 @@ export function lazySession<S, C extends Context>(
...
+/** Stores session data per chat by default */
 function defaultGetSessionKey(ctx: Context): string | undefined {
     return ctx.chat?.id.toString();
 }
 
+/** Returns a useful error message for when the session key is undefined */
+function undef<C extends Context>(
+    op: "access" | "assign",
+    getSessionKey: SessionOptions<C>["getSessionKey"],
+    opts: { lazy?: boolean } = {},
```
EdJoPaTo, line 322: I like the named Boolean argument 👌

EdJoPaTo (approved):

src/convenience/session.ts
KnorpelSenf: I would never have though about that. Thanks

src/convenience/session.ts
KnorpelSenf: Yeah I hate looking at a function signature and then there's this third argument _`true`_ … and I'm just sitting there, wondering what the hell that would be

KnorpelSenf: You probably feel the same :D

EdJoPaTo: The rust-analyzer provides inline hints so you can read the argument name of the variable.
It looks like `…, lazy: true)` while the actual source code is just `…, true)`. That helps a lot. But yeah, prefer enums (like `op` is done here) or named arguments, especially when they are optional. 👌

EdJoPaTo (approved):

PR #192, opened 2022-04-22
Avoid prompting for DEBUG env var access

KnorpelSenf:
That's how it was before Deno decided to always prompt by default. This PQ restores the old behaviour.


KnorpelSenf:
Approved in https://t.me/grammyjs/44965

PR #197, opened 2022-05-10
Add Deno shield

roj1512:


roj1512:
If its style not matching with the other badges is a problem, we can modify the styles of the shieldsio badges. The style of the Deno shield is not modifiable.

KnorpelSenf:
Yes, please align the design before we merge this.

roj1512:
@KnorpelSenf I’ve changed the colors of the shieldsio badges, they look similar to the Deno shield now, except for a small unmodifiable white gradient on the message part of them, which is less noticable. You decide :)

roj1512:
@KnorpelSenf How are they looking now?

EdJoPaTo:
The Deno shield has no color gradient, the others have.

image

KnorpelSenf (approved):
LGTM. Thanks! Good idea :)

KnorpelSenf:
In fact, `img.shields.io` allows us to use any logo, text, and design. If we want, we could switch back to the old designs, and simply avoid pulling in the Deno shield from `deno.com`. What do you think?

HeeroML:
👍Am Di., 10.Mai.2022 um 12:26, KnorpelSenf ***@***.***> schrieb:  
In fact, img.shields.io allows us to use any logo, text, and design. If we want, we could switch back to the old designs, and simply avoid pulling in the Deno shield from deno.com. What do you think?

roj1512:
@KnorpelSenf If you prefer, we can remove the version part on the npm badge and do all of the work with shieldsio.

KnorpelSenf:
I think it makes a lot of sense to have the same version on both shields. That way, it's very clear how we publish the identical versions for both runtimes.

roj1512:
That way you should use the Deno shield. Because there won't be a way to add the version to a manually made shieldsio badge.

EdJoPaTo:
Alternatively… isnt shields.io open source? :smirk:

KnorpelSenf:
You want to fork and add the feature we need?

roj1512:
@EdJoPaTo Does that mean you will host an instance for us, or do you mean to PR this feature? They don’t accept PRs easily. Reference: https://github.com/badges/shields/pull/7831

KnorpelSenf:
It seems like there is progress on the linked PQ now.

roj1512:
Yes, happily.

KnorpelSenf:
@roj1512 this PQ modified the section controlled by all-contributors. The styling will be overwritten whenever a new contributor gets listed in the README.

Should we maybe switch to the automatically generated badge the is described in https://github.com/all-contributors/all-contributors/issues/406?

PR #198, opened 2022-05-10
Use the new Telegram logo for the Bot API badge

roj1512:
This is the only possible way. If there was a parameter to force the icon to be chosen from simpleicons.org, it would be shorter.


KnorpelSenf (approved):
Oh well.

Issue #100, opened 2021-11-23
feat: Expose narrowed context and middleware types


KnorpelSenf:
It would be neat to have a helper type for every method on `Composer` (where it makes sense), such as `OnContext`, `CommandContext`, etc.

Accordingly, we should have the matching middleware types: `OnMiddleware`, `CommandMiddleware`, etc.

Requested by @borodutch in https://t.me/grammyjs/23687

PR #203, opened 2022-05-14
Implement filtering by chat type

### LINKED ISSUE #105: feat: Filtering for chat type
We should add `composer.chatType` which takes a `ChatType | ChatType[]` and filters for the given chat types.

KnorpelSenf:
TODO:

- [x] narrow down context type
- [x] add JSDoc

Closes #105.


KnorpelSenf:
This is blocked by #100. The narrowed context for chat types should be exposed. Will wait in order to prevent merge conflicts.

KnorpelSenf:
Types should ideally also reflect that channel posts are sent in channels, and messages aren't. There may be a lot more dependencies between these things. This will get a very complex type annotation, but it is probably worth it.

src/composer.ts
```diff
@@ -1,5 +1,6 @@
-import { type Context } from "./context.ts";
+import { type AliasProps, type Context } from "./context.ts";
 import { type Filter, type FilterQuery, matchFilter } from "./filter.ts";
+import { type Chat } from "./platform.deno.ts";
```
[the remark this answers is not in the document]
KnorpelSenf: Yes. The compilation process will take care of it.

kolay-v (approved):

KnorpelSenf:
Thanks for the review and the suggestion about optional types! I'll merge this as soon as we have a second approval :)

@all-contributors add @kolay-v for review

src/composer.ts
```diff
@@ -419,6 +420,42 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
...
+     * bot.chatType("private", ctx => { ... });
+     * // Channels only
+     * bot.chatType("channel", ctx => { ... });
+     * // Groups and supergroups only
+     * bot.chatType(["group", "supergroup"], ctx => { ... });
+     * ```
+     * Note that this combines very well with other filtering logic. This is how
+     * can you respond to `/start` commands from private chats:
+     * ```ts
+     * bot.chatType("private").command("start")
```
EdJoPaTo, line 439: I think there arent many usecases for using the middleware directly? Normally it will be a combination of filters like command, callbackQuery and so on. So the docs here should prefer using other commands rather than showing the usage with the middleware?

EdJoPaTo (commented):
I like this as it both simplifies some code and provides better typings. Worked in the bots I tried it both without changes and with using the new `chatType` method.

src/composer.ts
KnorpelSenf: I am not 100 per cent sure what you mean, are you suggesting to put this example higher up in the JSDoc comment, and skip the part about using `chatType` on its own?

EdJoPaTo: Most people will use it like this: `bot.chatType("private").command("start")`.

I don't see a real use case to use it like this: `bot.chatType("private", ctx => { ... })`. (Maybe except [stuff like this](https://github.com/HAWHHCalendarBot/TelegramBot/blob/3f9753c567694751006ef30ffc2101e3951d4388/source/parts/easter-eggs.ts#L15-L19))

I think the documentation should show prominently ("higher up") what most users will want to use.

KnorpelSenf: Good point, I'll fix it

src/composer.ts
```diff
@@ -419,6 +420,49 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
         );
     }
 
+    /**
+     * Registers some middleware for certain chat types only. For example, you
+     * can use this method to only receive updates from private chats. The four
+     * chat types are `"channel"`, `"supergroup"`, `"group"`, and `"private"`.
+     * This is especially useful when combined with other filtering logic. For
+     * example, this is how can you respond to `/start` commands from private
+     * chats:
```
EdJoPaTo, line 429:
```suggestion
     * example, this is how can you respond to `/start` commands only from
     * private chats:
```

EdJoPaTo (approved):

KnorpelSenf:
Thanks for the reviews!

Issue #194, opened 2022-05-08
feat: Add static methods to composer


KnorpelSenf:
Can work inside a namespace `Composer.static` which allows a generic implementation via proxies.

KnorpelSenf:
This isn't what was requested. @Loskir please open a new issue describing what you actually meant, if this is still relevant.

PR #208, opened 2022-05-15
Implement Composer.static

KnorpelSenf:
@Loskir @wojpawlik @KnightNiwrem it seems like it is not possible to do what is described in #194. The type signatures lose information when mapped. If you have any ideas how to implement this without duplicating the whole composer class, neither on the type level nor at runtime, please leave a comment.

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -815,7 +815,28 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
         });
         return composer;
     }
+
+    static static = new Proxy({} as ComposerStatic, {
+        get<C extends Context>(_: unknown, op: keyof ComposerStatic) {
+            return (...args: Parameters<Composer<C>[typeof op]>) => {
+                const composer = new Composer<C>();
+                // @ts-ignore fails to check overlapping function signatures
+                composer[op](...args);
+                return composer;
+            };
+        },
+    });
 }
+
+Composer.static.on("message", (ctx) => {
+    const t = ctx.message.text;
+});
+
+export type ComposerStatic = {
+    [op in keyof Omit<Composer<Context>, "middleware">]: <C extends Context>(
+        ...args: Parameters<Composer<C>[op]>
+    ) => ReturnType<Composer<C>[op]>;
+};
 
 // === Util functions and types
 function triggerFn(trigger: MaybeArray<string | RegExp>) {


wojpawlik:
How about just `type ComposerStatic = Omit<Composer<Context>, "use" | "middleware">`?

KnorpelSenf:
This won't allow custom context types.

We could create `static<C extends Context(): ComposerStatic` to fix this. But I'm not sure if it's a good idea.

wojpawlik:
`static get static<C extends Context>(): Composer<Context> {...}`

KnorpelSenf:
`An accessor cannot have type parameters.deno-ts(1094)`

Loskir:
What's your idea of implementing this? Will this help avoid duplication?
It sure looks less convenient, but we'll have to find balance between usability and maintainability.

KnorpelSenf:
It's already implemented. Check out the changes.

The problem is that it doesn't compile, and one possible solution is to make `static` a function rather than an object.

Loskir:
Ah I see. I've tried and for me it doesn't work either way

KnorpelSenf:
Maybe the question wasn't too clear. Let me try again:

It is possible to implement this feature with Composer.static() being a function, rather than a property, contrary to what we had initially discussed. Would such a design still solve your problem? Does it make a difference whether or not you have to put brackets?

Loskir:
I'm still not sure if we need these methods at all.
For `lazy` that was clear: it is a useful util and it can be easily made Composer-independent.
But what about methods like `composer.on`? Do they have uses in a static form? How they will be consumed? What about types?

KnorpelSenf:
`on` is a special case of lazy. I do not see why we need to provide a static version for one of them but not the other.

Loskir:
A logical idea about static methods is that we extract the logic from the same non-static method here and then use it internally instead. You can [see](https://github.com/telegraf/telegraf/blob/7aa720fe4cfc3d53fac9b2b729a8b4c912f54d44/src/composer.ts#L103) that done in Telegraf, for example

The problem is that we have way more complex type system here. For example, `on` method also changes the returned Composer type. So in order to use static `Composer.on` internally we would have to use `composer.cast` or something.

My point is that the usability of static methods is greatly reduced in grammY. I would rather expose some useful utils (maybe under `utils` file?) and leave the rest. The decisive factor here, I think, is whether the method changes the context. Following this logic: `errorBoundary`, `lazy`, `branch` — yes; `on`, `filter` — no

It's a difficult problem, so I suggest we take it slow and don't introduce such mindblowing changes

KnorpelSenf:
I think then this suggestion isn't suited, closing.

Issue #104, opened 2021-11-23
feat: Keyboard button builders and `Keyboard.from`


KnorpelSenf:
Currently, we can only create complete keyboards using the keyboard plugin. It may be useful to add helpers that create individual buttons.

We should then add a static method `Keyboard.from` which takes a 2D-array of button objects, and creates a keyboard from them. Optionally, it handles reshapes on the fly. As a result, you can also pass a 1D-array and a number of columns/rows and the created keyboard will have the specified dimensions.

All of the above should be done for inline keyboards too.

Requested by @Loskir in the discussion around https://t.me/grammyjs/11332

EdJoPaTo:
Putting n buttons into m columns is something helpful that should be part of the main library in my opinion.

See [grammy-inline-menu align.ts](https://github.com/EdJoPaTo/grammy-inline-menu/blob/main/source/buttons/align.ts) and its [test](https://github.com/EdJoPaTo/grammy-inline-menu/blob/main/source/buttons/align.test.ts) for inspiration.

Creating buttons via functions seems overkill to me as its literally only an object with two keys which are autosuggested from typings (even for JavaScript users). This would only create pointless boilerplate code which abstracts something that isnt worth abstracting.

Loskir:
> Creating buttons via functions seems overkill to me as its literally only an object with two keys which are autosuggested from typings

I disagree. Having to write `text` and `callback_data` every time is ridiculous

KnorpelSenf:
I am not sure if I'd ever use the button builders myself, but I do see that especially in functional patters such as `map` and `reduce` it can be very helpful to have util functions for this. Also, given that there seems to be demand for it, I really don't mind adding it. There's virtually no reason against that.

EdJoPaTo:
I don't see the need in functional patterns which I use all the time. I also don't see the argument of writing callback_data which is auto suggested and auto completed from IDEs.
```ts
const buttons = ['a', 'b'].map(o => ({name: o, callback_data: `something-${o}`}));
```

KnorpelSenf:
I think the idea is to be able to do
```ts
const buttons = Keyboard.from(['a', 'b'].map(o => Keyboord.button(o, `something-${o}`)));
```

Loskir:
Update: this is already partially implemented, we have `new Keyboard([[buttons]])`

Helper functions for buttons are not here yet, but we have a package (https://github.com/loskir/grammy-markup) that provides them.

```ts
new InlineKeyboard([
[...]

PR #210, opened 2022-05-16
feat: implement InputMediaBuilder

### LINKED ISSUE #101: feat: Media group builder
Just like we have a builder for `InlineKeyboard` objects, we want one that simplifies creating media group objects which can be passed to the `media` property of https://core.telegram.org/bots/api#sendmediagroup

WingLim:
**Create `builders` subdirectory which other builders can put themselves in it.**

InputMediaBuilder methods:

Basic methods:

- add
- build

Advanced methods:

All these methods use `Omit` to remove `type` property for each different `InputMedia` type, so that user can use it more convenient.

- photo
- video
- animation
- audio
- document

Static method:

`from` method to create a InputMediaBuilder with `InputMedia`.

- from

Furthermore, it have `constructor` to initialize with some `InputMedia`.

It means if you like OOP, you can use `new  InputMediaBuilder(/** data*/)`, or you can use `InputMediaBuilder.from(/** data*/)` if you like FP more.

TODO:
- [ ] Documentation for InputMediaBuilder

Close #101


WingLim:
@KnorpelSenf Plz take a look the code, if it's good for you, then I will write the documentation for it.

roj1512 (commented):
I recommend changing the file name to something simpler (one word), or to `input_media_builder.ts`. So it is consistent with either the other grammY files or official Deno libraries.

WingLim:
The static method `from` comes from #104, and according to https://github.com/grammyjs/grammY/pull/211#issuecomment-1128112737, we can use constructor to do this job.

I don't know which one is better?

WingLim:
> The static method `from` comes from #104, and according to [#211 (comment)](https://github.com/grammyjs/grammY/pull/211#issuecomment-1128112737), we can use constructor to do this job.
> 
> I don't know which one is better?

I have added them both, in case someone like the other style more.

src/convenience/builders/input_media.ts
```diff
@@ -0,0 +1,90 @@
...
+     * The nested array that holds the media.
+     */
+    public readonly media: InputMedia[] = [];
+
+    /**
+     * Initialize a new `InputMediaBuilder` with optional `InputMedia` array.
+     *
+     * @param media The media to initialize the builder with
+     */
+    constructor(media?: InputMedia[]) {
```
KnorpelSenf, line 21: This can be merged with the field declaration.

src/convenience/builders/input_media.ts
```diff
@@ -0,0 +1,90 @@
+import {
+    InputMedia,
+    InputMediaAnimation,
+    InputMediaAudio,
+    InputMediaDocument,
+    InputMediaPhoto,
```
KnorpelSenf, line 6:
```suggestion
    type InputMedia,
    type InputMediaAnimation,
    type InputMediaAudio,
    type InputMediaDocument,
    type InputMediaPhoto,
```

KnorpelSenf (changes requested):
What do you think about also providing something like `InputMediaBuilder.Photo` and the like? That way, we could create stuff using `InputMediaBuilder.Photo.from(...)` which can be passed to `InputMediaBuilder.from` in turn. In addition, it allows for `new InputMediaBuilder.Photo()` with chained method calls. The main builder could then also support the sub-builders, so that `build` does not have to be called on all of them.

I am also wondering if there is a better way to create these things than specifying a config object for every entry. Perhaps this is already possible with the idea described above?

WingLim:
I have a idea about this.

Add a `fromType` static method which accept 2 params:
1. The media array
2. The type of `InputMedia`

Then use a `GenericInputMediaBuilder` to constraint the type of the specific `InputMedia`(e.g. `InputMediaPhoto`) builder.

Let me show you the code to explain it more.

KnorpelSenf:
Please also add some tests.

KnorpelSenf (changes requested):
I now think it is overkill to have a builder for an array. I remember thinking that I would like to abstract that away, but now I believe this would be overkill and take it too far.

I used the rest your changes to create #437. This will subsequently be closed. Thank you anywy for this contribution (and sorry for the long delay)!

Issue #200, opened 2022-05-13
Slower response time compared to telegraf


AradAlvand:
Hi there! Firstly, I'd like to thank you for your remarkable work! I'm pretty new to the world of Telegram bots and before stumbling into grammY, I had been playing with telegraf for a few days and the lack of documentation and so on was really starting to drive me insane, but then I was absolutely delighted to come across grammY! I've found it to be a much nicer and overall far superior framework to both of the current big names, namely [telegraf](https://github.com/telegraf/telegraf) and [node-telegram-bot-api](https://github.com/yagop/node-telegram-bot-api). I think it's criminally underrated at the moment, I sincerely hope it gets more and more popular over time and hopefully at some point overtakes the big two :) Keep up the great work!

Okay, now regarding the issue I've been facing: The only thing I've been slightly disappointed about in grammY has been the noticeably slower response time I consistently experience compared to telegraf (and also node-telegram-bot-api, for that matter).
I'm not sure why this happens though, it's strange. I've created a dead-simple repro (you can find it in [this repo](https://github.com/AradAlvand/grammy-vs-telegraf)), you can just run either `npm run grammy` or `npm run telegraf` and go to http://t.me/telegraf_vs_grammy_bot to take a look at the result.

This is grammY:

https://user-images.githubusercontent.com/26527405/168333909-09eb2f9d-19d7-44a0-9ec5-6e5e832ac423.mp4

This doesn't look terribly slow if you look at it alone, but it's visibly (much) slower than telegraf — see below — and based on my experience this slowness gets more and more noticeable as the bot grows in complexity.

Now, here's telegraf:

https://user-images.githubusercontent.com/26527405/168334591-4da72297-f13f-4bbe-b59c-47fac59b7035.mp4

I think the difference is clear :)

Note that this is all under the same network connection and everything (For what it's worth, I also have a VPN enabled as I'm living in Iran and the government has unfortunately banned Telegram here.) 

Not sure why this is happening, as it seems to me that there shouldn't be a fundamental difference between how grammY and telegraf (or ntba) retrieve updates (this is the default polling technique btw, I've also tried grammY runner — included in [the repo](https://github.com/AradAlvand/grammy-vs-telegraf) — but it doesn't improve the performance in this particular case at all).

Thank you in advance.

KnorpelSenf:
Hi and welcome! Thanks for the kind words, glad you like it :)

[...]

PR #212, opened 2022-05-19
Extend default keepalive on Node to 10 s

KnorpelSenf:
There are two differences between how grammY and Telegraf perform requests. The keepalive time is 1 second on grammY and 10 seconds on Telegraf. The fetched URL is a string on grammY and a `URL` object on Telegraf. `node-fetch` normalises the URL internally.

This sets the default keepalive time for grammY on Node to 10 seconds.

Towards #200.


KnorpelSenf:
This did not fix anything.

PR #213, opened 2022-05-19
chore: update bundling script to Deno 1.22

KnorpelSenf:
Deno 1.22 removes `Deno.emit` and breaks the CI setup. This PQ fixes things as much as possible, but we still need to wait for the replacing tooling to mature before this actually fixes the problem.

Hopefully, we can simply update the dependencies after https://github.com/denoland/deno_emit/issues/17 is fixed and then this will work. It already does for grammY>=1.8.0.


KnorpelSenf:
We should completely remove all bundling from this repo, and instead change the URLs to redirect to https://bundle.deno.dev/

PR #216, opened 2022-05-20
feat: add multi sessions and session enhancing

### LINKED ISSUE #176: feat: Multiple sessions
It would be nice to able to connect to multiple sessions at the same time. The problem with static typing that prevented this from being implemented can be solved by requiring an exhaustive list of properties upon session plugin instantiation. This will also lead to code with higher concurrency.

### LINKED ISSUE #188: feat: Session migrations
Currently in sessions, adding or removing a session data property to the bot (initial) will not modify the existing session data of the chats. It would be cool if sessions could update the stored data to a new version by just modifying the initial data structure.

### LINKED ISSUE #247: Add timeouts to sessions
It would be cool to be able to specify that a session expires, i.e. pass a time to live and then the session data will be undefined as soon as this time has passed.

KnorpelSenf:
Closes #176.
Closes #188.
Closes #247.


Satont (approved):
In my opinion the code is satisfactory, I mean it's good, even though it's duplicated. In any case, we can release it and change it in the future if someone thinks of something specific that could be improved here. 

Otherwise, it looks good to me.

KnorpelSenf:
Alright, thanks for the input! I'll implement the missing multi lazy sessions and then merge it after some testing.

Satont:
> Alright, thanks for the input! I'll implement the missing multi lazy sessions and then merge it after some testing.

TIme to write tests for the library. :)

KnorpelSenf:
Fair point, I'm not as disciplined as you in that respect :D

I mostly limit myself to writing tests for the things that aren't obviously correct (middleware) or that I don't get right on first try—which is a bad practice, I'm afraid.

src/convenience/session.ts
```diff
@@ -385,20 +718,19 @@ export class MemorySessionStorage<S> implements StorageAdapter<S> {
...
+        this.storage.set(key, addExpiryDate(value, this.timeToLive));
     }
 
     delete(key: string) {
         this.storage.delete(key);
     }
 }
+
+function addExpiryDate<T>(value: T, ttl: number) {
+    if (ttl !== undefined && ttl < Infinity) {
```
Loskir, line 730: how is it undefined if it's required?

Loskir:
Is there a link to docs?

src/convenience/session.ts
KnorpelSenf: This is a mistake, I fixed it.

KnorpelSenf:
> Is there a link to docs?

No, they are not written yet.

Loskir:
would love to see then! it's hard to review the vision with just the code

KnorpelSenf:
The referenced issues describe how this is supposed to work. The tests are code examples. Is that enough to review? I'll write docs for the website later today.

Loskir:
Great! I mean, it would be super cool if we had a brief overview of changes and ideas behind them before we merge.

KnorpelSenf:
Yeah … I agree, I'll try to provide better descriptions for the PQs in the future

KnorpelSenf:
@Loskir could you do a final review, now knowing the docs?

Loskir:
Am I right that we can have a default storage adapter that the session falls back to? 

> If you wish to use the default configuration, you can specify an empty object

Would be nice to add the default parameters to the code example because this fact wasn't very obvious for me.

Also can we have it so that we don't have to specify an empty object for that to work? I'm thinking about an optional property so that no property = default adapter

KnorpelSenf:
> Am I right that we can have a default storage adapter that the session falls back to?

It will fall back to the same default that the regular session falls back to: storing things in RAM. Basically, the defaults for every fragment will work the exact same way the defaults are working now for the entire session.

> > If you wish to use the default configuration, you can specify an empty object
> 
> Would be nice to add the default parameters to the code example because this fact wasn't very obvious for me.

Alright, I'll fix this in the docs.

> Also can we have it so that we don't have to specify an empty object for that to work? I'm thinking about an optional property so that no property = default adapter

Right now, this is not possible. The implementation defines properties on `ctx.session` using `Object.defineProperty`. This happens upfront, so we need to know which properties are going to be needed. Hence, people have to specify this upfront.

However, this is of course an implementation detail. We could somehow rewrite the plugin once more and try to chose different abstractions so that the properties can be created lazily on demand using a proxy. This can be difficult because in strict mode, adding a property to `ctx.session` is async because it needs to do IO, but adding properties upon property access in strict mode must be sync, so the point is that there are some internal challenges. I don't think I'll look into this before the release, but we can open an issue if we want to be able to omit the config options.

Loskir:
What if we used proxies? That’s basically a regular session with certain properties overridden with different adapters, isn’t it?

KnorpelSenf:
Yes, that's what we would have to do. Due to the fact that we're both sync and async, we'd have to rewrite a lot of things here, though.

> so that the properties can be created lazily on demand using a proxy.

src/convenience/session.ts
```diff
@@ -173,41 +216,55 @@ export interface SessionOptions<S> {
...
+                ctx.session,
+                prop,
+                initial,
+            );
+            const key = await getSessionKey(ctx);
+            await s.init(key, { custom, lazy: false });
+            return s;
+        }));
+        await next(); // no catch: do not write back if middleware throws
+        if (ctx.session == null) propSessions.forEach((s) => s.delete());
```
Loskir, line 266: shouldn't we add await here?

Loskir: I see, it's synchronous

KnorpelSenf: It's like doing `ctx.session.foo = undefined` manually if people do `ctx.session`.

Loskir:
So `LazySessionFlavor` isn't compatible with `LazyMultiSessionFlavor`?

KnorpelSenf:
Correct. This makes sense—once, `ctx.session` is the promise, and accessing it will trigger a database read, and once, `ctx.session.foo` is a promise. Accessing `ctx.session` should do nothing. That way, we can have partial read operations.

I think this is missing in the docs. I will also add it.

Loskir:
So every plugin will have to add support for it?

KnorpelSenf:
Hmmmmmm that's a very good point

Loskir:
I have a hot take: what if we remove lazy multi sessions altogether for this release?

KnorpelSenf:
To what advantage

Loskir:
Not to ship a niche yet incompatible feature that we'll have to deal with later

KnorpelSenf:
There are currently no known plugins that use sessions.

Lazy multi sessions are a main reason to have multi sessions in the first place. The use case is the following: take the conversations plugin. With lazy multi sessions, you can only read the conversation data if you are inside a conversation, and only read your other data if you are outside of it. Adding support is a trivial change.

The i18n plugin which will be released soon can also be updated before the release.

Loskir:
> read the conversation data if you are inside a conversation

But you need this data in order to know you're inside a conversation. You can store it separately, of course.

> Adding support is a trivial change

`(await (await ctx.session).property)`?

This stuff is getting more and more obscure for a slim benefit of parallel writes (you don't even have parallel reads in lazy multi sessions). If your plugin is so complex it needs this sort of optimization, I'd rather move it to the plugin itself

KnorpelSenf:
I originally did not want to add multi sessions without lazy multi sessions. This would make lazy sessions a second-class citizen that only works with some features of sessions, but not all. However, it's true that the advantages of it are indeed debatable. While parallel reads could technically work, I doubt that it's worth the effort. Perhaps removing lazy multi sessions is a good idea, I will have to think about this. I'll have food and get back to you.

KnorpelSenf:
I have concluded that you're right, I'll remove them. It there is demand in the future (or we simply have a genius idea), we can reconsider how this should look.

KnorpelSenf:
@Loskir anything else?

Loskir:
What if we split the enhance plugin into two? "Enhance" doesn't hold much meaning by itself. And this way we could have isolated transformers for adapters: TTL and Migrations.

KnorpelSenf:
That was the other option how it could have been designed. However, this would make it quite nasty to combine just two transformers:
```ts
bot.use(
  session({
    storage: addTTL({
      storage: addMigrations({
        storage: new MongoDBAdapter({
          collection: sessionCollection
        }),
        migrations: {
          1: migrationFunction
        }
      }),
      millisecondsToLive: 300_000 // 5 min
    })
  })
)
```
Effectively, every added transformer increases nesting. If we want to support three or more of these enhancements, this will get painful. Having everything on a flat list solves the problem.

It is also more efficient, because we can wrap the data once, and then store all information alongside the session. Nested adapters would automatically lead to nested wrappers, which duplicates the overhead.

It is also easier to use because the order is determined by the plugin, rather than the user. Only one of the two orders makes sense, because otherwise the timeouts will get lots after migrating, so users actually have to understand the inner workings and make sure to use things in the right order. Having everything coordinated by the enhancement function solves the problem.

At first, it may look like the current design is less flexible, but this turns out not to be true: also the current design allows users to define their own enhancements if they want to. All this does is simplify the usage of two common enhancements.

Loskir:
I don't like how currently the enhancer wraps session data into an object with __v field. I'd rather store timestamp and version alongside the main data

Loskir:
> the timeouts will get lost after migrating

yeah that's true. it's hard to use the migrator when you have some fields in your session that you don't have control over

Loskir:
> The main data can be primitive values

True, didn't think about that

> isn't any less flexible
> isn't a nice API

The problem I see here is that we don't have a good API for enhanced sessions (unlike, for example, API transformers). So much so that we had to merge our first party enhancers into one so we don't have to deal with nesting. It's not going to get better with third-party enhancers if someone would want to make one.

Loskir:
But since it's not a new API but rather a quirky wrapper around a regular adapter, I think I'm fine with it

KnorpelSenf:
> I have concluded that you're right, I'll remove them. It there is demand in the future (or we simply have a genius idea), we can reconsider how this should look.

We are starting to see some demand for lazy multi sessions now: https://t.me/grammyjs/83263

PR #218, opened 2022-05-24
feat: add `ctx.has`

### LINKED ISSUE #201: feat: Add `ctx.has`
This should be a namespace to query the context object for certain properties, such as whether it contains a command or not.

Once added, this can be reused in the composer, hence improving SoC internally in the library.

KnorpelSenf:
Closes #201.


Mazrooee:
Best project

roziscoding (approved):

src/context.ts
```diff
@@ -19,6 +20,11 @@ import {
     type UserFromGetMe,
 } from "./platform.deno.ts";
 
+// === Util types
+export type MaybeArray<T> = T | T[];
+// deno-lint-ignore ban-types
+export type StringWithSuggestions<S extends string> = (string & {}) | S; // permits `string` but gives hints
```
Loskir, line 26: are you sure we want to export this?

KnorpelSenf: This isn|t exported from grammY, it is only exported because we want to provide the same suggestions in composer.ts.

KnorpelSenf:
@Loskir I did the renaming, please TAL

src/context.ts
```diff
@@ -32,6 +38,164 @@ export type AliasProps<U> = {
...
+        const hasCallbackQuery = checker.filterQuery(
+            "callback_query:game_short_name",
+        );
+        const trg = triggerFn(trigger);
+        return <C extends Context>(ctx: C): ctx is GameQueryContext<C> =>
+            hasCallbackQuery(ctx) &&
+            match(ctx, ctx.callbackQuery.game_short_name, trg);
+    },
+    inlineQuery(trigger) {
+        const hasCallbackQuery = checker.filterQuery("inline_query");
```
Loskir, line 191: Maybe hasInlineQuery?

KnorpelSenf: No. That would lead to `Context.has.hasInlineQuery`, which is redundant.

Loskir: No, i mean the const

KnorpelSenf: Oh. Good point, I'll fix it, thanks..

PR #219, opened 2022-05-25
feat(ci): release & commitlint workflows

roj1512:
- Initialized commitlint workflow to enforce the Conventional Commits style for commit messages.
- Initialized release workflow to creat releases with automatically-generated release notes on tag creations.
- Updated the versions for `actions/checkout`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,18 @@
+name: Main
+
+on:
+    push:
+        tags:
+            - "v*.*.*"
+
+jobs:
+    build:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v2
+
+            - name: Release
+              uses: softprops/action-gh-release@v1
+              with:
+                  generate_release_notes: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,10 +3,11 @@ name: Main
 on:
     push:
         tags:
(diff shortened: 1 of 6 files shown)


roj1512:
Yes, it is.

roj1512:
It will run 2 times for a pull request coming from the same repository (this repository). Should we do it on push only for main?

KnorpelSenf:
> automatically and add GitHub releases whenever a new tag is created on `main`

Do you want to do this here, too?

roj1512:
Should I?

KnorpelSenf:
I think that would make sense, yes. For me, that's the main point of the formats. Readability may also get better, but I must admit that this isn't typically an issue.

roj1512:
Thanks for the article you wrote on your opinions. I thought that I might think of doing both of them here, and I came up with the consequence that I should.

.github/workflows/release.yml
```diff
@@ -0,0 +1,18 @@
+name: Main
+
+on:
+    push:
+        tags:
+            - "v*.*.*"
```
KnorpelSenf, line 6: I think this will trigger if anyone in the organisation pushes a tag to any branch. Can we be stricter and somehow require that the tags are on the `main` branch? If not, I'll see whether I can disallow tag pushes from org members.

KnorpelSenf (changes requested):

KnorpelSenf (approved):
Thanks! Let's hope this works 😄

PR #221, opened 2022-05-25
fix(release.yml): correct the excluded authors

roj1512:


.github/release.yml
```diff
@@ -1,7 +1,6 @@
 changelog:
     exclude:
         authors:
-            - dependabot
```
EdJoPaTo, line 4: Why remove the dependabot?

roj1512: Are we using it? No.

PR #224, opened 2022-05-28
Add tests for most of the library

KnorpelSenf:
Adds a lot of drafts for the tests we always should've had.

### DIFF
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -65,7 +65,7 @@ jobs:
               run: deno task test
 
             - name: Code Coverage Report
-              uses: romeovs/lcov-reporter-action@v0.2.11
+              uses: andybelltree/lcov-reporter-action@v1.0.0
               with:
                   lcov-file: ./lcov.info
-                  github-token: ${{ secrets.GITHUB_TOKEN }}
+                  filter-changed-files: true
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -81,6 +81,8 @@ jobs:
               run: deno task test
 
             - name: Code Coverage Report
-              uses: mszlgr/lcov-reporter-action@v0.2.23
+              uses: andybelltree/lcov-reporter-action@v1.0.0
               with:
+                  title: Code Coverage Report
                   lcov-file: ./lcov.info
+                  delete-old-comments: true
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -68,3 +68,4 @@ jobs:
               uses: romeovs/lcov-reporter-action@v0.2.11
               with:
                   lcov-file: ./lcov.info
+                  github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -69,6 +69,8 @@ jobs:
         steps:
             - name: Setup repo
               uses: actions/checkout@v3
+              with:
+                  fetch-depth: 0
 
             - uses: denoland/setup-deno@main
               with:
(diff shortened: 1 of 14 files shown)


KnorpelSenf:
Note that coverage reports are easy to obtain using `deno test --coverage ./test/`, however, they analyse the context and the API class and we would ideally like to omit these files as testing them is pointless.

Satont:
Can you add gh action on any push for runing tests?

Satont:
Also `deno task` command much be appreciated.

KnorpelSenf:
> Can you add gh action on any push for runing tests?

Tests have always been running in CI on all three platforms since the early days of grammY.

> Also `deno task` command be appreciated.

Sure, I'll add it.

KnorpelSenf:
`deno task` scripts are added and `npm run` scripts are updated. Can you review?

KnorpelSenf:
@Satont are you aware of any code coverage reporters for LCOV files that work in GitHub Actions and that actually work? Neither romeovs/lcov-reporter-action nor mszlgr/lcov-reporter-action nor andybelltree/lcov-reporter-action are able to report test coverage for all tested files and without crashing, for a number of reasons.

Satont:
> @Satont are you aware of any code coverage reporters for LCOV files that work in GitHub Actions and that actually work? Neither `romeovs/lcov-reporter-action` nor `mszlgr/lcov-reporter-action` nor `andybelltree/lcov-reporter-action` are able to report test coverage for all tested files and without crashing, for a number of reasons.

Codecov. They have integration with gh actions. Also check https://v1rtl.site/blog/coverage-in-deno, might be helpful.

KnorpelSenf:
Thank you! Will check it out!

KnorpelSenf:
@Satont thanks, this looks great! If you approve, we can merge and continue working on the test coverage in future PQs.

deno.jsonc
```diff
@@ -1,4 +1,11 @@
 {
+    "tasks": {
+        "setup": "npm install && deno cache --check src/mod.ts",
+        "backport": "npm run backport",
+        "test": "deno test --seed=123456 ./test/",
+        "coverage": "deno task test --coverage=./test/cov_profile && deno coverage --lcov --output=./coverage.lcov ./test/cov_profile",
+        "bundle": "cd bundling && ./bundle-all.sh"
```
roj1512, line 7: Some of these tasks are duplicated in `package.json`. Maybe they should be removed either here or there.

KnorpelSenf: Thanks, fixed it up.

Satont: I guess he means you are duplciating scripts in package.json and deno.jsonc. It should be only in one place. So, you are not fixed the take.

Satont (approved):
Fix duplicates and then LGTM. 👍🏻

.github/workflows/deno.yml
```diff
@@ -56,10 +56,30 @@ jobs:
...
+              with:
+                  deno-version: v1.x
+
+            - name: Create coverage files
+              run: deno task coverage
+
+            - name: Collect coverage
+              uses: codecov/codecov-action@v1.0.10 # upload the report on Codecov
+              with:
+                  file: ./coverage.lcov
```
roj1512, line 85: I don't know if it is better or affects the speed, but you can collect the coverage, too, with one `checkout` and `setup-deno`, both in a single job.

deno.jsonc
KnorpelSenf: Well, the logic is not duplicated. Having npm scripts for everything is mostly a convenience thing for Node users. Should we remove all npm scripts completely?

.github/workflows/deno.yml
KnorpelSenf: I'm not sure what you mean, how can a single job perform both the checkout and the Deno installation?

roj1512: I mean doing the coverage collection in the `test` job, but that seems it is going to be not helpful as it is run 3 times and producing the same result.

PR #228, opened 2022-06-14
chore: add no-prompt to task

KnorpelSenf:


wojpawlik (approved):
`backport` in `package.json` should be updated as well, or just removed.

PR #229, opened 2022-06-21
Update to Bot API 6.1

KnorpelSenf:
Performs the changes announced in https://core.telegram.org/bots/api#june-20-2022

### DIFF
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -121,16 +121,16 @@ export class Api<R extends RawApi = RawApi> {
     }
 
     /**
-     * Use this method to specify a url and receive incoming updates via an outgoing webhook. Whenever there is an update for the bot, we will send an HTTPS POST request to the specified url, containing a JSON-serialized Update. In case of an unsuccessful request, we will give up after a reasonable amount of attempts. Returns True on success.
+     * Use this method to specify a URL and receive incoming updates via an outgoing webhook. Whenever there is an update for the bot, we will send an HTTPS POST request to the specified URL, containing a JSON-serialized Update. In case of an unsuccessful request, we will give up after a reasonable amount of attempts. Returns True on success.
      *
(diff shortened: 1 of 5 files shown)


src/core/api.ts
```diff
@@ -121,16 +121,16 @@ export class Api<R extends RawApi = RawApi> {
...
      *
-     * If you'd like to make sure that the Webhook request comes from Telegram, we recommend using a secret path in the URL, e.g. https://www.example.com/<token>. Since nobody else knows your bot's token, you can be pretty sure it's us.
+     * If you'd like to make sure that the webhook was set by you, you can specify secret data in the parameter secret_token. If specified, the request will contain a header “X-Telegram-Bot-Api-Secret-Token” with the secret token as content.
      *
      * Notes
      * 1. You will not be able to receive updates using getUpdates for as long as an outgoing webhook is set up.
      * 2. To use a self-signed certificate, you need to upload your public key certificate using certificate parameter. Please upload as InputFile, sending a String will not work.
      * 3. Ports currently supported for Webhooks: 443, 80, 88, 8443.
      *
-     * NEW! If you're having any trouble setting up webhooks, please check out this amazing guide to Webhooks.
```
EdJoPaTo, line 133: I think this "NEW" is something that shouldn't be in the source documentation. Looking at a function header and its docs the "new" isn't really helpful?


But I think its somewhat automatically parsed from the official docs where it also disappeared?

EdJoPaTo (approved):

src/core/api.ts
KnorpelSenf: It's not parsed, but yes, it reflects the Bot API docs.

KnorpelSenf:
@all-contributors add Mi3liX9 for code!

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60894

PR #230, opened 2022-06-21
Add support for secret header validation

KnorpelSenf:
Check out the new https://core.telegram.org/bots/api#setwebhook

### DIFF
--- a/src/convenience/frameworks.deno.ts
+++ b/src/convenience/frameworks.deno.ts
@@ -1,10 +1,13 @@
 // deno-lint-ignore-file no-explicit-any
 
+const SECRET_HEADER = "X-Telegram-Bot-Api-Secret-Token";
+
 /** std/http web server */
 const stdHttp = (req: Request) => {
     let resolveResponse: (res: Response) => void;
     return {
         update: req.json(),
+        header: req.headers.get(SECRET_HEADER) || undefined,
         end: () => {
             if (resolveResponse) resolveResponse(new Response());
         },
@@ -12,6 +15,15 @@ const stdHttp = (req: Request) => {
             if (resolveResponse) {
                 const res = new Response(json, {
                     headers: { "Content-Type": "application/json" },
+                });
+                resolveResponse(res);
+            }
+        },
+        unauthorized: () => {
+            if (resolveResponse) {
+                const res = new Response('"unauthorized"', {
+                    status: 401,
+                    statusText: "secret token is wrong",
                 });
                 resolveResponse(res);
             }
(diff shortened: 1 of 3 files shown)


src/convenience/frameworks.deno.ts
```diff
@@ -44,6 +61,13 @@ const serveHttp = (requestEvent: Deno.RequestEvent) => ({
         requestEvent.respondWith(
             new Response(JSON.stringify(json), { status: 200 }),
         ),
+    unauthorized: () =>
+        requestEvent.respondWith(
+            new Response('"unauthorized"', {
+                status: 401,
+                statusText: "secret token is wrong",
```
Loskir, line 68: I'm not sure if statusText is meant for that. We should probably put that text into the first argument

src/convenience/webhook.ts
```diff
@@ -91,7 +136,12 @@ export function webhookCallback<C extends Context = Context>(
             await bot.init();
             initialized = true;
         }
-        const { update, respond, end, handlerReturn } = server(...args);
+        const { update, respond, unauthorized, end, handlerReturn, header } =
+            server(...args);
+        if (header !== token) {
```
Loskir, line 141: shouldn't we rather use `token && header !== token`? to have a way to completely disable this check

src/convenience/frameworks.node.ts
```diff
@@ -10,52 +12,68 @@ const http = (req: IncomingMessage, res: ServerResponse) => ({
...
+    unauthorized: () => res.send(401, "secret token is wrong"),
 });
 
 /** AWS lambda serverless functions */
 const awsLambda = (event: any, _context: any, callback: any) => ({
     update: JSON.parse(event.body),
+    header: event.headers[SECRET_HEADER],
     end: () => callback(null, { statusCode: 200 }),
     respond: (json: string) => callback(null, { statusCode: 200, body: json }),
+    unauthorized: () => callback(null, { statusCode: 401 }),
```
roj1512, line 76: No body?

KnorpelSenf: https://t.me/grammyjs/58046

src/convenience/webhook.ts
KnorpelSenf: If Telegram delivers a token, but it is not supplied here, then there is a mismatch and the check will fail. Currently, the two configurations must match exactly.

What you are suggesting is to ignore if Telegram sends a token, is that better?

KnorpelSenf:
I have tested this with a bot and it works, but we are going to perform a follow-up PQ by @kraftwerk28 which will align the different frameworks. I suggest to merge and release as soon as https://github.com/grammyjs/grammY/pull/230#discussion_r902485165 is sorted out.

PR #235, opened 2022-07-09
Fix fastify webhook updates when secretToken given

oott123:
Fastify will normalize all headers to lower case form, so we always get `undefined` by using `req.headers[SECRET_HEADER]`.

This PR fixes that by using `toLowerCase` to transform the given header key.

Not sure If CloudFlare or AWS Lambda will do that too, so they are leaved unmodified.


KnorpelSenf (approved):
Great fix, thanks!

KnorpelSenf:
@all-contributors add @oott123 for bug and code

KnorpelSenf:
@oott123 your changes are now shipped in the patch version v1.9.1.

PR #240, opened 2022-07-16
fix: http webhook header when no secret token

### LINKED ISSUE #239: Vercel no longer works with `express` middleware
Version 1.9.0 [introduced(https://github.com/grammyjs/grammY/compare/v1.8.3...v1.9.0?w=1#diff-7e0cb595b6c70fce01f4db0b178c13debcafa5fde3df0e4afba3b00322c31c74R25) the usage of `req.header` which is valid for express, but not for Vercel.

The example for usage with Vercel uses the `express` middleware, which no longer works. I've verified that the `fastify` middleware is able to work with Vercel, since it [uses](https://github.com/grammyjs/grammY/compare/v1.8.3...v1.9.0?w=1#diff-7e0cb595b6c70fce01f4db0b178c13debcafa5fde3df0e4afba3b00322c31c74R55) `req.headers[SECRET_HEADER]` and `req.headers` does exist in Vercel's request object.

I'm willing to create a PR to either change the example bot to use `fastify` instead of `express`, but I'm also willing to create a middleware specifically for Vercel. Which one would be better?

roziscoding:
Using `String` here makes the function return `'undefined'` instead of
`undefined`, thus breaking the verification process

Fixes #239

### DIFF
--- a/src/convenience/frameworks.node.ts
+++ b/src/convenience/frameworks.node.ts
@@ -7,12 +7,14 @@ const SECRET_HEADER = "X-Telegram-Bot-Api-Secret-Token";
 const http = (req: IncomingMessage, res: ServerResponse) => ({
     update: new Promise<any>((resolve, reject) => {
         const chunks: Buffer[] = [];
-        req.on("data", (chunk) => chunks.push(chunk)).once("end", () => {
-            const raw = Buffer.concat(chunks).toString("utf-8");
-            resolve(JSON.parse(raw));
-        }).once("error", reject);
+        req.on("data", (chunk) => chunks.push(chunk))
+            .once("end", () => {
+                const raw = Buffer.concat(chunks).toString("utf-8");
+                resolve(JSON.parse(raw));
+            })
+            .once("error", reject);
     }),
-    header: String(req.headers[SECRET_HEADER.toLowerCase()]),
+    header: req.headers[SECRET_HEADER.toLowerCase()],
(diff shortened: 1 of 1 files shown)


KnorpelSenf (approved):
LGTM, thank you!

KnorpelSenf:
@roziscoding however, CI tells me I'm wrong. There could be an array of headers, rather than a single header. Could you fix this by accounting for that case?

You can check if your code compiles by running `npm install && npm run backport`, or `deno task backport`.

roziscoding:
Sure! I didn't know to run backport. Should this be added to the contributing guidelines? (if it's already there and I've missed, I'm sorry lol)

KnorpelSenf:
Good point, I'll make sure that this will make its way into the guide, thanks!

KnorpelSenf (approved):
LGTM, thank you so much!

KnorpelSenf:
We usually require all commits to be signed by contributors. I will overrule now and merge anyway, but I recommend you to look into this in the future :)

KnorpelSenf:
@all-contributors add @roziscoding  for bug and code

roziscoding:
> We usually require all commits to be signed by contributors. I will overrule now and merge anyway, but I recommend you to look into this in the future :)

Oh boy. I usually sign all my commits. I must've forgotten to set it up in my newest setup. Thanks for the heads up. I'll do it right now :)

PR #249, opened 2022-07-28
feat: implement InlineQueryResultBuilder

### LINKED ISSUE #102: feat: InlineQueryResult builder
Just like we have a builder for `InlineKeyboard` objects, we want one that simplifies creating inline query result objects which can be passed to the `results` property of https://core.telegram.org/bots/api#answerinlinequery

ssistoza:
👋 Hey grammY Team, 
Closes #102 

Hopefully no had a chance to work on this yet..but wanted to give it a go myself!

The InlineQueryResultBuilder does not restrict to 50 max for using it in https://core.telegram.org/bots/api#answerinlinequery. I haven't added any docs yet just to take a first pass and see if I'm going in the direction you were thinking about.

The builder add by the type of query result. For types that have cached types, the consumer can either supply a generic and set it true or use the friendlier option suffixed with `Cache`

```ts
const queries = new InlineQueryResultBuilder();
queries.voice({/*Will expect non-cached type*/})
queries.voice<true>({/*Will expect cached type*/})

// A more explicit option rather than seeing true.
queries.voiceCache({/**/})
```
 
A build must be called in order to get all of  the `Array<InlineQueryResult>`

```ts
const inlineQueryResults = queries.build()
```


KnorpelSenf:
Hi there! Thanks for the contribution!

> Hopefully no had a chance to work on this yet..but wanted to give it a go myself!

You're the first!

> The InlineQueryResultBuilder does not restrict to 50 max for using it in https://core.telegram.org/bots/api#answerinlinequery. I haven't added any docs yet just to take a first pass and see if I'm going in the direction you were thinking about.

grammY typically doesn't enforce limits like that in the library. Telegram might change these limits without prior notice, so this isn't useful anyway.

> or use the friendlier option suffixed with `Cache`

Ideally, there should be only one obvious way to do things, rather than two redundant ones with the same result. I think I like it better to give it an explicit name, rather than passing `true` on the type level.

Otherwise, this looks pretty much how I expected the plugin to work. There are two open questions:

1. The current design still requires users to specify large config objects for every result. Is there a better way, maybe a way that allows us to build up these objects via chained functions, as well? Think: `.id("I'd").title("title").cached()`.
2. Often, inline query results are created from arrays. For example, one might want to fetch a list of entries from an API, and then turn these into a list of results. It would be great to have a more functional way of working with this plugin. Essentially, something that works nicely with `map`, `reduce`, etc. @Loskir is the one who is eager about this being added to keyboards and inline keyboards, too, as they have the same problem. We want to fix all of these things in the same update. Do you have any suggestions for this syntax?

ssistoza:
> 1. The current design still requires users to specify large config objects for every result. Is there a better way, maybe a way that allows us to build up these objects via chained functions, as well? Think: .id("I'd").title("title").cached().

Yeah, 😞. It's hard to build out objects with type safety.  I think there are two ways two go about this:

1. We can either supply a default to all required properties when the builder is constructed `new InlineQueryAudioBuilder()`. This is probably the easiest route for ensuring type safety.  
2. I think is is my preferred route figure out the types as build out those chains. I did a bit of experimenting on this in `src/convenience/inline-query/audio.ts`. I'd like to continue to this figure out in order to get proper typing. 

📸-07-29 at 17 00 30@2x


>2. Often, inline query results are created from arrays. For example, one might want to fetch a list of entries from an API, and then turn these into a list of results. It would be great to have a more functional way of working with this plugin. Essentially, something that works nicely with map, reduce, etc. @Loskir is the one who is eager about this being added to keyboards and inline keyboards, too, as they have the same problem. We want to fix all of these things in the same update. Do you have any suggestions for this syntax?

At this moment, i don't have any suggestions. hmm, I probably don't have much understanding here but my thoughts on this would probably be to push list creation to the ones responsible for consuming them (at least in the case for InlineQueryResultBuilder).  Tho i think this becomes less functional:

```ts
const someApiResult = []

const newResults = someApiResult((r) => {
  if (r.type == 'audio') {
    return new InlineQueryResultAudioBuilder(r).build()
  }
  /// ...  other builders as needed.
})
```
Hopefully, this kind of what your thinking about?

KnorpelSenf:
#104 is related to the discussion about a more functional style

Loskir:
I don’t use inline mode that much, but I don’t see a problem in not having any builders at all
For some reason this thing feels completely different from keyboards. Maybe because it’s just a 1d array, maybe because there are so much more possible properties

KnorpelSenf:
Check out the docs of `@grammYdocsbot`, it isn't very pretty IMO: https://github.com/grammyjs/docs-bot/blob/9b98f74886e1c9d191f35d72aac733165bb2f6e2/bot.ts#L51

ssistoza:
>I assume you are planning on creating 19 (?) different builders for all the different types of results, each of them being able to construct the respective inline query result, is that correct?

Yeah i'm thinking a builder for each of their respective query results tho I'm open to keeping it more flexible (as in one giant glob). I think its a much better dev experience if we type hint what they should be able to create per result rather seeing the entire chunk.

> const results = products.map(p =>
  new InlineQueryResult()
    .id(p.id)
    .title(p.name)
    .content(p.name + '\n' + p.description)
)

Yep this is exactly what i was thinking but the dev would only be able to create a single type of InlineQueryResult. `new InlineQueryResultAudio()`

Loskir:
Maybe
```ts
InlineQueryResult.article('id', {/* title, description, etc */}
```

KnorpelSenf:
I am not sure if this will be too pretty. Look at venues, they have 6 required parameters. This will be very long function calls.

Loskir:
The first param is always id, the second is an object with the rest

Loskir:
Maybe we just make it a single parameter. Basically what it'll do is to automatically set the `type` and provide autocompletion

KnorpelSenf:
This PQ already does that. It provides `queries.voice({ ... options with autocomplete ... })`.

There's no reason to do `queries.voice(InlineQueryResult.voice({ ... options with autocomplete ... }))`.

Perhaps we just don't do anything here :)

Loskir:
I don't see the point in providing a builder for the whole answer. After all, it's just an array

KnorpelSenf:
Keyboard buttons are also just an array, and yet you are asking for builders instead of just passing the raw array to `keyboard.add`.

Looking at all the inline query code that is written, we can clearly do something better than making everyone specify raw object structures. The question is not so much if we need any builder at all, but rather to which level it should go.

If we figure that every individual query result should be done without builder, that's fine. Suggesting to just ditch the plugin entirely because you personally don't need it isn't too constructive.

What we could indeed do is to provide static methods for the functional approach (`InlineQueryResult.voice`) and instance methods for the builder pattern. This would allow for both

```ts
const queries = new InlineQueryResultBuilder();
queries.voice({/*Will expect non-cached type*/})
queries.voiceCache({/**/})
```

as well as

```ts
values.map(({ id, name, description }) =>
    InlineQueryResult.article({ id, title: name, description }))
```

What do you think?

Loskir:
> Keyboard buttons are also just an array, and yet you are asking for builders instead of just passing the raw array to keyboard.add.

I'm asking for builders for the individial buttons (which we don't have)

I didn't say "ditch the plugin", I said "I'd prefer a plain array of builders over one builder that builds an array"

KnorpelSenf:
Ahh I see, I misunderstood that, sorry. Yes, that makes sense. @ssistoza what is your opinion about this?

ssistoza:
@KnorpelSenf So just to confirm I think the last builder I had just needs to be tweaked instance methods return an array of results and the static methods return just a single result?

Kind of like this:
```ts
const audioInlineQuery = InlineQueryResultBuilder.audio({ /* typing */})
const audioCached = InlineQueryResultBuilder.audioCache({ /**/})

const queries = new InlineQueryResultBuilder()
  .audio({ /* */ })
  .audioCache({ /* */ })
  .build()
```

KnorpelSenf:
> @KnorpelSenf So just to confirm I think the last builder I had just needs to be tweaked instance methods return an array of results

That's the question. Do we need this if we already have static methods?

Instead of
```ts
const queries = new InlineQueryResultBuilder()
  .audio({ /* */ })
  .audioCache({ /* */ })
  .build()
```
one could just as well do
```ts
const queries = [
  InlineQueryResult.audio({ /* ... */ }),
  InlineQueryResult.audioCache({ /* ... */ })
]
```
Assuming that people usually generate results one fly, rather than specifying manually which entries are to be returned, it would maybe be best to _only_ ship static methods. The question is if we should assume that.

ssistoza:
> Assuming that people usually generate results one fly, rather than specifying manually which entries are to be returned, it would maybe be best to only ship static methods. The question is if we should assume that.

Hmmm..I'd probably start with that assumption first, it seems that is the most prominent use-case and when needs change or grow we could add onto it. I feel like it much more important to be able to build one a single result first before being able to build an array of results.

KnorpelSenf:
Please add some tests after migrating to the approach with static methods that we discussed.

ssistoza:
Sorry for ghosting, year 2022 was a bit rough at the end.  
I can finally start picking this up again.

KnorpelSenf:
Awesome, glad you're back! LMK once this is ready for review.

KnorpelSenf:
Awesome! I'm done for today, but I feel like spending more time on the core lib now so I'm looking forward to reviewing this soon!

KnorpelSenf (changes requested):
I generally like the approach of only having static methods that generate typed objects, so the concept here is great 👌🏻

What does not work so well is that we basically leave all the complexity of creating these objects to the user. There are many funky details about inline queries and we should abstract them away.

- Some arguments are required, some are optional. Required things should be pulled in the function signature (like with bot.api).
- It should be easier to return text messages.
- Invoice message support is strange in the API, we should flatten this case
- Document mime types effectively give us two methods, so why make this a single method with a switch param?
- Same for videos

I found it easier to build on your contribution and quickly apply these changes myself, I hope you do not mind.

KnorpelSenf:
I did not adjust the tests yet and docs are missing, too. That needs to be fixed before merging. In the meantime, could you check out the updated code and LMK what you think?

src/convenience/inline_query.ts
```diff
@@ -0,0 +1,694 @@
...
+        return {
+            type: "document",
+            id,
+            title,
+            document_file_id,
+            ...options,
+        };
+    },
+    /**
+     * Builds an InlineQueryResultGame as specified by
```
KnorpelSenf, line 246:
```suggestion
     * Builds an InlineQueryResultGame object as specified by
```

KnorpelSenf:
@ssistoza all done now, please lemme know what you think :)

KnorpelSenf:
I made a mess, I misunderstood the API. IMO this proves the point that we should help users get this right, but I obviously need to fix a few things first.

KnorpelSenf:
Tests are still missing. https://github.com/grammyjs/website/pull/808 will have to be adjusted, too.

ssistoza:
It looks great, I thought we scrapped the idea of making some of the properties buildable but this makes sense since they are technically optional.

I'll get working on the tests and coverage.

KnorpelSenf:
I'll work on the docs in the meantime.

test/convenience/inline_query.test.ts
```diff
@@ -0,0 +1,2322 @@
+import { assertObjectMatch } from "https://deno.land/std@0.192.0/testing/asserts.ts";
```
KnorpelSenf, line 1:
```suggestion
import { assertObjectMatch } from "../deps.test.ts";
```

test/convenience/inline_query.test.ts
```diff
@@ -0,0 +1,2322 @@
...
+        });
+
+        describe("with venue", () => {
+            it("should build an InlineQueryResultAudio with venue in message content", () => {
+                const audio = InlineQueryResultBuilder.audio(
+                    "id",
+                    "title",
+                    "https://grammy.dev/",
+                    { caption: "cap" },
+                ).venue("Grammy Venue", 83, 136, "1 Grammy Venue", {
```
KnorpelSenf, line 112: Sorry for being pedantic about this one, please correct it in all other places 🤓 
```suggestion
                ).venue("grammY Venue", 83, 136, "1 grammY Venue", {
```

test/convenience/inline_query.test.ts
```diff
@@ -0,0 +1,2322 @@
...
+                            longitude: 136,
+                            address: "1 Grammy Venue",
+                            foursquare_id: "grammy-foursquare-id",
+                        },
+                    });
+                });
+            });
+
+            describe("with contact", () => {
+                it("should build an ZIP InlineQueryResultDocument with contact in message content", () => {
```
KnorpelSenf, line 651: nitpick:
```suggestion
                it("should build a ZIP InlineQueryResultDocument with contact in message content", () => {
```

KnorpelSenf (changes requested):
Now that is what I call an impressive test coverage, great job 👏🏻

Just nitpicks left.

KnorpelSenf (approved):
Just 11 months. What a journey. Thank you so much! 🎉

KnorpelSenf:
@all-contributors add @ssistoza for the idea, the code, and the tests!

KnorpelSenf:
This will be merged in the coming days when we're ready to release.

ssistoza:
I just wanted to thank you for all the support and your patience! It has been awhile and I really appreciate it!

KnorpelSenf:
Consider signing your commits in the future.

PR #251, opened 2022-07-31
feat: make express, koa, fastify available on Deno

KnorpelSenf:
As requested in https://t.me/grammyjs/65130

### DIFF
--- a/src/convenience/frameworks.shared.ts
+++ b/src/convenience/frameworks.shared.ts
@@ -1,3 +1,7 @@
+// deno-lint-ignore-file no-explicit-any
+
+export const SECRET_HEADER = "X-Telegram-Bot-Api-Secret-Token";
+
 /** express web framework */
 const express = (req: any, res: any) => ({
     update: Promise.resolve(req.body),
--- a/src/convenience/frameworks.shared.ts
+++ b/src/convenience/frameworks.shared.ts
@@ -1,0 +1,44 @@
+/** express web framework */
+const express = (req: any, res: any) => ({
+    update: Promise.resolve(req.body),
+    header: req.header(SECRET_HEADER),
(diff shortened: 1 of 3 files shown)


Loskir (approved):

src/convenience/frameworks.shared.ts
```diff
@@ -0,0 +1,48 @@
+// deno-lint-ignore-file no-explicit-any
+
+export const SECRET_HEADER = "X-Telegram-Bot-Api-Secret-Token";
+
+/** express web framework */
+const express = (req: any, res: any) => ({
+    update: Promise.resolve(req.body),
+    header: req.header(SECRET_HEADER),
```
darvesh, line 8: Shouldn't it be [request.get](https://expressjs.com/en/5x/api.html#req.get)?

KnorpelSenf: It's an alias, isn't it? I think `req.header` is a more accurate term for a request header than `req.get`.

darvesh: It is not anywhere in the documentation, though. Node http which Express extends, the method is `getHeaders` and `headers` is an object.

darvesh: Oh sorry, I didn't read the last line. Nvm lol

KnorpelSenf: It's stated at exactly the point in the docs that you linked to.
Screenshot_20220731-201047
↑ here

darvesh: Yeah, sorry for the trouble :)

KnorpelSenf: No worries, thanks for reviewing! :)

PR #253, opened 2022-08-01
fix: stricter chat type types

### LINKED ISSUE #252: Defining chatType to not be a channel does not guarantee ctx.msg.from
image

I should not have to do the undefined check for this to work on Deno with TS.

```typescript
bot.chatType(['private','group','supergroup']).command('test', async (ctx) => {
	if (ctx.message === undefined) {
		return
	}
	const userID = ctx.msg.from.id
```

KnorpelSenf:
Closes #252


C0rn3j (approved):
Works for me

KnorpelSenf:
@all-contributors add @C0rn3j for bug and review

PR #259, opened 2022-08-04
fix(docs): change description for keyboard.build

Loskir:


KnorpelSenf (changes requested):
Please add a test.

src/convenience/keyboard.ts
```diff
@@ -196,6 +196,14 @@ export class Keyboard {
      * end if necessary so you can specify more options in `reply_markup`.
      */
     build() {
+        if (
+            this.selective || this.one_time_keyboard || this.resize_keyboard ||
+            this.input_field_placeholder
+        ) {
+            console.warn(
```
KnorpelSenf, line 203: Should we use debug instead of console? Otherwise it won't be possible to disable the logging in case this usage is intentional.

Loskir: maybe. can we make it so it's on by default?

Loskir:
How do I test it? Should I like spy on the console function?

KnorpelSenf:
Yes, that would work well. The session tests have many examples for how to use spies.

src/convenience/keyboard.ts
KnorpelSenf: I don't think so

Loskir:
Maybe cancel this feature? It only affects old users anyway and I don't see much people complaining about it

Loskir:
Instead maybe deprecate `build` altogether?

KnorpelSenf:
Yes. We could remove it for 2.0 and just leave it as-is until then. Close this?

src/convenience/keyboard.ts
```diff
@@ -196,6 +196,14 @@ export class Keyboard {
      * end if necessary so you can specify more options in `reply_markup`.
      */
```
Loskir, line 197: This comment is outdated. We need to remove it and instead write something like "Returns the keyboard that was build. Note that it doesn't return resize_keyboard or other options that may be set. You don't usually need to use this method"

KnorpelSenf: Good point. We can use this PQ to do exactly that.

KnorpelSenf (approved):
Thank you!

KnorpelSenf:
Consider signing your commits in the future. I will merge anyway now :)

PR #268, opened 2022-08-25
fix: plugin types compatibility issue with ctx.has

### LINKED ISSUE #263: bug: Type 'MyContext' does not satisfy the constraint 'Context'. Types of property 'has' are incompatible.
Hello!

I get the error `Type 'MyContext' does not satisfy the constraint 'Context'. Types of property 'has' are incompatible. ` when I try to create a new Bot. The problem only happens with version 1.10 of Grammy and not with the 1.9.2 version.

```ts
type MyContext = HydrateFlavor<Context> & SessionFlavor<SessionData>
type MyApi = HydrateApiFlavor<Api>

export const bot = new Bot<MyContext, MyApi>(`${process.env.BOT_TOKEN}`)
```

photo_2022-08-13_12-38-37

Thanks.

Loskir:
Fixes #263

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -2116,7 +2116,7 @@ export type GameQueryContext<C extends Context> = Filter<
     "callback_query:game_short_name"
 >;
 
-type InlineQueryContextE = FilterE<'inline_query'>
+type InlineQueryContextE = FilterE<"inline_query">;
 /**
  * Type of the context object that is available inside the handlers for
  * `bot.inlineQuery`.
--- a/src/context.ts
+++ b/src/context.ts
@@ -2136,7 +2136,7 @@ type ChatTypeContextCore<T extends Chat["type"]> =
     & Record<"update", ChatTypeUpdate<T>> // ctx.update
     & ChatType<T> // ctx.chat
     & ChatTypeRecord<"msg", T> // ctx.msg
-    & AliasProps<ChatTypeUpdate<T>>;
+    & AliasProps<ChatTypeUpdate<T>>; // ctx.message etc
 /**
  * Type of the context object that is available inside the handlers for
  * `bot.chatType`.
@@ -2149,7 +2149,7 @@ type ChatTypeContextCore<T extends Chat["type"]> =
  */
 export type ChatTypeContext<C extends Context, T extends Chat["type"]> =
     & C
-    & ChatTypeContextCore<T>; // ctx.message etc
+    & ChatTypeContextCore<T>;
 type ChatTypeUpdate<T extends Chat["type"]> =
     & ChatTypeRecord<
         | "message"
--- a/src/context.ts
+++ b/src/context.ts
@@ -3,7 +3,7 @@ import { type Api, type Other as OtherApi } from "./core/api.ts";
 import { type Methods, type RawApi } from "./core/client.ts";
 import {
     type Filter,
-    type FilterE,
+    type FilterCore,
     type FilterQuery,
     matchFilter,
 } from "./filter.ts";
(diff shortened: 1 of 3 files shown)


Loskir:
This PR fixes the issue that resulted in typescript error being thrown when extending `C extends Context` `C & {something: 1}` (as discussed [in Telegram](https://t.me/grammyjs/70601))

However, it features additive implementation of `NarrowMatch` that was [proven](https://t.me/grammyjs/70653) to work not as expected.

I think that this caveat is minor compared to the original issue. If we are unable to come up with a better alternative, this should be merged until we either find a better solution or split `match` into two properties in 2.0.

The original problem is very serious as it breaks numerous plugins

src/filter.ts
```diff
@@ -433,6 +433,10 @@ type Combine<U, K extends string> = U extends unknown
     ? U & Partial<Record<Exclude<K, keyof U>, undefined>>
     : never;
 
+export type FilterE<Q extends FilterQuery> = PerformQueryE<
```
KnorpelSenf, line 436: This is a great PQ, I'd just love to have a better suffix than `E`. What does it stand for, extension maybe?

Loskir: Yes. I agree :D

KnorpelSenf: Let me throw a bunch of random names at you:
- FilterExtension
- FilterCore
- FilterAddition
- FilterCasing
- FilterShell
- FilterCover
- FilterSide
- FilterWing
- FilterAugmentation
- FilterEnhancement
- FilterAppendix
- FilterAttachment
- FilterAddendum

I think I prefer `FilterCore`. Which one do you like best?

Loskir: `FilterCore`, too

KnorpelSenf: Alright, let's rename then. Can you take care of this?

Loskir: What about exports and typedoc comments?

KnorpelSenf: I don't really think we need to export the `*Core` types. I see no reason to do this, as userland code would likely always want to use the full types. If there's ever demand for this, we can still export it in the future.

JSDoc comments are only necessary for exported stuff. Internal stuff may or may not have it, I'm mostly only doing this when things are hard to understand without comments.

Loskir: Great. Then it's done

KnorpelSenf:
@EdJoPaTo should we wait for your review?

EdJoPaTo (approved):

src/context.ts
```diff
@@ -2124,10 +2149,7 @@ export type InlineQueryContext<C extends Context> = Filter<
  */
 export type ChatTypeContext<C extends Context, T extends Chat["type"]> =
     & C
-    & Record<"update", ChatTypeUpdate<T>> // ctx.update
-    & ChatType<T> // ctx.chat
-    & ChatTypeRecord<"msg", T> // ctx.msg
-    & AliasProps<ChatTypeUpdate<T>>; // ctx.message etc
+    & ChatTypeContextCore<T>; // ctx.message etc
```
Loskir, line 2152:
```suggestion
    & ChatTypeContextCore<T>;
```

src/context.ts
```diff
@@ -2112,6 +2131,12 @@ export type InlineQueryContext<C extends Context> = Filter<
     C,
     "inline_query"
 >;
+
+type ChatTypeContextCore<T extends Chat["type"]> =
+    & Record<"update", ChatTypeUpdate<T>> // ctx.update
+    & ChatType<T> // ctx.chat
+    & ChatTypeRecord<"msg", T> // ctx.msg
+    & AliasProps<ChatTypeUpdate<T>>;
```
Loskir, line 2139:
```suggestion
    & AliasProps<ChatTypeUpdate<T>>; // ctx.message etc
```

KnorpelSenf (approved):
Thank you so much!

This fix will be shipped alongside Bot API 6.2 support.

KnorpelSenf:
@Loskir please consider signing your commits.

I'll merge anyway now, but this would be nice to have for future contributions :)

PR #273, opened 2022-08-28
feat: update to Bot API 6.2

KnorpelSenf:
Also updates deps, docs, and rearranges imports based on Deno's new feature

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -19,6 +19,7 @@ import {
     type InputMediaVideo,
     type LabeledPrice,
     type Message,
+    type MessageEntity,
     type PassportElementError,
     type Update,
     type User,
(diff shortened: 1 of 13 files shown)


roj1512:
What is Deno's new feature?

KnorpelSenf:
I have no idea when it was introduced, but Deno can now organise imports automatically through the LSP, even though https://github.com/denoland/deno/issues/12672 is still pending.

EdJoPaTo (approved):
Diff seems reasonable, still works in two of my bots so it does not seem to break anything (which I would assume based on the diff).

deptyped (approved):

src/context.ts
```diff
@@ -544,7 +545,7 @@ export class Context implements RenamedUpdate {
     }
 
     /**
-     * Context-aware alias for `api.copyMessage`. Use this method to copy messages of any kind. Service messages and invoice messages can't be copied. The method is analogous to the method forwardMessage, but the copied message doesn't have a link to the original message. Returns the MessageId of the sent message on success.
+     * Context-aware alias for `api.copyMessage`. Use this method to copy messages of any kind. Service messages and invoice messages can't be copied. A quiz poll can be copied only if the value of the field correct_option_id is known to the bot. The method is analogous to the method forwardMessage, but the copied message doesn't have a link to the original message. Returns the MessageId of the sent message on success.
```
roj1512, line 548: Do we need back tick for `correct_option_id`?

KnorpelSenf: No. As stated in point one [here](https://github.com/grammyjs/types#differences-to-the-bot-api), we don't manually add formatting to the parts of the docs that are copied from the website. (At least not yet.)

This is only done for manually written docs, such as the prefixing sentence at the beginning of this line.

Loskir (approved):

PR #277, opened 2022-09-06
chore: make deno the default formatter for VSCode

roziscoding:
as discussed in https://grammyjs.t.me/73886

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,5 +3,8 @@
     "editor.codeActionsOnSave": { "source.fixAll": true },
     "deno.enable": true,
     "deno.lint": true,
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
Can we apply this to linting and testing, too?

KnorpelSenf:
> we should make deno the default overall formatter, as opposed to only applying it to typescript

Very good point, please go ahead

Loskir:
Should we limit it to specific file types?

KnorpelSenf:
We actually format everything with Deno, so I don't see how that would help

KnorpelSenf:
@all-contributors add @roziscoding for infra and ideas :)

PR #279, opened 2022-09-16
chore(ci): type-check third-party Deno bots

wojpawlik:


.github/workflows/third_party.yml
```diff
@@ -0,0 +1,49 @@
...
+                deno-version: v1.x
+
+            - name: Create import_map.json
+              run: |
+                set -o pipefail;
+                curl --silent https://cdn.deno.land/grammy/meta/versions.json |
+                jq '
+                    (.latest / ".")[0] as $major
+                    | .versions
+                    | map(select(startswith("\($major).")) | {
```
wojpawlik, line 42:
```suggestion
                    | map(select(startswith($major + ".")) | {
```
is probably more readable

roj1512: and less Swift-ish. Nice.

.github/workflows/third_party.yml
```diff
@@ -0,0 +1,49 @@
+name: Third-party bots
```
KnorpelSenf, line 1:
```suggestion
name: Type-check third-party bots
```

KnorpelSenf (approved):
LGTM, thank you so much!

PR #291, opened 2022-10-13
feat: make init optionally unreliable

KnorpelSenf:
As requested by @roziscoding for doing a custom error handling in the browser that involves users explicitly clicking retry before handling errors

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -231,11 +231,25 @@ export class Bot<
      * Initializes the bot, i.e. fetches information about the bot itself. This
      * method is called automatically, you usually don't have to call it
      * manually.
+     *
+     * This method allows you to pass `{ unreliable: true }` which will make the
+     * method work unreliably. This means that it could initialize the bot in
+     * some random cases, and throw an error in other cases. This is because
+     * errors of the underlying API calls will not be handled correctly. Flood
+     * waits will be ignored, temporary internal server issues will cause
+     * crashes, and intermediate network outages will not be anticipated. You
+     * might want to use this option if you care about the underlying errors in
+     * order to build your own reliability system around the initialization
+     * process.
+     *
+     * @param options Optional options object
      */
-    async init() {
+    async init({ unreliable }: { unreliable?: boolean } = {}) {
         if (!this.isInited()) {
             debug("Initializing bot");
-            this.mePromise ??= withRetries(() => this.api.getMe());
+            this.mePromise ??= unreliable
+                ? this.api.getMe()
+                : withRetries(() => this.api.getMe());
             let me: UserFromGetMe;
             try {
                 me = await this.mePromise;


roziscoding (approved):

roziscoding:
A bit harsh on the JSDoc, but I really just want to get an error if an error happens, so I'll take what I can have

KnorpelSenf:
> A bit harsh on the JSDoc

It's important to be transparent when implementing something dangerous.

src/bot.ts
```diff
@@ -231,11 +231,25 @@ export class Bot<
...
+     * @param options Optional options object
      */
-    async init() {
+    async init({ unreliable }: { unreliable?: boolean } = {}) {
         if (!this.isInited()) {
             debug("Initializing bot");
-            this.mePromise ??= withRetries(() => this.api.getMe());
+            this.mePromise ??= unreliable
+                ? this.api.getMe()
+                : withRetries(() => this.api.getMe());
```
Satont, line 252: I'd like to see `withRetries` with growing sleep each retry. Currently it's only for `429` error code. But if telegram api is down then it will returns `500` and you will just flood them without sleeping.

KnorpelSenf: I think a capped exponential backoff is certainly a good idea. The runner package already does this anyway. This change should be performed regardless of whether this PQ lands.

src/bot.ts
```diff
@@ -231,11 +231,25 @@ export class Bot<
      * Initializes the bot, i.e. fetches information about the bot itself. This
      * method is called automatically, you usually don't have to call it
      * manually.
+     *
+     * This method allows you to pass `{ unreliable: true }` which will make the
```
EdJoPaTo, line 235: I think the name `unreliable` is horribly self documenting. I read the jsdoc text and its still not fully clear to me what this does. Then I looked into the code and I understood what it does but that shouldn't be needed.

I assume a name like `disableInternalErrorhandling`, `disableInternalRetries` or `useExternalErrorhandling` would be more helpful to understand what it does or what it is meant to do. Especially if its not meant for everyone, just use a long variable name there which describes more than a single adverb.

src/bot.ts
```diff
@@ -231,11 +231,25 @@ export class Bot<
...
+     * method work unreliably. This means that it could initialize the bot in
+     * some random cases, and throw an error in other cases. This is because
+     * errors of the underlying API calls will not be handled correctly. Flood
+     * waits will be ignored, temporary internal server issues will cause
+     * crashes, and intermediate network outages will not be anticipated. You
+     * might want to use this option if you care about the underlying errors in
+     * order to build your own reliability system around the initialization
+     * process.
```
EdJoPaTo, line 243: I think the reader should be ensured that this is nothing they need and its not enabled by default.

```suggestion
     * process. Defaults to `false`.
```

KnorpelSenf:
I just generally disagree with this approach. It's like a hack to avoid something which we could solve instead. There are in fact two problems and they have different solutions.

We're missing control around aborting the initialisation (solution: add an abort signal).

We're causing high load on browsers when offline (solution: backoff between requests).

I think these two things are much preferable over what I've done here. I'll close this as soon as I have a superseding PQ.

PR #293, opened 2022-10-18
feat: middleware for chosenInlineResult

### LINKED ISSUE #215: feat: Add `bot.chosenInlineResult`
this registers middleware for [chosen_inline_result](https://core.telegram.org/bots/api#choseninlineresult) that will be executed for a certain result_id(s)  

```ts
bot.chosenInlineResult('resultId', async ctx => {...})
```

MrVSiK:
Closes #215


src/composer.ts
```diff
@@ -520,6 +520,37 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
...
 
+    /**
+     * Registers middleware for the chosen inline result of an inline query that
+     * was chosen by the user and sent to their chat partner. Telegram sends an
+     * inline query to your bot whenever a user types “@your_bot_name ...” into
+     * a text field in Telegram. You bot will then receive the entered search
+     * query and can respond with a number of results (text, images, etc) that
+     * the user can pick from to send a message _via_ your bot to the respective
+     * chat. Check out https://core.telegram.org/bots/inline to read more about
+     * inline bots.
```
EdJoPaTo, line 531: The first sentence feels a bit complicated to me. Normally I assume that the first sentence gives a broad explanation about the topic why this exists so its easier to understand the area where this function is relevant. Maybe something like "Registers middleware for InlineQueryResults by the given id or ids."

Then the text should continue to explain the Result, not the inline query in general. I assume its copy pasted and work in progress?

src/context.ts
```diff
@@ -8,9 +8,9 @@ import {
     matchFilter,
 } from "./filter.ts";
 import {
+    InlineQueryResult,
     type Chat,
     type ChatPermissions,
-    type InlineQueryResult,
```
EdJoPaTo, line 13: I assume this was a mistake? It should still be a type import.

EdJoPaTo (changes requested):
Just a few things I noticed from briefly looking over this PR. Others will probably review more into the implementation.

I think your editor automatically formatted some stuff differently than Deno formats code. Running `deno fmt` should fix that.

src/composer.ts
```diff
@@ -520,6 +520,37 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
...
+     *   // Your code
+     * })
+     * ```
+     *
+     * @param resultId An id or array of ids
+     * @param middleware The middleware to register
+     */
+    chosenInlineResult(resultId: MaybeArray<string | RegExp>,
+        ...middleware: Array<InlineQueryMiddleware<C>>
+    ): Composer<InlineQueryContext<C>> {
```
KnorpelSenf, line 550: This is incorrect, it should be a new type called `ChosenInlineResultContext` which reflects the chosen inline result. An inline query is not present in these updates.

KnorpelSenf (changes requested):
These changes look mostly good, thanks!

Please run the formatter in order to remove all the whitespace changes.

Also, you'll need to add a new alias for a context type if you want your changes to be correct:

src/context.ts
MrVSiK: Yep, I must have missed that. I'll update it

src/composer.ts
MrVSiK: Gotcha. I'll add it in a new commit

src/context.ts
```diff
@@ -10,7 +10,7 @@ import {
 import {
     type Chat,
     type ChatPermissions,
-    type InlineQueryResult,
+    InlineQueryResult,
```
EdJoPaTo, line 13: removing the type keyword should also be reverted.

```suggestion
    type InlineQueryResult,
```

Loskir:
Please change it to ChosenInlineResultContext, not just ChosenInlineResult

src/composer.ts
```diff
@@ -520,6 +521,35 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
         return this.filter(Context.has.inlineQuery(trigger), ...middleware);
     }
 
+    /**
+     * Registers middleware for InlineQueryResults by the given id or ids.
+     * Telegram sends an inline query to your bot whenever a user types
```
EdJoPaTo, line 526: I'm not entirely happy with this description yet. First it tells you something about InlineQueryResult, then it changes the topic explaining what Inline queries are and then about the response. I think it should be in a different order. The introducing sentence, then it does something for the chosen results, then a brief "chosen results of what, chosen results of inline queries".

Not sure if my "this currently has a strange order" feeling is well enough explained with this comment?

MrVSiK: How about something like this?
`* Registers middleware for the ChosenInlineResult by the given id or ids.
     * ChosenInlineResult represents a result of an inline query that was
     * chosen by the user and sent to their chat partner. Check out
     * https://core.telegram.org/bots/api#choseninlineresult to read more
     * about chosen inline results.`

KnorpelSenf:
Looks pretty good already, will do a proper review soon!

MrVSiK:
Will do

KnorpelSenf:
Was a little busy recently, sorry. For the ongoing delay. Will look into it soon!

KnorpelSenf (approved):
Looks great! Thanks!

KnorpelSenf:
@all-contributors please add @MrVSiK for code!

KnorpelSenf:
This will be merged soon alongside https://github.com/grammyjs/grammY/pull/276#pullrequestreview-1103163328.

MrVSiK:
Awesome. Thank you so much!!

src/context.ts
```diff
@@ -2254,6 +2288,22 @@ type InlineQueryContextCore = FilterCore<"inline_query">;
...
 
+type ChosenInlineResultContextCore = FilterCore<"chosen_inline_result">;
+/**
+ * Type of the context object that is available inside the handlers for
+ * `bot.chosenInlineResult`.
+ *
+ * This helper type can be used to narrow down context objects the same way how
+ * annotate `bot.chosenInlineResult` does it. This allows you to context objects in
+ * middleware that is not directly passed to `bot.chosenInlineResult`, hence not
+ * inferring the correct type automatically. That way, handlers can be defined
```
EdJoPaTo, line 2299: This is not relevant to this specific PR as other doc comments also include this, but I think its easier to read then. Maybe it should be improved in another PR?

```suggestion
 * annotate `bot.chosenInlineResult` does it. This allows you to give context to objects in
 * middleware that is not directly passed to `bot.chosenInlineResult`, hence not
 * inferring the correct type automatically. That way, handlers can be defined
```

EdJoPaTo (approved):

src/context.ts
KnorpelSenf: I'll update the docs string in a later PQ, thanks for pointing this out :)

KnorpelSenf:
Consider signing your commits in the future.

PR #296, opened 2022-10-20
feat: add delete_webhook_on_polling start option

d0ctr:
I want to propose a new option `delete_webhook_on_polling` for the `PollingOptions`.

If this options is set to False, Bot.start() method will skip call to `deleteWebhook` forbidding long polling to start if somewhere webhook is set.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -46,7 +46,7 @@ export interface PollingOptions {
     /**
      * Pass False to prevent webhook deletion on the start of long polling.
      * Note that this will result in immediate error response from Telegram
-     * server is webhook is set somewhere else.
+     * server if webhook is set somewhere else.
      */
     delete_webhook_on_polling?: boolean;
     /**
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -44,6 +44,12 @@ export interface PollingOptions {
      */
     drop_pending_updates?: boolean;
     /**
+     * Pass False to prevent webhook deletion on the start of long polling.
+     * Note that this will result in immediate error response from Telegram
+     * server is webhook is set somewhere else.
+     */
+    delete_webhook_on_polling?: boolean;
+    /**
      * A callback function that is useful for logging (or setting up middleware
      * if you did not do this before). It will be executed after the setup of
      * the bot has completed, and immediately before the first updates are being
@@ -120,6 +126,7 @@ export class Bot<
     private pollingRunning = false;
     private pollingAbortController: AbortController | undefined;
     private lastTriedUpdateId = 0;
+    private deleteWebhookOnPolling = true;
 
     /**
      * Gives you full access to the Telegram Bot API.
@@ -358,11 +365,17 @@ a known bot info object.",
             debug("Simple long polling already running!");
             return;
         }
-        await withRetries(() =>
-            this.api.deleteWebhook({
-                drop_pending_updates: options?.drop_pending_updates,
-            })
-        );
+
+        if (options?.delete_webhook_on_polling === false) {
+            await withRetries(() =>
+                this.api.deleteWebhook({
+                    drop_pending_updates: options?.drop_pending_updates,
+                })
+            );
+        }
+        else {
+            debug("deleteWebhook is skipped");
+        }
 
         // All async ops of setup complete, run callback
         await options?.onStart?.(this.botInfo);


KnorpelSenf:
This sounds like an https://xyproblem.info. What are you actually trying to solve with this?

The straightforward way to do achieve the same goal is call getWebhookInfo in order to inspect whether or not a webhook was set. Why is this not an option in your use case?

d0ctr:
@KnorpelSenf, in the first place it was actually unexpected for me that starting long polling in grammY deletes webhook. I think that not doing it is more obvious (and a bit more consistent) approach but since this a feature of grammY, I would want to have more configurability here.

KnorpelSenf:
You are neither providing a problem description nor a use case. I'm not willing to merge this if this change doesn't solve any problem and the configuration isn't applicable in any scenario.

Calling start is supposed to start the bot. Unnecessarily throwing errors in some cases that in turn must be investigated and fixed manually is not a useful design. The grammY runner illustrates this, as it only performs polling at maximal performance without offering the convenience of bot.start which “just works.” You'd be surprised how many times people are running into problems because of this.

How are you going to handle the case in which pending updates should be dropped but the webhook should not be deleted?

d0ctr:
So my actual case here is that I use the same bot for testing locally (polling) and for production (webhook). I want to make sure that I will not disable bot in prod unintentionally.

I agree with you that it is convenient that `bot.start()` "just works". However, this configuration is optional and by using it I (or anyone else) agree that from this point there may be an exception.

As for your question, I don't see that your case is relevant. If person intends to drop pending updates and start polling by `bot.start()` then they must understand that it is not possible without deleting webhook. And vice versa, if they don't intend to delete webhook then they can't drop pending updates.

KnorpelSenf:
> So my actual case here is that I use the same bot for testing locally (polling) and for production (webhook).

Why?

d0ctr:
Bc I don't want to create additional environment (like db), I am good as it is now.

KnorpelSenf:
That's unrelated. You can have two bots on the one end but the same code and database on the other end.

d0ctr:
Okay, looks like this won't turn in any direction that will be fine with both me and you. Therefore, I close this.

KnorpelSenf:
This isn't about the two of us, it's about finding the right solution to a problem. Using the same credentials in both dev and prod environments is a flawed approach, and the best way to mitigate this is always to separate the environments cleanly

If you want to share the database, that's fine, sometimes I do it, too, but I guarantee you that as soon as your project is complex enough to turn this into a problem, you're going to look into separering the environments here, too—rather than branching out in all places of your code that touch the database. The same is true for bot tokens.

See it that way: if grammY would have this option, you'd be using it and then you'd not solve the underlying problem. Now that grammY doesn't support or encourage such setups with clashing credentials, you'll need to properly disentangle your dev-prod dependencies, which is much better :)

Even though we didn't reach consensus on this one, please don't let that stop you from your next contribution.

EdJoPaTo:
@KnorpelSenf, you are right with all your points but I think you are missing a major point relevant to grammY as a Telegram Bot API. The most important problem in my eyes is the fact that the production and the development environment would steal (Telegram-)updates from each other.

(Simplified: ) Assuming both the production bot and the dev bot are running. Both want to get updates from Telegram. Telegram will send some updates. But you do not know which are which. Are these updates relevant for the codebase which polled them (for example prod) or are they relevant for the other (for example dev)? Very likely one will steal the updates of the other.

This means having two codebases for the same bot would result in lost messages for production which is certainly not what is wanted.

Yes you could push (Telegram-)updates all into the same queue, database or whatever and work from there with filters, but in that case you still only need one bot codebase which does that, the rest is independent from the bot itself.

So when ignoring all the credentials, databases dev / production stuff, its still a bad thing to approach for the Telegram Bot API itself. Either messages are lost for their relevant codebase or it's an independent problem of the API itself.
That would be my reason to not allow a change like this, it would compromise a stable bot experience.

---

That's my way of handling production/dev:

Personally I have a bunch of production Bot API keys for the individual production bots and two testing bot API Keys (one for me, one for my testers). I will use the latter two for testing without interfering with production. For some bots I clone the data down from the production bot to work on the same data (without manipulating it), but most commonly I have a clean testing environment separate from the production environment.

---

As a final note, the same as @KnorpelSenf already said in the last sentence: Its awesome to see your initiative to pick up the code, improve it and provide a Pull Request here. Sometimes I am overlooking some things while providing a PR. That's why reviews exist, to think with multiple people about the same problem. Thanks again for your initiative @d0ctr!

KnorpelSenf:
> The most important problem in my eyes is the fact that the production and the development environment would steal (Telegram-)updates from each other.

Can you elaborate how that could happen for developement purposes, you're surely running an instance locally and your production system elsewhere?

PR #297, opened 2022-10-21
feat: webhook adapter for next.js api routes

waptik:
This adapter will enable bots to work smoothly when using next.js api routes as webhook endpoint. See https://t.me/grammyjs/86659 for how it started.

### DIFF
--- a/src/convenience/frameworks.node.ts
+++ b/src/convenience/frameworks.node.ts
@@ -69,7 +69,7 @@ const azure = (context: any, req: any) => ({
 /** Vercel/Next.js Serverless Functions */
 const vercel = (req: any, res: any) => ({
     update: Promise.resolve(req.body),
-    header: req.headers[SECRET_HEADER] as string,
+    header: req.headers[SECRET_HEADER],
     end: () => res.end(),
     respond: (json: string) => res.status(200).json(json),
     unauthorized: () => res.status(401).send("secret token is wrong"),
--- a/src/convenience/frameworks.node.ts
+++ b/src/convenience/frameworks.node.ts
@@ -53,11 +53,10 @@ const awsLambda = (event: any, _context: any, callback: any) => ({
 const azure = (context: any, req: any) => ({
     update: Promise.resolve(req.body),
     header: context.res.headers[SECRET_HEADER],
-    end: () =>
-        context.res = {
-            status: 200,
-            body: "",
-        },
+    end: () => (context.res = {
+        status: 200,
+        body: "",
+    }),
     respond: (json: string) => {
         context.res.set("Content-Type", "application/json");
         context.res.send(json);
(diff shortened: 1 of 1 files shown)


src/convenience/frameworks.node.ts
```diff
@@ -67,13 +66,23 @@ const azure = (context: any, req: any) => ({
     },
 });
 
+/** Vercel/Next.js Serverless Functions */
+const vercel = (req: any, res: any) => ({
+    update: Promise.resolve(req.body),
+    header: req.headers[SECRET_HEADER] as string,
```
KnorpelSenf, line 72:
```suggestion
    header: req.headers[SECRET_HEADER],
```

src/convenience/frameworks.node.ts
```diff
@@ -53,11 +53,10 @@ const awsLambda = (event: any, _context: any, callback: any) => ({
 const azure = (context: any, req: any) => ({
     update: Promise.resolve(req.body),
     header: context.res.headers[SECRET_HEADER],
-    end: () =>
-        context.res = {
-            status: 200,
-            body: "",
-        },
+    end: () => (context.res = {
```
KnorpelSenf, line 56: This change is useless and only makes the code harder to read, please revert

src/convenience/frameworks.node.ts
```diff
@@ -67,13 +66,23 @@ const azure = (context: any, req: any) => ({
...
+});
+
 // please open a PR if you want to add another
 export const adapters = {
     http,
     https: http,
     worktop,
     "aws-lambda": awsLambda,
     azure,
+    vercel,
```
KnorpelSenf, line 85: This name is misleading. Not all bots hosted on Vercel should use this adapter.

Is it possible to use this adapter when using Next.js outside of Vercel?

KnorpelSenf (changes requested):

src/convenience/frameworks.node.ts
waptik: It was generated after running `deno fmt` but sure.

src/convenience/frameworks.node.ts
waptik: Yes, it's possible. So should i name it `next.js` or `next` or `nextjs`?

src/convenience/frameworks.node.ts
KnorpelSenf: Oh

src/convenience/frameworks.node.ts
KnorpelSenf: Either `nextjs` or `next.js` or `next-js` is fine by me. Just `next` would be confusing. Pick whichever one you prefer.

src/convenience/frameworks.node.ts
waptik: So what should i do? It still formats after i save the file

KnorpelSenf: Then we can perform the changes that way. I'll try to reproduce it and then open a bug report for the Deno formatter. Don't worry about it for now.

KnorpelSenf (approved):
Thank you!

src/convenience/frameworks.node.ts
waptik: Okay, thanks!

PR #318, opened 2022-11-22
[FEAT] New web bundle

roziscoding:
New bundling script generates `bundling/bundle` folder which can be `npm publish`ed creating `@grammyjs/web`.

The `package.json` for the new package is automatically generated from the current `package.json` to avoid version mismatches.

Tested with vite, but not with CF workers.

Co-authored-by: Wojciech Pawlik <woj.pawlik@gmail.com>


KnorpelSenf (changes requested):
1. Can we retain the README section about the CDN? We don't shut it down.
2. Should the Netlify CDN config be changed to redirect to https://bundle.deno.dev?

wojpawlik:
Constantly rebuilding every release is a pain. Let's just not upload new versions to CDN.

roziscoding:
> 1. Can we retain the README section about the CDN? We don't shut it down.
> 2. Should the Netlify CDN config be changed to redirect to https://bundle.deno.dev?

1. Sure. Should we say it's not the preferred way of doing this or something?
2. I think that makes sense, but it requires 1 to explain the new URL format, right? Because we don't have separate bundles anymore

wojpawlik:
So, let's update CDN to be just a redirect to bundle.deno.dev or sth, and update the docs afterwards.

KnorpelSenf:
And yet there's no good reason to break people's setups, is there? It is a one-time action to set it up and it will require zero maintenance.

wojpawlik:
Sure, feel free to set up a redirect. Documenting it would just cause unnecessary confusion tho.

.github/workflows/release.yml
```diff
@@ -3,7 +3,7 @@ name: Main
 on:
     push:
         tags:
-            - "*"
+            - "v1.*"
```
KnorpelSenf, line 6: I don't want to miss this when going for 2.0.
```suggestion
            - "v*"
```

README.md
```diff
@@ -122,13 +122,14 @@ You may also be interested in [why we support Deno](https://grammy.dev/resources
...
 
-You can download them from the URL `https://get.grammy.dev/[ES version lowercased]@[grammY version including v-prefix].js`.
-For example, the most recent source on `main` in ES6 is available from <https://get.grammy.dev/es6@dev.js>.
+You can just install it with npm or yarn and use it in place of the `grammy` package.
 
-So far, this is mainly useful for running bots on Cloudflare Workers.
+If you need a CDN, you can use jsDelivr or UNPKG like with the following URLs:
+
+jsDelivr: `https://cdn.jsdelivr.net/npm/@grammyjs/web@<grammy-version>`
+UNPKG: `https://unpkg.com/@grammyjs/web@<grammy-version>`
```
KnorpelSenf, line 132: Should we keep this?
```suggestion

grammY: `https://get.grammy.dev/es6@<grammy-version>.js`
jsDelivr: `https://cdn.jsdelivr.net/npm/@grammyjs/web@<grammy-version>`
UNPKG: `https://unpkg.com/@grammyjs/web@<grammy-version>`
```

bundling/bundle-browser.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+set -eu
+
+# Cloudflare Workers don't support top-level await
+sed -i~ 's/^if (isDeno)/if (false)/' ../src/platform.deno.ts
```
KnorpelSenf, line 5: I don't like this, mostly because regex on source is very ugly and hard to maintain. The whole point is to have one clean file per platform. If we want to support different platforms that require different platform-specific code, we should create a new file and `cp` our way through.

bundling/bundle-browser.sh
```diff
@@ -0,0 +1,38 @@
...
+jq '. + {
+  "name": "@grammyjs/web",
+  "dependencies": {},
+  "scripts": {},
+}' <../package.json >bundle/package.json
+
+echo "Copying node files"
+cp -R ../out bundle/
+cp ../LICENSE bundle/
+cp ../README.md bundle/
```
KnorpelSenf, line 31:
```suggestion
cp -rv ../LICENSE ../README.md ../out bundle/
```

bundling/bundle-browser.sh
```diff
@@ -0,0 +1,38 @@
...
+shopt -s globstar
+rm bundle/**/*.js
```
KnorpelSenf, line 33: Wouldn't it be better to explicitly copy over the files we need, rather than copying over everything and then deleting some specific parts of it?

KnorpelSenf (changes requested):

bundling/bundle-browser.sh
roziscoding: It would, but glob patterns are very sketchy and I couldn't get cp to reproduce the existing folder structure but ignore js files

KnorpelSenf: `rsync`?

bundling/bundle-browser.sh
roziscoding: can we rewrite to prevent top level awaits?

KnorpelSenf: If we provide a new file `platform.web.ts`, we can write it w/o TLA, yes.

.github/workflows/release.yml
wojpawlik: You don't want to accidentally push a v2 beta as stable `latest` either

README.md
wojpawlik: Definitely not if it's gonna serve top-level await.

.github/workflows/release.yml
KnorpelSenf: Hm. Fair enough. Let's keep it this way.

README.md
KnorpelSenf: Why not? Browsers support this.

bundling/bundle-browser.sh
roziscoding: I haven't used it before. 
I'll look into it

README.md
roziscoding: Maybe we should specify that they are different budles?
Since gramY CDN isn't being fed this new one

KnorpelSenf:
I want to look into this in my holidays

bundling/bundle-browser.sh
```diff
@@ -0,0 +1,36 @@
...
+  "scripts": {},
+}' <../package.json >bundle/package.json
+
+echo "Copying node files"
+cp -rv ../LICENSE ../README.md ../out bundle/
+shopt -s globstar
+rm bundle/**/*.js
+
+echo "Bundling source"
+deno bundle ../src/mod.ts bundle/out/mod.js
```
KnorpelSenf, line 34: … then there is https://deno.land/manual@v1.29.1/tools/bundler#bundling-for-the-web:

_The output of deno bundle is intended for consumption in Deno and not for use in a web browser or other runtimes. That said, depending on the input it may work in other environments._

_If you wish to bundle for the web, we recommend other solutions such as [esbuild](https://esbuild.github.io/)._

wojpawlik: Esbuild wasn't smart enough to omit `if (false)` when I last tired it. Unlike SWC used by `deno bundle`.

KnorpelSenf: `if (false)` is a big hack that should be avoided anyway. Do you remember how we went from bash scripts to d2n? I'd like to make a similar move here, and end up using good tools rather than writing custom scripts.

roziscoding:
Closing in favor of #339

PR #325, opened 2022-12-07
 Fix: missing chat action added "choose_sticker"

abdoo9:
I noticed that  "choose_sticker" chat action is missing so I made this pull request to add it


KnorpelSenf:
@all-contributors add @abdoo9 for bug and code!

PR #329, opened 2022-12-10
feat: add SvelteKit adapter

KnorpelSenf:
Adds support for https://kit.svelte.dev/

### DIFF
--- a/src/convenience/frameworks.node.ts
+++ b/src/convenience/frameworks.node.ts
@@ -75,6 +75,15 @@ const nextJs = (req: any, res: any) => ({
     unauthorized: () => res.status(401).send("secret token is wrong"),
 });
 
+/** Sveltekit Serverless Functions */
+const sveltekit = (event: any) => ({
+    update: Promise.resolve(event.request.json()),
+    header: event.request.headers[SECRET_HEADER],
+    end: () => event.respond.end(),
+    respond: (json: string) => event.respond.status(200).json(json),
+    unauthorized: () => event.request.status(401).send("secret token is wrong"),
+});
+
 // please open a PR if you want to add another
 export const adapters = {
     http,
@@ -83,6 +92,7 @@ export const adapters = {
     "aws-lambda": awsLambda,
     azure,
     "next-js": nextJs,
+    sveltekit,
     ...sharedAdapters,
 };
 export const defaultAdapter = "express";
--- a/src/convenience/frameworks.node.ts
+++ b/src/convenience/frameworks.node.ts
@@ -76,13 +76,35 @@ const nextJs = (req: any, res: any) => ({
 });
 
 /** Sveltekit Serverless Functions */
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
I assume that `response` and `unauthorized` will also fail

ak4zh:
I think we need to create `new Response()` similar to `std/http`.

KnorpelSenf:
@KnightNiwrem since you already have a setup running, do you want to fix this up?

ak4zh:
This seems to work, I copied the missing parts from stdHttp adapter

```js
const sveltekit = (event: any) => {
    let resolveResponse: (res: Response) => void;
    return {
        update: Promise.resolve(event.request.json()),
        header: event.request.headers[SECRET_HEADER],
        end: () => {
            if (resolveResponse) resolveResponse(new Response());
        },
        respond: (json: string) => {
            if (resolveResponse) {
                const res = new Response(json, {
                    headers: { "Content-Type": "application/json" },
                });
                resolveResponse(res);
            }
        },
        unauthorized: () => {
            if (resolveResponse) {
                const res = new Response("secret token is wrong", {
                    status: 401,
                });
                resolveResponse(res);
            }
        },
        handlerReturn: new Promise((resolve) => {
            resolveResponse = resolve;
        }),
    }
};
```

KnightNiwrem:
> This seems to work, I copied the missing parts from stdHttp adapter
> 
> ```js
> const sveltekit = (event: any) => {
>     let resolveResponse: (res: Response) => void;
>     return {
>         update: Promise.resolve(event.request.json()),
>         header: event.request.headers[SECRET_HEADER],
>         end: () => {
>             if (resolveResponse) resolveResponse(new Response());
>         },
>         respond: (json: string) => {
>             if (resolveResponse) {
>                 const res = new Response(json, {
>                     headers: { "Content-Type": "application/json" },
>                 });
>                 resolveResponse(res);
>             }
>         },
>         unauthorized: () => {
>             if (resolveResponse) {
>                 const res = new Response("secret token is wrong", {
>                     status: 401,
>                 });
>                 resolveResponse(res);
>             }
>         },
>         handlerReturn: new Promise((resolve) => {
>             resolveResponse = resolve;
>         }),
>     }
> };
> ```

Would `const sveltekit = ({ request: Request }) => {` be a better option, since we don't intend to use anything else in `event` other than `event.request`.

KnorpelSenf:
That code doesn't parse, but yes, it'd be great to have the type safety and do `event: { request: Request }` or `{ request }: { request: Request }`, which would also make clear what we depend on.

KnorpelSenf:
I'll update the code in a sec

KnightNiwrem:
Screen Shot 2022-12-11 at 11 25 09 AM

Verified

KnightNiwrem (approved):

KnorpelSenf:
@all-contributors add @ak4zh for review

PR #330, opened 2022-12-15
chore: ignore deno.lock using the config

roj1512:


KnorpelSenf:
Can you expand the scope of this PQ and ensure Deno 1.29 compatibility? This includes TS 4.9.3 support.

deno.jsonc
```diff
@@ -32,5 +32,6 @@
                 "./bundling/bundles"
             ]
         }
-    }
+    },
+    "lock": false
```
KnorpelSenf, line 36: Can you make this the first line in the config file? It gets a bit lost here

KnorpelSenf:
Pin d2n until @wojpawlik fixes his setup to prevent publishing broken versions

wojpawlik:
I disabled auto updates until deno-bin is fixed.

roj1512:
@KnorpelSenf We're passing.

KnorpelSenf (approved):
TYSM!

PR #339, opened 2022-12-27
feat: built-in web bundle

roziscoding:
Besided being called inside `bundle-all.sh`, `deno task bundle-web` can also be used to create the new bundle.
Two files will be created inside the `out` folder: web.js and web.d.ts

The `web.d.ts` file only re-exports everything from `out/mod.d.ts`. This
file is needed because typescript refuses to search for types for the
`web.js` file in anywhere other than a `web.d.ts` placed next to it.
That behavior should be overriden by the `typesVersions` property in
`package.json`, but for some reason it's not. The `typesVersions`
property is necessary to make typescript detect the `/web` entrypoint,
though.

### DIFF
--- a/bundling/bundle-web.ts
+++ b/bundling/bundle-web.ts
@@ -1,0 +1,24 @@
+import { bundle } from "https://deno.land/x/emit@0.12.0/mod.ts";
+
+// Parse args
+const [release, source = `https://deno.land/x/grammy@${release}/mod.ts`] =
+    Deno.args;
+if (!release) throw new Error("No release specified!");
+
+// Bundle code
+const { code: bundledCode } = await bundle(source, {
+    compilerOptions: {
+        sourceMap: false,
+        inlineSources: false,
+        inlineSourceMap: false,
+    },
+});
+
+await Deno.writeTextFile(
+    "../out/web.js",
+    bundledCode.replace(/\/\/# sourceMappingURL=.*\n/, ""),
+);
+await Deno.writeTextFile(
+    "../out/web.d.ts",
+    'export * from "./mod"',
+);
--- a/bundling/bundle-web.ts
+++ b/bundling/bundle-web.ts
@@ -13,10 +13,10 @@ const load = (specifier: string) => {
         const baseLength = specifier.length - ".deno.ts".length;
         specifier = specifier.substring(0, baseLength) + ".web.ts";
     }
-    console.log(specifier); // TODO: remove
     return cache.load(specifier);
 };
 
+console.log(`Bundling version '${release}' from ${source} ...`);
 // Bundle code
 const { code: bundledCode } = await bundle(source, {
     load,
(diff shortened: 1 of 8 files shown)


src/types.web.ts
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * Check out the
+     * [documentation](https://grammy.dev/guide/files.html#uploading-your-own-file)
+     * on sending files with `InputFile`.
+     */
+    public readonly filename?: string;
+    /**
+     * Constructs an `InputFile` that can be used in the API to send files.
+     *
+     * @param file A path to a local file or a `Buffer` or a `ReadableStream` that specifies the file data
```
KnorpelSenf, line 40: Docs are wrong here

KnorpelSenf:
@roziscoding LMK if you need help fixing up CI. I'll be back later today or maybe tomorrow. I'll also want to test everything thoroughly once more.

src/types.web.ts
roziscoding: I see. What about this?

```suggestion
     * @param file A URL of a file or a `Blob` or any form of `Uint8Array` that specifies the file data
```

roziscoding:
I guess we don't need to backport `platform.web.ts`, do we?
The web code has nothing to do with the backport code

KnorpelSenf (commented):
LGTM but will test all possible combinations manually before approving.

- from npm tarball
- from npm link
- from GitHub install

- from a JS file
- from an MJS file
- from a CJS file
- from a TS file
- from TS build output

- in Deno
- in Node
- in a browser script
- in a browser module
- in a browser with Deno tooling 
- in a browser with Node tooling 

- compile time
- runtime

roziscoding:
Will ESM work on browser scripts?

KnorpelSenf:
No, that's one of the impossible combinations, I think. I'll test it.

KnorpelSenf (approved):
Amazing changes! I tried this and one can now easily `import { Bot } from "grammy/web"` in order to run a bot inside the browser. Thank you so much!

KnorpelSenf:
@all-contributors add @roziscoding for this idea!

bundling/bundle-web.ts
```diff
@@ -0,0 +1,41 @@
...
+if (!release) throw new Error("No release specified!");
+
+// Rewrite imports from .deno.ts to .web.ts
+const cache = createCache();
+const load = (specifier: string) => {
+    if (specifier.endsWith(".deno.ts")) {
+        const baseLength = specifier.length - ".deno.ts".length;
+        specifier = specifier.substring(0, baseLength) + ".web.ts";
+    }
+    return cache.load(specifier);
```
wojpawlik, line 16:
```suggestion
    return cache.load(specifier.replace(/\.deno\.ts$/, ".web.ts"));
```

KnorpelSenf:
> @KnorpelSenf
> 
> @roziscoding already contributed before to ideas

oh … right, good

bundling/bundle-web.ts
KnorpelSenf: Do you think that code is more readable?

roziscoding:
Happy to see this merged! :D

PR #340, opened 2022-12-27
feat: add web-specific code

KnorpelSenf:
Adds code that is specific to the web. This means:
- the supported web frameworks differ
- the supported input file types differ
- there is no need for TLA because the only time we need this is when running inside Deno and have to query for permission to read the DEBUG env var

This build should run in all evergreen browsers as well as Cloudflare Workers and other edge runtimes.


roziscoding:
I'm guessing we have some non-shared dependencies then?
I'll wait until later™ for the explanation to review further, but I see no red flags right now

roziscoding:
Oh. I see. No TLA now
But it being a web version, do we need to get rid of TLA?

KnorpelSenf:
CFW

KnorpelSenf:
@roziscoding please take another look

roziscoding (approved):
Seems about right :)

roziscoding:
Wait
The changes in package.json are missing

KnorpelSenf:
I'm targeting your PQ

KnorpelSenf:
This isn't supposed to be merged into main

roziscoding:
Oh
Then let's f'ing go :D

KnorpelSenf:
*not directly

roziscoding:
Can I do the honors then?

KnorpelSenf:
Absolutely, go ahead 🤓

Issue #345, opened 2022-12-27
Multi Sessions and Conversations


receter:
I don’t manage to get conversations to work with multi sessions. It does work with normal configuration (not session type "multi") but if I try to change it to multi it stops working.

```
  bot.use(session({
    type: "multi",
    user: {
      storage: new MemorySessionStorage(),
      initial: createInitialSessionDataUser,
      getSessionKey: (ctx) => ctx.from?.id.toString()
    }
  }));

  bot.use(conversations());
  bot.use(createConversation(conversationA));
```

This works though:

```
  bot.use(session({
    initial() {
      // return empty object for now
      return {};
    },
  }));

  bot.use(conversations());
  bot.use(createConversation(conversationA));
```

KnorpelSenf:
Remember that when you say it “doesn't work”, we cannot login to your computer and figure out what's wrong. You need to explain what you are expecting to happen and what is actually happening :)

KnorpelSenf:
Are there any errors? What is the behaviour? Does it compile?

receter:
Thanks for your fast reply. I really like the library and the docs, I had a lot of fun getting into it already!

If I use lazySessions does throw an error. With just session it does not throw but the conversations do not seem to work either.

Here is a demo repository with a minimal example of the issue: https://github.com/receter/grammy-multi-sessions-conversations

KnorpelSenf:
Multi sessions require you to specify all fragments of the session upfront. This means you'll also have to specify `conversation` here: https://github.com/receter/grammy-multi-sessions-conversations/blob/7a31d2d19babfa990d41f668e4304711934fe497/src/configureTelegramBot.js#L18-L22

This should probably go in the conversation docs.

receter:
This should work right?
```javascript
  bot.use(lazySession({
    type: "multi",
    test: {},
    conversation: {}
  }));
```
It still throws:
```
  error: TypeError: Cannot read properties of undefined (reading 'conversation')
      at Object.session (/Users/andreasriedmuller/PhpstormProjects/grammy-multi-sessions-conversations/node_modules/@grammyjs/conversations/out/conversation.js:259:36)
```

receter:
But this seems to be an issue with `lazySession` only, it does work with `session`.
```javascript
  // works
  bot.use(session({
    type: "multi",
    test: {},
    conversation: {}
  }));
```

KnorpelSenf:
[...]

PR #347, opened 2022-12-27
feat: protect against lazy multi sessions mistake

KnorpelSenf:
Prevents non-TS people from making the same mistake as reported in #345.


KnorpelSenf:
@receter I'd be interested in your input, too.

Satont (approved):

PR #349, opened 2022-12-30
feat: add support for Bot API 6.4

KnorpelSenf:
https://core.telegram.org/bots/api#december-30-2022

There is a small breaking change in this release. Two paramters in `editFormTopic` have become optional. As a result, they were removed from the respective method in `Api`. This is a known disadvantage of the convenience of having parameters in the function signature.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
         "contribs": "all-contributors"
     },
     "dependencies": {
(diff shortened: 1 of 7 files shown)


roziscoding (approved):
LGTM

PR #358, opened 2023-01-16
feat: add native CloudFlare Workers adapter

KnorpelSenf:
So far, only worktop is supported. With this PQ, we can do
```ts
addEventListener('fetch', webhookCallback(bot, "cloudflare"))
```


KnorpelSenf:
Using https://t.me/grammyjs/104281 as an approval, the person does not have a GH account

PR #360, opened 2023-01-29
feat: add Cloudflare Workers module and hono adapters

KnorpelSenf:
Props to the person without name on Telegram, see https://t.me/grammyjs/104653.

The holo adapter is duplicated because the `.web.ts` files currently are not included in the Deno and Node builds. Should we add these dependencies in order to be DRY here?


KnorpelSenf:
We added the dependency between the files. They were included in the releases anyway. The holo adapter now isn't duplicated anymore.

src/convenience/frameworks.web.ts
```diff
@@ -34,6 +34,61 @@ const cloudflare = (
     };
 };
 
+/** Native CloudFlare workers (module worker) */
+const cloudflare_module = (
```
KnorpelSenf, line 38:
```suggestion
const cloudflareModule = (
```

src/convenience/frameworks.web.ts
```diff
@@ -43,4 +98,9 @@ const worktop = (req: any, res: any) => ({
     unauthorized: () => res.send(401, "secret token is wrong"),
 });
 
-export const adapters = { cloudflare, worktop };
+export const adapters = {
+    cloudflare,
+    "cloudflare-mod": cloudflare_module,
```
KnorpelSenf, line 103:
```suggestion
    "cloudflare-mod": cloudflareModule,
```

dcdunkan (approved):
LGTM

KnorpelSenf:
Thanks for the review! ❤️

PR #365, opened 2023-02-03
fix: make reply* methods contact user upon chat join requests

KnorpelSenf:
Is this desired?

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -549,7 +549,10 @@ export class Context implements RenamedUpdate {
         signal?: AbortSignal,
     ) {
         return this.api.sendMessage(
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
Note that this would be a breaking change, so it cannot be merged before 2.0.

KnorpelSenf:
Now that I think about this, this suggestion is not very intuitive. The point of `ctx.reply` is to send a message in the same chat. This would completely go against this idea.

One has to use
```ts
await ctx.api.sendMesasge(ctx.chatJoinRequest.user_chat_id, "Hi!")
```
to contact a user after a chat join request.

I will close this. Please comment if you have strong opinions about picking this up again.

PR #366, opened 2023-02-05
feat: add `has` and `readAll` methods to `StorageAdapter` interface

roziscoding:
These methods will be used mainly by the chat-members and the broadcast
plugins


src/convenience/session.ts
```diff
@@ -85,6 +85,14 @@ export interface StorageAdapter<T> {
...
      */
     delete: (key: string) => MaybePromise<void>;
+    /**
+     * Checks whether a key exists in the storage
+     */
+    has?: (key: string) => MaybePromise<boolean>;
+    /**
+     * Lists all keys
+     */
+    readAll?: () => Iterable<T> | AsyncIterable<T>;
```
KnorpelSenf, line 95: It does not list all the keys. But perhaps we should consider doing this instead. I think only retrieving values without the corresponding keys is rarely useful, so I see three alternatives here:
1. Provide `readAllKeys` only
2. Provide `readAllKeys` and `readAllEntries` which returns `[key, value]` pairs
3. Provide `readAllKeys` and `readAllEntries` and `readAllValues`

Also, we have to think about `readAll` from the current memory sessions. Should we align the name in the adapter for 2.0 (deprecate until then) or rather align the name in the interface to `readAll`?

KnorpelSenf:
@Satont I'm very curious what your opinion is on this

src/convenience/session.ts
roziscoding: > It does not list all the keys. But perhaps we should consider doing this instead. I think only retrieving values without the corresponding keys is rarely useful, so I see three alternatives here:

You're right.
I haven't thought about the implementation of the broadcast plugin enough to be able to tell if we'll need keys, entries, or values, so I'd say we go with 3. to be future proof.

> Should we align the name in the adapter for 2.0 (deprecate until then)

I'm assuming by this you mean we add the three new methods and have `readAll` call `readValues` (current behavior), then remove it on 2.0, right? If so, I think this is the way to go

roziscoding:
> @Satont I'm very curious what your opinion is on this

I'll wait for this before I move on

src/convenience/session.ts
KnorpelSenf: Yes, sounds good then!

Satont:
I'm wondering about implementation of something like that. Already expect to "crutches".

I'm interested about use case such methods, can you explain a little more about what this is for??

roziscoding:
Initially, for the `broadcast` plugin.
The `chat-members` stores information about users in the storage, and the `broadcast` plugin will need to list all those members in order to send messages to them.
`chat-members` need `has` to check wether a member has already been saved, and `broadcast` need `readAll` / `readEntries`

roziscoding:
So, should I go ahead with this?
@KnorpelSenf @Satont

Satont:
> > I'm interested about use case such methods, can you explain a little more about what this is for??
> 
> Initially, for the `chat-members` and the `broadcast` plugins.
> The `chat-members` plugin stores information about users in the storage, and the `broadcast` plugin will need to list all those members in order to send messages to them.
> `chat-members` need `has` to check wether a member has already been saved, and `broadcast` need `readAll` / `readEntries`

Ok. Do you plan implement all those methods in storages?

roziscoding:
> Ok. Do you plan implement all those methods in storages?

I can try to, yes. But the idea of havin the optional properties is so we don't necessarily have to do that.

Satont:
> I can try to, yes. But the idea of havin the optional properties is so we don't necessarily have to do that.

Ok, then go on.

Satont (approved):

src/convenience/session.ts
```diff
@@ -85,6 +85,30 @@ export interface StorageAdapter<T> {
...
+     */
+    readAllEntries?: () =>
+        | Iterable<[key: string, value: T]>
+        | AsyncIterable<[key: string, value: T]>;
+    /**
+     * Lists the values for all the keys that match the RegExp pattern
+     * @param pattern RegExp to match keys against
+     * @returns Values with a matching key
+     */
+    readManyValues?: (pattern: RegExp) => Iterable<T> | AsyncIterable<T>;
```
roziscoding, line 111: @KnorpelSenf what do you think of this? I see it being useful to allow chat-members to read all the members from a given chat

KnorpelSenf, line 111: I*m concerned about this one. Not only that there's no immediate use case for it, but mainly that executing regular expressions on potentially distributed data stores is a hard problem to solve. It seems like the most efficient solution for this would be to enumerate all keys, and then filter them down. However, if the data storage has no special support for it, then we should not implement this logic in every store, but rather do it plugin-side.

Can you elaborate why you added this, or revert if it's not necessary to have it?

KnorpelSenf: Ah I see, ignore the comment below then

KnorpelSenf: I think if we need to enumerate based on structural properties of the key, we should rather explicitly encode this in the string. For example, we could roll with a two-level keyed store (like wide-column) or similar approaches.

KnorpelSenf: Which regex is it?

roziscoding: I'll do it on the plugin's side

KnorpelSenf (changes requested):
LGTM.

Please implement these changes for the built-in memory sessions. Then we can merge.

src/convenience/session.ts
```diff
@@ -407,7 +425,7 @@ function fillDefaults<S>(opts: SessionOptions<S> = {}) {
         debug(
             "Storing session data in memory, all data will be lost when the bot restarts.",
         );
-        storage = new MemorySessionStorage();
+        storage = new MemorySessionStorage<S>();
```
KnorpelSenf, line 428: Is this no longer inferred?

src/convenience/session.ts
```diff
@@ -85,6 +85,24 @@ export interface StorageAdapter<T> {
      * Deletes a value for the given key from the storage.
      */
     delete: (key: string) => MaybePromise<void>;
+    /**
+     * Checks whether a key exists in the storage
```
KnorpelSenf, line 89:
```suggestion
     * Checks whether a key exists in the storage.
```

src/convenience/session.ts
```diff
@@ -85,6 +85,24 @@ export interface StorageAdapter<T> {
...
+     */
+    has?: (key: string) => MaybePromise<boolean>;
+    /**
+     * Lists all keys
```
KnorpelSenf, line 93:
```suggestion
     * Lists all keys.
```

src/convenience/session.ts
```diff
@@ -85,6 +85,24 @@ export interface StorageAdapter<T> {
...
+     */
+    readAllKeys?: () => Iterable<string> | AsyncIterable<string>;
+    /**
+     * Lists all values
```
KnorpelSenf, line 97:
```suggestion
     * Lists all values.
```

src/convenience/session.ts
```diff
@@ -85,6 +85,24 @@ export interface StorageAdapter<T> {
...
+     */
+    readAllValues?: () => Iterable<T> | AsyncIterable<T>;
+    /**
+     * Lists all keys with their values
```
KnorpelSenf, line 101:
```suggestion
     * Lists all keys with their values.
```

KnorpelSenf (changes requested):
One last question. When using the new storage adapter, do we ever need to differentiate between sync and async iterables? If yes, how could this be done?

src/convenience/session.ts
roziscoding: this is what I get on the `storage` variable if I remove the annotation:
image

KnorpelSenf: Fascinating, I wonder why … but this change is obviously good

roziscoding:
> One last question. When using the new storage adapter, do we ever need to differentiate between sync and async iterables?

We shouldn't need to, because sync iterables can be consumed with `await` the same way async ones can.

KnorpelSenf:
Awesome, that's what I was hoping for!

KnorpelSenf (approved):
LGTM, thanks!

PR #375, opened 2023-03-08
fix: incorrect type in `getMyDefaultAdministratorRights`

dcdunkan:
It should be `get...` not `set...` in the `other` parameter.


roj1512 (approved):
LGTM.

KnorpelSenf (approved):
Thank you!

PR #376, opened 2023-03-08
feat: make all adapters available on all platforms

KnorpelSenf:
This PQ makes all frameworks available on all platforms by no longer splitting them up between files.

This has a number of benefits:
- greatly simplifies the framework setup
- no longer makes any assumptions about where adapters exist and where they don't, as many web frameworks actually get ported between the various runtimes
- removes a cyclic dependency between files

@kraftwerk28 can you review this?


dcdunkan (approved):
LGTM.

roj1512 (approved):
Stamp-rubber LGTM.

KnorpelSenf:
@dcdunkan @roj1512 can you share which adapters you've tested?

src/convenience/frameworks.ts
```diff
@@ -0,0 +1,375 @@
...
+                });
+                resolveResponse(res);
+            }
+        },
+        handlerReturn: new Promise((resolve) => {
+            resolveResponse = resolve;
+        }),
+    };
+};
+/** Native CloudFlare workers (service worker) */
```
KnorpelSenf, line 261:
```suggestion

/** Native CloudFlare workers (service worker) */
```

KnorpelSenf:
Did other manual testing, merging.

dcdunkan:
> @dcdunkan @roj1512 can you share which adapters you've tested?

Uhh... none. I just _LGTM_-ed the changes because they looked good to me :D

PR #377, opened 2023-03-09
feat: exponential backoff for retries and permit cancelling `bot.init`

KnorpelSenf:
Makes sure that `bot.init` does not retry too fast.

Allows passing an `AbortSignal` to `bot.init` so the initialization process can be cancelled anytime.

Makes sure that calling `bot.stop` will abort any pending initialization.

Makes sure that polling does not start if the bot is stopped while the `onStart` callback is running.

Closes #291.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -525,13 +525,24 @@ you can circumvent this protection against memory leaks.`);
 /**
  * Performs a network call task, retrying upon known errors until success.
  *
+ * If the task errors and a retry_after value can be used, a subsequent retry
+ * will be delayed by the specified period of time.
+ *
+ * Otherwise, if the first attempt at running the task fails, the task is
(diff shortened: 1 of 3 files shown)


roziscoding (approved):

src/bot.ts
```diff
@@ -525,33 +537,79 @@ you can circumvent this protection against memory leaks.`);
...
-                    if (retryAfter !== undefined) await sleep(retryAfter);
+                    if (retryAfter !== undefined) {
+                        await sleep(retryAfter, signal);
+                        mustDelay = false;
+                    }
                     continue;
                 }
             }
             throw error;
+        } finally {
```
KnorpelSenf, line 578: This should not be entered when the error is rethrown. This needs to be reworked before it can be merged.

KnorpelSenf: Fixed by #379.

PR #379, opened 2023-03-09
fix: do not delay when rethrowing an error

KnorpelSenf:
Fixes https://github.com/grammyjs/grammY/pull/377#discussion_r1130669856.


src/bot.ts
```diff
@@ -556,20 +556,26 @@ async function withRetries<T>(
...
+        let mustDelay = false;
         try {
             result = { ok: true, value: await task() };
-            mustDelay = false;
         } catch (error) {
             debugErr(error);
-            if (error instanceof HttpError) continue;
+            if (error instanceof HttpError) {
+                mustDelay = true;
+                continue;
```
EdJoPaTo, line 566: mustDelay is set to true then the `continue` jumps to the next while loop iteration and directly sets it to false in line 559. Something is fishy here.

EdJoPaTo (changes requested):

src/bot.ts
KnorpelSenf: The `finally` block takes care of this. It's not the prettiest code I've written, but it should work, eh?

KnorpelSenf: Yes, it's not very nice. It makes sense when you think about it, but it does not make sense until you do. Do you approve anyway?

src/bot.ts
```diff
@@ -556,20 +556,26 @@ async function withRetries<T>(
     const INITIAL_DELAY = 100; // ms
     let delay = INITIAL_DELAY;
     while (!result.ok) {
-        let mustDelay = true;
+        let mustDelay = false;
```
EdJoPaTo, line 559: Guess there should be a comment for someone like me in the future… 😬

```suggestion
        let mustDelay = false; // handled in `finally`
```

EdJoPaTo (approved):
Guess it should be a bit better documented with that finally logic but other than that its fine.

src/bot.ts
```diff
@@ -552,36 +552,63 @@ async function withRetries<T>(
...
+            delay = true;
+            strategy = "retry";
+        } else if (error instanceof GrammyError) {
+            if (error.error_code >= 500) {
+                delay = true;
+                strategy = "retry";
+            } else if (error.error_code === 429) {
+                const retryAfter = error.parameters.retry_after;
+                if (retryAfter !== undefined) {
+                    // sleep directly, don't let this affect the backoff
```
EdJoPaTo, line 578: maybe also reset the backoff as its known that the server works as it should?

src/bot.ts
```diff
@@ -552,36 +552,63 @@ async function withRetries<T>(
...
+        if (error instanceof HttpError) {
+            delay = true;
+            strategy = "retry";
+        } else if (error instanceof GrammyError) {
+            if (error.error_code >= 500) {
+                delay = true;
+                strategy = "retry";
+            } else if (error.error_code === 429) {
+                const retryAfter = error.parameters.retry_after;
+                if (retryAfter !== undefined) {
```
EdJoPaTo, line 577: not sure if this is better but this ensures that whatever the API returned is really a number. Its the strict inverse of checking what its not.
```suggestion
                if (typeof retryAfter === 'number') {
```

src/bot.ts
```diff
@@ -552,36 +552,63 @@ async function withRetries<T>(
...
+            }
+        }
+
+        if (delay) {
+            // Do not sleep for the first retry
+            if (lastDelay !== INITIAL_DELAY) {
+                await sleep(lastDelay, signal);
+            }
+            const oneHour = 1 * 60 * 60 * 1000; // ms
+            lastDelay = Math.min(oneHour, 2 * lastDelay);
```
EdJoPaTo, line 593:
```suggestion
            const ONE_HOUR = 1 * 60 * 60 * 1000; // ms
            lastDelay = Math.min(ONE_HOUR, 2 * lastDelay);
```

EdJoPaTo, line 593: one hour sounds like a lot… yeah its broken, yeah its broken for some time now but an hour is a lot in times of constantly available services. Maybe reduce to 20 min? Poking a server every 20 min is still not that often.

EdJoPaTo, line 593: Maybe add a randomised part to that number like ±1%? This prevents that all failing grammY clients with the same logic are hitting roughly at the same time again. If the server is under heavy load or whatever the reason was, its not a good idea to come back all together next time again 👻

EdJoPaTo (commented):
Code is way more readable! I have a few ideas and thoughts to throw in

src/bot.ts
KnorpelSenf: Good idea! I'll add this.

src/bot.ts
KnorpelSenf: I'm using the same interval that Telegram uses when delivering webhooks. But yes, we can reduce it, I don't have too strong opinions here.

KnorpelSenf: I'm not quite sure I understand how this could happen. There's nothing global that syncs up the clients, so the requested should technically already be spread out, no?

EdJoPaTo: Assuming there are 10 bots running hundreds of requests. Then the server fails due to whatever reason. All of the bots start the backdown. As the server failed synchronously for every bot (its the same server) all bots are synced up. At some point the server recovers and is available again. Then it gets hit by all the bots with the same back up delay logic at nearly the same time. Which might cause an overload, kill the server, fill up queues, …

KnorpelSenf: This code is only used during startup. That means you'll have to launch all of these 10 bots exactly at the same time.

EdJoPaTo: ah right, we are at the bot.init. Then yeah, its fine.

EdJoPaTo (approved):

PR #386, opened 2023-03-10
fix: add typesVersions for /types export

roziscoding:
Makes TypeScript aware of where the types for `grammy/types` are actually located.


KnorpelSenf (approved):
Thanks!

PR #387, opened 2023-03-10
refactor: avoid using a symbol for input files

KnorpelSenf:
A common issue is that mismatches between patch versions lead to type errors. This can happen when a plugin imports a version of grammY that is different from the version of grammY that the bot code uses, or if two plugins rely on different versions of grammY.

At runtime, these mismatches do not cause any problems because grammY follows semver, and it also does not rely on any `instanceof` checks that could break due to different class identities. Hence, small version mismatches cause needlessly much headache.

The main contributor to these errors is the symbol that is used in `InputFile`. `InputFile` is a parameter of many API methods. API methods are used in the `Api` class. The `Api` class has an instance which is on the context object at `ctx.api`. The context type is checked in many places, such as when installing plugins. As the symbol used by `InputFile` is different for two different versions of grammY (because that's what symbols are used for), this means that the context objects are not assignable to each other. As a result, a type error is thrown.

This PQ exposes the `toRaw` method on `InputFile` and moves away from a symbol to a regular property. This means that `toRaw` now is part of grammY's public API. It is documented as internal, but it has not changed its shape in quite a bit of time, so it should not be a problem to keep it that way.

### DIFF
--- a/src/types.deno.ts
+++ b/src/types.deno.ts
@@ -1,7 +1,6 @@
 // === Needed imports
 import { basename } from "https://deno.land/std@0.178.0/path/mod.ts";
 import { iterateReader } from "https://deno.land/std@0.178.0/streams/mod.ts";
-import { debug as d, isDeno, toRaw } from "./platform.deno.ts";
 import {
     type ApiMethods as ApiMethodsF,
     type InputMedia as InputMediaF,
@@ -13,6 +12,7 @@ import {
     type InputSticker as InputStickerF,
     type Opts as OptsF,
 } from "https://deno.land/x/grammy_types@v3.0.0/mod.ts";
+import { debug as d, isDeno } from "./platform.deno.ts";
 
 const debug = d("grammy:warn");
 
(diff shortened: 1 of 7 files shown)


KnightNiwrem (approved):

src/types.deno.ts
```diff
@@ -87,7 +87,13 @@ export class InputFile {
...
     }
-    async [toRaw](): Promise<
+    /**
+     * Internal method. Do not use.
+     *
+     * Converts this instance into a binary representation that can be sent to
+     * the Bot API server in the request body.
+     */
+    async toRaw(): Promise<
         Uint8Array | Iterable<Uint8Array> | AsyncIterable<Uint8Array>
```
wojpawlik, line 97: Ideally, there should be a `.stream()` method that always returns `AsyncIterable<Uint8Array>`, to match `Blob`.

wojpawlik (approved):

src/types.deno.ts
KnorpelSenf: I will add this as soon as there's demand, thanks for the suggestion.

PR #388, opened 2023-03-10
fix: guess correct file name when editing media

### LINKED ISSUE #317: bug: `editMessageMedia` incorrect behaviour
I had an issue during update message with animation.

I create message like this

```ts
const file = await Deno.open(LIVE_PREVIEW_FILE, { read: true })
const post = await this.bot.api.sendAnimation(this.channel, new InputFile(file.readable), params)
```

I use mkv file without sound - it work fine, message created with animated media preview.
image

Then in a couple minutes, i trying to update media like this.
```ts
const file = await Deno.open(LIVE_PREVIEW_FILE, { read: true })
await this.bot.api.editMessageMedia(
  this.announce.chat,
  this.announce.id,
  {
    type: 'animation',
    media: new InputFile(file.readable),
    caption: message,
    parse_mode: 'HTML',
  }
 )
```

And got a message without animation preview, but with attached file media.dat
image

I tried to use gif from `new ULR(https://....)`for update and it works.
Is it my fault, telegram, or grammy?

KnorpelSenf:
When guessing a filename, no longer rely on the top-level property that contains the `InputFile` instance. Instead, directly look at the key behind which the `InputFile` instance is stored, however deep that may be. Also, if we are inside `InputMedia*`, use the media type instead. Also also, improve the list of known keys so that we get better at guess them.

Fixes #317.

### DIFF
--- a/src/core/payload.ts
+++ b/src/core/payload.ts
@@ -129,7 +129,15 @@ async function* payloadToMultipartItr(
     yield enc.encode(`\r\n--${boundary}--\r\n`);
 }
 
-type ExtractedFile = { id: string; origin: string; file: InputFile };
+/** Information about a file extracted from a payload */
+type ExtractedFile = {
+    /** To be used in the attach:// string */
+    id: string;
+    /** Hints about where the file came from, useful for filename guessing */
+    origin: string;
+    /** The extracted file */
+    file: InputFile;
+};
 /**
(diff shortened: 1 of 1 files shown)


shevernitskiy (approved):
LGTM!:)

roziscoding (approved):

PR #390, opened 2023-03-11
feat: allow getSessionKey to use extended ctx type

KnightNiwrem:
As of now, we are unable to use narrowed Context type for `getSessionKey` because the parameter type of `ctx` is fixed to `Context` - [TS Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgIQjANHAwhAdjAUwA8M4BnAss4PTAZUurwDEAbAQwDdo4BfOAGZQIIOAHIA5lHYgQATzEBuALAAoNcHwEoA9gGMCcBlRq4AIuxjtEvNTDlhDAWTk4tJOAF5seQh4BkRoymbFzQADzGTOaW7AB8Kqr2jnDMwiBufjDhWHDEhLgAJmQ+7jBxXthwgUhCIgBccAByeE0ArqwcAEasBDkA2mJ1IGIAuhW8iWp6eGTwXWiVuAQA7iho4S6Z+XEAFGJiAJSJM7hzguk44BAUUJULMAB0AsCshFC7u3owRIeN30Q4MASmkRNsSOFkgQIAI4ACKp4KgBCJEA57pY5qNQAemxzQgQKobUocAAFtoCGphldILdHm0KLt2GQ5Lg9HAvj9MMsSIcvBUEGo4MK4bN4AztJU0cNEiLRWdxbcAJKFSoSqCPYCFWUi9grdjAeA8mC7TGqXhmnF4lrwRxQECGwiq8lQSmqakiWnaemMigmPC7QWqOUSAgwKKmADSBDk-x+-LgAAMACQIaXpTWFXiJ9BCkWaQ3AdisRqmhOBi2582HM1AA)

This change attempts to provide `getSessionKey`'s `ctx` parameter with the narrowed `Context` type so that the following can be accomplished.

Screenshot 2023-03-11 at 12 31 47 PM


KnorpelSenf:
I remember trying this but due to some covariance/contravariance problems these changes would break a lot of bots. It is a conscious decision not to have it parameterised.

Perhaps I was wrong, so we could well spend some time on testing this once more. I'm just very careful here.

src/convenience/session.ts
```diff
@@ -131,7 +131,7 @@ export interface SessionOptions<S> {
      * The default implementation will store sessions per chat, as determined by
      * `ctx.chat?.id`.
      */
-    getSessionKey?: (ctx: Context) => MaybePromise<string | undefined>;
+    getSessionKey?: (ctx: C) => MaybePromise<string | undefined>;
```
KnorpelSenf, line 134:
```suggestion
    getSessionKey?: (ctx: Omit<C, "session">) => MaybePromise<string | undefined>;
```

KnorpelSenf (changes requested):
I no longer know what the issue was. Let's ship this and see if this breaks anything. There is a chance that I am confusing this with `initial`, which could take a context object but does not.

@EdJoPaTo it would be helpful to have you test these changes on a few bots of yours.

One thing we certainly have to change is the availability of `ctx.session` inside `getSessionKey`:

src/convenience/session.ts
```diff
@@ -108,7 +108,7 @@ export interface StorageAdapter<T> {
 /**
  * Options for session middleware.
  */
-export interface SessionOptions<S> {
+export interface SessionOptions<S, C extends Context> {
```
KnorpelSenf, line 111:
```suggestion
export interface SessionOptions<S, C extends Context = Context> {
```
or else it will be breaking

KnorpelSenf (changes requested):

KnorpelSenf:
Also, please fixup CI.

KnorpelSenf:
> Also, please fixup CI.

nvm, I did this now

src/convenience/session.ts
KnightNiwrem: Are you able to produce a breaking example? I tried to do only
```ts
const sus = session();
```

But it seems to assume the extending class as the implied default type parameter.

KnightNiwrem: Hmm.. I guess it could be breaking if somebody (for whatever reasons) was trying to do:
```ts
const opts: SessionOptions<SessionFlavor<{}>> = {}
```

Now, they would have to change it to
```ts
const opts: SessionOptions<SessionFlavor<{}>, MyContext> = {}
```

A rather unlikely case, but I guess..

KnorpelSenf (approved):
LGTM, thanks! 🎉

PR #392, opened 2023-03-11
fix: type of createNewStickerSet > stickers is an array

KnorpelSenf:
As reported in https://t.me/grammyjs/118620


Mi3liX9 (approved):
Tried! works well

PR #393, opened 2023-03-11
build: discontinue CDN builds

KnorpelSenf:
They started failing because we had to pin the Deno version to 1.21 and we want to use TS features introduced after that. No one is actually using the CDN, so we might as well stop uploading new versions. The old ones will stay there for a while (until next major).

Instead, people can use https://bundle.deno.dev which appeared in the meantime and does the same thing.

This PQ
- removes the bundling infra
- removes the CI setup
- updates the docs in the README

Note that the web builds at `grammy/web` remain unaffected, as they do not rely on deprecated APIs. In fact, this PQ updates the tooling for them, too.


roj1512 (approved):
LGTM.

PR #399, opened 2023-03-23
feat: narrow down ctx.from inside private chats

KnorpelSenf:
We trust https://t.me/grammyjs/123017


KnorpelSenf:
Approved in https://t.me/grammyjs/123050 already. Waiting for another pair of eyes.

Loskir (approved):

PR #400, opened 2023-03-23
test: add tests for composer.ts

KnorpelSenf:
Tests matter

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,8 +1,11 @@
 {
     "editor.formatOnSave": true,
-    "editor.codeActionsOnSave": { "source.fixAll": true },
+    "editor.codeActionsOnSave": {
+        "source.fixAll": true
+    },
     "deno.enable": true,
     "deno.lint": true,
     "deno.config": "./deno.jsonc",
-    "editor.defaultFormatter": "denoland.vscode-deno"
+    "editor.defaultFormatter": "denoland.vscode-deno",
+    "vue.features.codeActions.enable": false
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,6 +6,5 @@
     "deno.enable": true,
     "deno.lint": true,
     "deno.config": "./deno.jsonc",
-    "editor.defaultFormatter": "denoland.vscode-deno",
-    "vue.features.codeActions.enable": false
+    "editor.defaultFormatter": "denoland.vscode-deno"
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,8 +1,6 @@
 {
     "editor.formatOnSave": true,
-    "editor.codeActionsOnSave": {
-        "source.fixAll": true
-    },
+    "editor.codeActionsOnSave": { "source.fixAll": true },
     "deno.enable": true,
     "deno.lint": true,
     "deno.config": "./deno.jsonc",
--- a/package.json
+++ b/package.json
@@ -26,9 +26,9 @@
     "devDependencies": {
         "@types/debug": "^4.1.7",
         "@types/node": "^12.20.55",
-        "@types/node-fetch": "^2.6.2",
+        "@types/node-fetch": "2.6.2",
         "all-contributors-cli": "^6.24.0",
-        "deno2node": "^1.7.2"
+        "deno2node": "^1.8.1"
     },
     "files": [
         "out/"
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -269,12 +269,7 @@ class ApiClient<R extends RawApi> {
             ? createFormDataPayload(payload, (err) => streamErr.catch(err))
             : createJsonPayload(payload);
         const sig = controller.signal;
-        const options = {
-            ...opts.baseFetchConfig,
-            // deno-lint-ignore no-explicit-any
-            signal: sig as any,
-            ...config,
-        };
+        const options = { ...opts.baseFetchConfig, signal: sig, ...config };
         // Perform fetch call, and handle networking errors
         const successPromise = fetch(
             url instanceof URL ? url.href : url,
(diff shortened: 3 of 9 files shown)


test/composer.test.ts
```diff
@@ -1,20 +1,57 @@
...
 
+describe("BotError", () => {
+    const ctx = new Context(
+        // deno-lint-ignore no-explicit-any
+        { message: { text: "test" } } as any,
+        // deno-lint-ignore no-explicit-any
+        0 as any,
+        // deno-lint-ignore no-explicit-any
+        0 as any,
+    );
```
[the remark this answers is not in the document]
KnorpelSenf: Fair enough, I'll change it

test/deps.test.ts
```diff
@@ -1,7 +1,7 @@
 export {
     readAll,
     readerFromIterable,
-} from "https://deno.land/std@0.178.0/streams/mod.ts";
+} from "https://deno.land/std@0.181.0/streams/mod.ts";
```
dcdunkan, line 4:
```suggestion
} from "https://deno.land/std@0.184.0/streams/mod.ts";
```

in other places too.

dcdunkan (changes requested):

src/platform.deno.ts
```diff
@@ -17,7 +17,7 @@ if (isDeno) {
 
 // === Export system-specific operations
 // Turn an AsyncIterable<Uint8Array> into a stream
-export { readableStreamFromIterable as itrToStream } from "https://deno.land/std@0.178.0/streams/mod.ts";
+export { readableStreamFromIterable as itrToStream } from "https://deno.land/std@0.181.0/streams/mod.ts";
```
dcdunkan, line 20: Here as well? to 0.184.0

dcdunkan (approved):

src/platform.deno.ts
KnorpelSenf: Yes.
```suggestion
export { readableStreamFromIterable as itrToStream } from "https://deno.land/std@0.184.0/streams/mod.ts";
```

KnorpelSenf:
Need to investigate why CI fails.

roziscoding (approved):

.vscode/settings.json
```diff
@@ -1,8 +1,11 @@
...
-    "editor.codeActionsOnSave": { "source.fixAll": true },
+    "editor.codeActionsOnSave": {
+        "source.fixAll": true
+    },
     "deno.enable": true,
     "deno.lint": true,
     "deno.config": "./deno.jsonc",
-    "editor.defaultFormatter": "denoland.vscode-deno"
+    "editor.defaultFormatter": "denoland.vscode-deno",
+    "vue.features.codeActions.enable": false
```
KnorpelSenf, line 10: Please revert

roziscoding: 🤦🏻

src/core/client.ts
```diff
@@ -269,7 +269,12 @@ class ApiClient<R extends RawApi> {
             ? createFormDataPayload(payload, (err) => streamErr.catch(err))
             : createJsonPayload(payload);
         const sig = controller.signal;
-        const options = { ...opts.baseFetchConfig, signal: sig, ...config };
+        const options = {
+            ...opts.baseFetchConfig,
+            // deno-lint-ignore no-explicit-any
```
EdJoPaTo, line 274: Do not use `as any` and ignore the lint to work around a type issue. If it’s a type issue tell that directly to typescript via ts-ignore-error

test/composer.test.ts
```diff
@@ -1,20 +1,51 @@
...
+        assertStringIncludes(e(42), "of type number");
+        assertStringIncludes(e(() => {}), "of type function");
+    });
+});
+
 describe("Composer", () => {
     let composer: Composer<Context>;
-    const ctx = { message: { text: "" } } as Context;
+    const ctx = new Context(
+        // deno-lint-ignore no-explicit-any
```
EdJoPaTo, line 39: Again, use ts-ignore-error over `as any` and a lint comment. (I will omit the other hints to this in this file)

test/composer.test.ts
```diff
@@ -23,48 +54,435 @@ describe("Composer", () => {
 
     it("should call handlers", async () => {
         composer.use(middleware);
-        await composer.middleware()(ctx, next);
+        await exec();
```
EdJoPaTo, line 57: What’s the benefit of hiding this in the `exec` function?

EdJoPaTo (changes requested):

src/core/client.ts
KnorpelSenf: This is not related to this PQ. I can fix it separately at a later point.

test/composer.test.ts
KnorpelSenf: I expected `exec` to be rather complex so I did not want to repeat myself. But it's also kinda opaque, so I'm not happy either way. Suggestions? Inline `exec` everywhere?

src/core/client.ts
EdJoPaTo: This PQ adds the as any so I think it’s done in this one?

KnorpelSenf: It no longer adds this. This is outdated by now.

package.json
```diff
@@ -26,9 +26,9 @@
     "devDependencies": {
         "@types/debug": "^4.1.7",
         "@types/node": "^12.20.55",
-        "@types/node-fetch": "^2.6.2",
+        "@types/node-fetch": "2.6.2",
```
EdJoPaTo, line 29: Why is the requirement changed for the types but not for the dependency itself? Shouldn’t be both or neither changed?

KnorpelSenf: No. The types are broken so we pin them for us until they're fixed. That doesn't mean we cannot permit library users to use newer versions.

test/composer.test.ts
KnorpelSenf: Is that upvote a yes to the inlining?

package.json
```diff
@@ -26,9 +26,9 @@
...
         "all-contributors-cli": "^6.24.0",
-        "deno2node": "^1.7.2"
+        "deno2node": "^1.8.1"
```
dcdunkan, line 31: I guess the version needs to be updated in deno.jsonc backport task as well?

KnorpelSenf: Yes

PR #410, opened 2023-04-25
Webhook Secret Token fix for NextJS

inji-gg:
# Context
Next JS header keys are lower. The fix is simple … just make the header key lowercase as is already done for a few other adapters (fastify, http)

# Test
## Set Webhook
`api/init.ts`
```ts
const bot = new Bot(env.TELEGRAM_BOT_API_TOKEN, { botInfo });

async function handler(request: NextApiRequest, response: NextApiResponse) {
  await bot.api
    .setWebhook(env.TELEGRAM_WEBHOOK_URL, {
      secret_token: env.TELEGRAM_WEBHOOK_SECRET,
    })
    .then((_) => bot.api.getWebhookInfo())
    .then((data) => {
      console.log(data);
      response.status(200).json({
        body: "OK",
      });
    })
    .catch((error) => {
      console.log(error));
      response.status(500).json({ error });
    });
}

export default handler;
```

## Webhook Callback
`api/listen.ts`
```ts
const bot = new Bot(env.TELEGRAM_BOT_API_TOKEN, { botInfo });
bot.on("message:text", async (ctx) => {
  console.log(ctx);
  await ctx.reply("Hello world!");
});

export default webhookCallback(
  bot, 
  "next-js", 
  "throw", 
  8000, 
  env.TELEGRAM_WEBHOOK_SECRET,
);
```

### DIFF
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -206,7 +206,7 @@ const azure: FrameworkAdapter = (context, req) => ({
 /** Next.js Serverless Functions */
 const nextJs: FrameworkAdapter = (req, res) => ({
     update: Promise.resolve(req.body),
-    header: req.headers[SECRET_HEADER],
+    header: req.headers[SECRET_HEADER.toLowerCase()],
     end: () => res.end(),
     respond: (json) => res.status(200).json(json),
     unauthorized: () => res.status(401).send(WRONG_TOKEN_ERROR),


KnorpelSenf (approved):
Thanks!

KnorpelSenf:
@all-contributors add @inji-gg for bug and code

KnorpelSenf:
@inji-gg consider signing your commits in the future. I'll merge anyway now :)

PR #413, opened 2023-05-25
Entities from reply

PonomareVlad:
Useful `entities` method can be expanded to support **messages from reply**

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -408,8 +408,8 @@ export class Context implements RenamedUpdate {
     entities<T extends MessageEntity["type"]>(
         types: MaybeArray<T>,
     ): Array<MessageEntity & { type: T; text: string }>;
-    entities(types?: MaybeArray<MessageEntity["type"]>) {
-        const message = this.msg;
+    entities(types?: MaybeArray<MessageEntity["type"]>, fromReply?: boolean) {
+        const message = fromReply ? this.msg?.reply_to_message : this.msg;
         if (message === undefined) return [];
 
         const text = message.text ?? message.caption;


KnorpelSenf:
This is inconsistent or incomplete. No other context shortcuts take `reply_to_message` into account this way.

However, I do agree that replies are a bit of a gap in what grammY supports. This should be fixed. Here are some examples of what we'd like to do:

- Which entities does the message that is being replied to have?
- Does the message that is being replied to have a `/start` command?
- Is the message that is being replied to an image?
- Does the message that is being replied to have a caption that matches a regular expression?

Let's discuss a few ideas how to tackle this.

1. Take this pull request and apply it to everything else. This would mean that we have to introduce a lot of options everywhere.
2. Provide a namespace `ctx.replyTo` which lets us test for all sorts of things, such as `ctx.replyTo.message` for the message object, `ctx.replyTo.hasCommand("start")` to check for `/start` commands, `ctx.replyTo.entities` to cover this PQ here, and so on.
3. Provide a grammY plugin that does all of these things somehow.
4. Maybe yet another solution?

KnorpelSenf:
@PonomareVlad are you interested in taking this further?

PonomareVlad:
@KnorpelSenf Yes, I’m a little sick, if it’s convenient, I’ll try to participate at the end of this week

KnorpelSenf:
No worries, take your time. Just wasn't sure if you had abandoned this. Get well soon! :)

PonomareVlad:
@KnorpelSenf I agree, thank you !

PonomareVlad:
> Let's discuss a few ideas how to tackle this.

So, as a grammY user, I like options **2** and **3**

In addition to them, I can suggest that you consider adding a new parameter for Filter Queries:

```js
bot.on(":reply_to_message")
```

PonomareVlad:
@KnorpelSenf In any case, I think it's worth migrate this discussion into new Issue, because I won't be able to quickly prepare PR for any of the listed options 🌚

KnorpelSenf:
I can suggest that you consider adding a new parameter for Filter Queries:
> 
> ```js
> bot.on(":reply_to_message")
> ```

We did not add this yet because it implies that things like `:reply_to_message:text` exist, but this is not possible to implement with the current design. The reasoning was that it's better not to do something at all if you can't do it right. This is why we did not want to implement this suggestion in the first place. However, I'm open to discuss this again, as it is more than 2 years ago that we talked about this. If you feel like doing that, ask in the group chat or open an issue to start the discussion again.

> @KnorpelSenf In any case, I think it's worth to migrate this discussion into new Issue, because I won't be able to quickly prepare PR for any of the listed options new_moon_with_face

Alright, check out #429. We can discuss there and open a new pull request once we reached consensus. I will close this until then.

PR #414, opened 2023-05-25
Get file data from message

PonomareVlad:
A separate method for obtaining file data, without calling the `getFile` method

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -1054,17 +1054,21 @@ export class Context implements RenamedUpdate {
      * **Official reference:** https://core.telegram.org/bots/api#getfile
      */
     getFile(signal?: AbortSignal) {
-        const m = orThrow(this.msg, "getFile");
-        const file = m.photo !== undefined
+        const file = this.getFileData();
+        return this.api.getFile(orThrow(file, "getFile").file_id, signal);
+    }
+
+    getFileData() {
+        const m = orThrow(this.msg, "getFileData");
+        return m.photo !== undefined
             ? m.photo[m.photo.length - 1]
             : m.animation ??
-                m.audio ??
-                m.document ??
-                m.video ??
-                m.video_note ??
-                m.voice ??
-                m.sticker;
-        return this.api.getFile(orThrow(file, "getFile").file_id, signal);
+            m.audio ??
+            m.document ??
+            m.video ??
+            m.video_note ??
+            m.voice ??
+            m.sticker;
     }
 
     /** @deprecated Use `banAuthor` instead. */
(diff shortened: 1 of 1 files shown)


KnorpelSenf (changes requested):
What is the point of this? It seems like the only field these objects have in common is the `file_id`. Why not directly provide this value, rather than a big union of unrelated objects that share the `file_id`?

Note that in order to use more fields from this method, one would have to branch out over the returned object again. This would defeat the purpose of the introduced method.

KnorpelSenf:
Do you think it's a good idea to change this to `ctx.fileId` instead?

PonomareVlad:
> What is the point of this?

I used this method to conveniently obtain **MIME** and **file name** that are not available in the [getFile](https://core.telegram.org/bots/api#getfile) method:

> Note: This function may not preserve the original file name and MIME type. You should save the file's MIME type and name (if available) when the File object is received.

That is, the actual file type did not matter to my bot.

> Note that in order to use more fields from this method, one would have to branch out over the returned object again.

I agree, I did not take into account the differences in the formats of the file object, and relied on `ctx.getFileData()?.mime_type` expression.

> Do you think it's a good idea to change this to ctx.fileId instead?

Yes, it sounds like the right solution for most users 🙂

PonomareVlad:
> Do you think it's a good idea to change this to ctx.fileId instead?

@KnorpelSenf I can try to finalize it in the coming days, do I understand correctly that it should be a getter ?

KnorpelSenf:
> > What is the point of this?
> 
> I used this method to conveniently obtain **MIME** and **file name** that are not available in the [getFile](https://core.telegram.org/bots/api#getfile) method:

Ah right, now I see the point, thanks! I agree that it would be nice to support all other fields of files, not just file identifiers. So there several ways to go about this:

1. Introduce `ctx.fileId`, `ctx.fileUniqueId`, `ctx.thumbnail`, `ctx.mimeType`, and so on.
1. Introduce `ctx.file` which contains `ctx.file.id`, `ctx.file.uniqueId`, `ctx.file.thumbnail`, `ctx.file.mimeType`, and so on
1. Only support file identifiers and leave the rest to users
1. Introduce nothing, close this pull request, and leave everything to users.

Each of these fields gets its value from a different set of sources, as e.g. video notes have no mime type, voice notes have no thumbnail, and so on.

I'm going to ask in the [community chat](https://t.me/grammyjs) what people think about this.

KnorpelSenf:
To be honest, there are a lot of things to add to the context object if we want to have true support for files. I don't like options 3 or 4 very much because they are not solving the problem to a satisfying extent.

However, options 1 and 2 add quite a bit of stuff to the context that most people won't need. I think the right place for these changes is the `files` plugin: [grammy.dev/plugins/files](https://grammy.dev/plugins/files.html). We should add all of these things there. Installing the plugin is opt-in, so we can retain a smaller core lib and still have support for a wealth of shortcuts.

I will close this pull request but I would be happy to see a subsequent one against the @grammjys/files repo!

PR #415, opened 2023-05-25
Payload methods export

PonomareVlad:
These methods can be useful when sending _raw_ requests to the Telegram API or third-party services

### DIFF
--- a/src/mod.ts
+++ b/src/mod.ts
@@ -51,4 +51,9 @@ export {
     type Transformer,
     type WebhookReplyEnvelope,
 } from "./core/client.ts";
+export {
+    createJsonPayload,
+    createFormDataPayload,
+    requiresFormDataUpload
+} from "./core/payload.ts";
 export { GrammyError, HttpError } from "./core/error.ts";


KnorpelSenf:
These methods are internal. They have been changed several times in the past, and by exporting them, we'd lock their API. Also, they are tailored towards internal usage by the client, so they don't really provide a good user experience and they cannot be applied very well to different use cases. Those are the reasons why I am not comfortable with exposing these methods.

Raw requests can already be sent to the Telegram Bot API via `api.raw`. grammY does not aim to be a client for third-party services, so this is out of scope.

Can you explain a use case where `api.raw` cannot be used, and where this pull request would solve the problem?

PonomareVlad:
I agree with you, I use these methods in my project, where it is necessary to transfer files by `multipart` requests and their implementation in **grammY** turned out to be more compatible than in **Undici** using `FormData`.

I think it would be correct to close this PR and create a fork of this module in a separate package.

Thank you for your feedback ! 🤗

KnorpelSenf:
I understand, that is a bit amusing to hear. If you use Deno, feel free to access the internal API of the payload module by simply importing from `payload.ts`. I am unsure if Node supports something similar. No backwards compatibility is guaranteed here :)

Naturally, you are right that a fork works well, too. Cheers

PR #417, opened 2023-05-28
docs: fix misspellings

roj1512:
-  Add @mrmaster009 as a contributor
- docs: fix misspellings


KnorpelSenf:
> * Add @mrmaster009 as a contributor

I don't see this anywhere. Do you want to add this?

KnorpelSenf (approved):
Ah you undid this, I assume not then. Otherwise, go ahead with a subsequent PQ.

roj1512:
It was a very old commit in another branch. That guy is in the contributors list now.

PR #419, opened 2023-05-30
Implement new StorageAdapter interface

roziscoding:


src/convenience/session.ts
```diff
@@ -65,6 +65,14 @@ export interface LazySessionFlavor<S> {
     set session(session: MaybePromise<S | null | undefined>);
 }
 
+type CascadingKey = string[];
+type Key = string | CascadingKey;
```
KnorpelSenf, line 69: Perhaps they should be exported? It would then make sense to rename them.

src/convenience/session.ts
```diff
@@ -76,35 +84,52 @@ export interface StorageAdapter<T> {
...
+    values?: (prefix?: Key) => MaybeAsyncIterable<T>;
     /**
-     * Lists all keys with their values.
+     * Lists all key-value pairs via an iterator. Optionally, only lists pairs
+     * with keys that start with a given prefix.
      */
-    readAllEntries?: () =>
-        | Iterable<[key: string, value: T]>
-        | AsyncIterable<[key: string, value: T]>;
+    entries?: (prefix?: Key) => MaybeAsyncIterable<KV<T>>;
```
KnorpelSenf, line 114: These methods need to be reintroduced in order to prevent a breaking change. We can deprecate them in favour of the new methods.

src/convenience/session.ts
```diff
@@ -76,35 +84,52 @@ export interface StorageAdapter<T> {
      * Reads a value for the given key from the storage. May return the value or
      * undefined, or a promise of either.
      */
-    read: (key: string) => MaybePromise<T | undefined>;
+    read: (key: Key) => MaybePromise<T | undefined>;
```
KnorpelSenf, line 87: How is it even possible to do this in a non-breaking way?

src/convenience/session.ts
```diff
@@ -671,34 +787,51 @@ export class MemorySessionStorage<S> implements StorageAdapter<S> {
...
     }
 
-    delete(key: string) {
-        this.storage.delete(key);
+    *keys(prefix?: Key) {
+        for (const key of this.storage.keys(cascade(prefix ?? []))) yield key;
+    }
+    *entries(prefix?: Key) {
+        for (const key of this.keys(prefix)) {
+            const value = this.read(key);
```
KnorpelSenf, line 826: This is ugly and makes me wonder if the Trie should directly support timeouts. This will be useful because it can drop values while enumerating them.

KnorpelSenf:
Closing in favour of https://github.com/grammyjs/chat-members/pull/6

PR #420, opened 2023-05-31
feat: keyboard button builders

### LINKED ISSUE #104: feat: Keyboard button builders and `Keyboard.from`

KnorpelSenf:
Adds keyboard button builders.

- static method to create buttons from labels via `Keyboard.text("label")`
- static method to create keyboards from buttons via `Keyboard.from(array)`
- static method to create keyboards from labels via `Keyboard.from(stringArray)`
- instance method to clone keyboards via `keyboard.clone()`
- instance method to append keyboards via `keyboard.append(other0, other1, other2)`
- instance method to transpose rows and columns via `keyboard.toTransposed()`
- instance method to reflow buttons via `keyboard.toWrapped()`

The same works for inline keyboards.

Closes #104.


Loskir:
Maybe add `prepend` as well?

Loskir:
Maybe add `addRow` as an alias for `row`? So that 

```ts
const keyboard = new Keyboard()
keyboard.addRow(....)
```

sounds more natural

KnorpelSenf:
> 1. I can't see why an option to specify the maximum number of buttons in the first row in `reflow` is particularly useful. Are there any use-cases?

Yes. By default, the last row will be half. By allowing the number of buttons on the first row to be specified, we can make the first row half. For 10 items, we can do `Keyboard.from(array).reflow(3, { first: array.length % 3 })` which will give us

```
[  0  ]
[1 2 3]
[4 5 6]
[7 8 9]
```
rather than

```
[0 1 2]
[3 4 5]
[6 7 8]
[  9  ]
```

> 2. I don't think `reflow` should be an in-place method. It should rather be a static method receiving a 1-d array of buttons. Maybe name it `chunk` as well

I added clone for these cases so one can do `keyboard.clone().reflow()` if in-place is a problem. How about we remove `clone` and always return copies of the current keyboard? I see no benefit of static methods here.

I expected `Keyboard.from(data).reflow()` to be the most common use case, hence in-place made sense.

> 3. I see that it follows a common theme, but it annoys me so much that `transpose` is an in-place method.

Same.

> 4. Is `fromColumns` any useful? Seems like a very specific method too.

It's useful if you want to map an array of strings to a list of buttons that all are places in a single column. I encouter this quite often with `Keyboard.fromColumns(["Welcome", "Settings", "About"])`. Not providing this method either requires an explicit transpose call or a `.map(x => [x])`.

KnorpelSenf:
> Maybe add `prepend` as well?

I see no benefit, just flip the two values?

By the way, should this return a copy, too? In other words, should we create `concat` rather than `append`?

KnorpelSenf:
> Maybe add `addRow` as an alias for `row`?

What is the benefit of redundancy?

Loskir:
Forget about `addRow`, I just realized it accepts `row(button, button, button)` not `row([button, button, button])`

Loskir:
One use case that I'm thinking about rn is conditional buttons, e.g.

```ts
const keyboard = Keyboard.from([[button]])
if (userLoggedIn) {
  keyboard.append([[Keyboard.text('Profile')]])
}
```

`append` seems like a great option for this. But for this to work it should accept not only `Keyboard`s, but also `Button[][]`
Or maybe you have other ideas? We are not limited by existing methods here

KnorpelSenf:
> > > Maybe add `addRow` as an alias for `row`?
> > 
> > 
> > What is the benefit of redundancy?
> 
> just like you added `fromRows` that is the same as `from`.
> 
> The use cases are shifting from chaining `new Keyboard().text().row()` to separate calls
> 
> ```ts
> const keyboard = Keyboard.from(initialData)
> if (something) {
>   keyboard.row(someConditionalData)
> }
> ```
> 
> If this was new code, it would make a lot more sense to call it `addRow` instead of `row`. So why not add it to make such code more readable?

I don't see the pattern that you're referring to. That sounds like you also want to rename it to `keyboard.addText("")` and `keyboard.addRequestLocation("")` and so on?

KnorpelSenf:
> Forget about `addRow`, I just realized it accepts `row(button, button, button)` not `row([button, button, button])`

Oh okay

Loskir:
Great idea! We could make `Keyboard.from(new Keyboard())` a noop and then wrap anything passed in `Keyboard.from`

KnorpelSenf:
Nice! Will do

Loskir:
Speaking of `reflow` and `fromColumn`, how about this?

```ts
const keyboard = Keyboard.chunk([a, b, c], 1) // [[a], [b], [c]]

const keyboard = Keyboard.chunkReverse([a, b, c, d, e], 3) // [[a, b], [c, d, e]]
```

src/convenience/keyboard.ts
```diff
@@ -501,18 +502,19 @@ export class Keyboard {
...
-     * @param buttons A two-dimensional button array
+     * @param source A two-dimensional button array
      * @param options Optional options for the custom keyboard
      */
     static from(
-        buttons: (string | KeyboardButton)[][],
+        source: (string | KeyboardButton)[][] | Keyboard,
         options: Omit<ReplyKeyboardMarkup, "keyboard"> = {},
     ): Keyboard {
+        if (source instanceof Keyboard) return source;
```
Loskir, line 512: well this part is not always correct. what if we wanted to override options from another keyboard?

`const k2 = Keyboard.from(k1, {...})`

KnorpelSenf:
> Speaking of `reflow` and `fromColumn`, how about this?

Those are just array helper methods, right? So they don't really create keyboards

src/convenience/keyboard.ts
KnorpelSenf: Good catch, will fix

Loskir:
> Those are just array helper methods, right? So they don't really create keyboards

Maybe they do the transformation and then pass the result to `Keyboard.from`

KnorpelSenf:
> > Those are just array helper methods, right? So they don't really create keyboards
> 
> Maybe they do the transformation and then pass the result to `Keyboard.from`

What is the purpose of this? Is that any better than directly performing this inside they `from` methods?

I don't want to encourage people to work on button arrays themselves. It should all be possible via `map` and keyboard methods. That's why I'm hesitantant to expose intermediate button arrays.

Loskir:
I agree, you can achieve the same with external functions like `chunk`:

```ts
Keyboard.from(chunk([a, b, c], 2))
```

Does it mean we should remove methods like `fromRows`, `fromColumns` and `reflow`?

Loskir:
I'm not saying we should bundle these methods. I'm saying we should leave this up to users to decide what to use

Loskir:
Let's make a list of use cases that we should provide some utils for

KnorpelSenf:
The three things I had in mind was:
- build rectangular keyboards from linear arrays
- build wrapped keyboards from linear arrays
- build vertical keyboards from linear arrays

This is a bit abstract, but my code covers these things basically. Feel free to send your list here so we can reach agreement on what to do

Loskir:
- add rows/buttons to existing keyboard

Loskir:
> * build rectangular keyboards from linear arrays
> * build wrapped keyboards from linear arrays
> * build vertical keyboards from linear arrays

seems like all we need is an abstraction over `chunk`. question is should it be in-place, static or just an utility

KnorpelSenf:
> seems like all we need is an abstraction over `chunk`

Do you mean this literally (build an abstraction and use `npm:chunk` internally) or figuratively (have a similar implementation)?

KnorpelSenf:
If we were to create instance methods that do not mutate the object, would that be a good abstraction?

Basically, we should change transpose, append, and reflow and have them create new instances.

Loskir:
`append` should remain in-place. 

`reflow` should accept 1-d, it doesn't care about previous structure at all. 

`transpose` is a tricky one. what are the use cases for it?

KnorpelSenf:
> `append` should remain in-place.

We could change it to work similar to Array#concat and rename it to concat. We don't have to, though.

> `reflow` should accept 1-d, it doesn't care about previous structure at all.

What do you mean by 1-d?

> `transpose` is a tricky one. what are the use cases for it?

Personally, I'd mostly use it for turning an array into a column selection, or for a list of pairs into a two-column keyboard.

KnorpelSenf:
How about we do this:

- static things (`from`) create keyboard instances from source data
- non-static things (`reflow`, `transpose`) create keyboard instances from keyboard instances.

No mutations.

Loskir:
> We could change it to work similar to Array#concat and rename it to concat. We don't have to, though.

In my head `append` is mostly useful for conditional rendering as I showed before. Making it return a new keyboard instead of modifying it in-place would require declaring it as let and writing `keyboard = keyboard.concat(somethingElse)` which is not ideal. We could add a static `concat` method for doing `Keyboard.concat(keyboardA, keyboardB)` though.

KnorpelSenf:
I believe append will also be useful for creating a prefix menu which gets a custom footer. But this is probably much more relevant for inline menus. So let's leave append the way it is.

KnorpelSenf:
Have you used numpy? I'm taking inspiration from them, it's a good API. Create an array once, then interact with it using chained method calls in other to process it the way you like.

Loskir:
yeah, i see the similarities now

KnorpelSenf:
So how about this?
- no more reshaping in from: drop fromRows, fromColumns
- no more mutations during reshaping: transpose and reflow use clone internally

Loskir:
Sounds good. We might want to rename them similar to JS's `sort -> toSorted`

KnorpelSenf:
Good point!

Reflow lets us define the number of columns. Should we add a possiblity to reshape based on a given number of rows? I don't think so, but do you?

Loskir:
I don't think so either

KnorpelSenf:
Awesome, then we know what is left to do here now

src/convenience/keyboard.ts
```diff
@@ -257,6 +354,120 @@ export class Keyboard {
...
+     * [a b c]     [  a  ]
+     * [d e f]  ~> [b c d]    (3 colums, 1 on first row)
+     * [g h i]     [e f g]
+     * [  j  ]     [h i j]
+     * ```
+     *
+     * @param columns Maximum number of buttons per row
+     * @param options Optional option for the first row
+     */
+    toReflowed(columns = 4, options = { first: columns }) {
```
Loskir, line 427: i'd change it to be something like `reverseFlow: boolean` instead of `first: number`. current implementation seems to be too specific

src/convenience/keyboard.ts
```diff
@@ -17,16 +19,27 @@ import {
  *   .text('A').text('B').row()
  *   .text('C').text('D')
  *
- * // Now you can either pass it directly:
+ * // Now you send it like so:
```
KnorpelSenf, line 22:
```suggestion
 * // Now you can send it like so:
```

src/convenience/keyboard.ts
```diff
@@ -17,16 +19,27 @@ import {
...
+ * keyboard, you can use the static equivalents which every button has. This
+ * will create a two-dimensional keyboard button array. The resulting array can
+ * be turned into a keyboard instance.
+ *
+ * ```ts
+ * // Data source:
+ * const data = [['a', 'b'], ['c', 'd']]
+ *
+ * // Build a custom keyboard:
+ * const keyboard = Keyboard.from(data.map(row => row.map(Keyboard.text)))
```
KnorpelSenf, line 38:
```suggestion
 * const keyboard = Keyboard.from(data)
```

src/convenience/keyboard.ts
```diff
@@ -257,6 +355,120 @@ export class Keyboard {
...
+     * [a b c]     [  a  ]
+     * [d e f]  ~> [b c d]    (3 colums, 1 on first row)
+     * [g h i]     [e f g]
+     * [  j  ]     [h i j]
+     * ```
+     *
+     * @param columns Maximum number of buttons per row
+     * @param options Optional option for the first row
+     */
+    toReflowed(columns = 4, flow: "bottom" | "top" = "top") {
```
EdJoPaTo, line 428:
```suggestion
    toFlowed(columns = 4, flow: "bottom" | "top" = "top") {
```

The output is a clone of it so nothing is done in-place. 're' should be removed from the name. (Joke: `toFlawed`).

I also don't really like the name flow here as it's not that easy to assume what it does from the name. Yes you can read the documentation but still. If there is a better name it should be used instead.
Personally the method is named [`getButtonsAsRows` in my code](https://github.com/EdJoPaTo/grammy-inline-menu/blob/23a6d0f2424b47c662db050703e933d808955a47/source/buttons/align.ts#L28) which is also not ideal.
Maybe something with align?

EdJoPaTo (changes requested):

test/convenience/keyboard.test.ts
```diff
@@ -63,6 +63,76 @@ describe("Keyboard", () => {
...
+        t([["a", "b"], ["c", "d"], ["e"]], [["a", "c", "e"], ["b", "d"]]);
+        t(
+            [["a", "b"], ["c"], ["d", "e", "f"]],
+            [["a", "c", "d"], ["b", "e"], ["f"]],
+        );
+        const keyboard = Keyboard.from([["a", "b", "c"], ["d", "e"], ["f"]]);
+        assertEquals(keyboard.clone().toTransposed().toTransposed(), keyboard);
+    });
+
+    it("can be reflowed", () => {
```
EdJoPaTo, line 85:
```suggestion
    it("can be wrapped", () => {
```

test/convenience/keyboard.test.ts
```diff
@@ -138,4 +208,121 @@ describe("InlineKeyboard", () => {
...
+        );
+        const keyboard = InlineKeyboard.from([
+            [["a", "a"], ["b", "b"], ["c", "c"]],
+            [["d", "d"], ["e", "e"]],
+            [["f", "f"]],
+        ]);
+        assertEquals(keyboard.clone().toTransposed().toTransposed(), keyboard);
+    });
+
+    it("can be reflowed", () => {
```
EdJoPaTo, line 248:
```suggestion
    it("can be wrapped", () => {
```

EdJoPaTo (dismissed):

src/convenience/keyboard.ts
```diff
@@ -265,8 +495,38 @@ export class Keyboard {
...
+    }
 }
 
+type InlineKeyboardButtonSource =
+    | string
+    | [text: string, data?: string]
+    | InlineKeyboardButton;
+type InlineKeyboardSource =
+    | InlineKeyboardButtonSource[][]
+    | InlineKeyboard;
```
KnorpelSenf, line 528: Export those?

KnorpelSenf:
See https://github.com/EdJoPaTo/wikidata-misfit-bot/pull/89

test/convenience/keyboard.test.ts
```diff
@@ -63,6 +72,83 @@ describe("Keyboard", () => {
...
+        }
+        t([["a"]], [["a"]]);
+        t([["a", "b", "c"]], [["a"], ["b"], ["c"]]);
+        t([["a", "b"], ["c", "d"], ["e"]], [["a", "c", "e"], ["b", "d"]]);
+        t(
+            [["a", "b"], ["c"], ["d", "e", "f"]],
+            [["a", "c", "d"], ["b", "e"], ["f"]],
+        );
+        const keyboard = Keyboard.from([["a", "b", "c"], ["d", "e"], ["f"]]);
+        assertEquals(keyboard.clone().toTransposed().toTransposed(), keyboard);
```
wojpawlik, line 91: Rename the method to `.transpose()`, or make it not mutate.

KnorpelSenf: It does not mutate

KnorpelSenf: We can remove the clone from the test. It's superfluous and confusing.

wojpawlik: Please do! And consider simplifying the name to `.transposed()`.

KnorpelSenf: @Loskir suggested that `toX` is consistent with JS, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted

Will adjust the test.

src/convenience/keyboard.ts
```diff
@@ -184,7 +291,18 @@ export class Keyboard {
      * @param url An HTTPS URL of a Web App to be opened with additional data
      */
     webApp(text: string, url: string) {
-        return this.add({ text, web_app: { url } });
+        return this.add(Keyboard.webApp(text, url));
+    }
+    /**
+     * Adds a new web app button. The Web App that will be launched when the
```
dcdunkan, line 297:
```suggestion
     * Creates a new web app button. The Web App that will be launched when the
```

Like the other static methods' docs

dcdunkan (approved):

src/convenience/keyboard.ts
KnorpelSenf: Whoopsie daisy, thanks a bunch

test/convenience/keyboard.test.ts
```diff
@@ -63,6 +72,83 @@ describe("Keyboard", () => {
...
+        }
+        t([["a"]], [["a"]]);
+        t([["a", "b", "c"]], [["a"], ["b"], ["c"]]);
+        t([["a", "b"], ["c", "d"], ["e"]], [["a", "c", "e"], ["b", "d"]]);
+        t(
+            [["a", "b"], ["c"], ["d", "e", "f"]],
+            [["a", "c", "d"], ["b", "e"], ["f"]],
+        );
+        const keyboard = Keyboard.from([["a", "b", "c"], ["d", "e"], ["f"]]);
+        assertEquals(keyboard.toTransposed().toTransposed(), keyboard);
```
wojpawlik, line 91: This is begging for a property-based test.

src/convenience/keyboard.ts
```diff
@@ -257,6 +375,120 @@ export class Keyboard {
...
+     * [a b c]     [  a  ]
+     * [d e f]  ~> [b c d]    (3 colums, bottom wrapping)
+     * [g h i]     [e f g]
+     * [  j  ]     [h i j]
+     * ```
+     *
+     * @param columns Maximum number of buttons per row
+     * @param wrap Optional wrapping behavior
+     */
+    toWrapped(columns = 4, wrap: "top" | "bottom" = "top") {
```
wojpawlik, line 448: Why the default value?

src/convenience/keyboard.ts
```diff
@@ -257,6 +375,120 @@ export class Keyboard {
...
+        const original = this.keyboard;
+        const wrapped = wrapArray(original, columns, wrap);
+        return this.clone(wrapped);
+    }
+    /**
+     * Creates and returns a deep copy of this keyboard.
+     *
+     * Optionally takes a new grid of buttons to replace the current buttons. If
```
wojpawlik, line 456: Why?

test/convenience/keyboard.test.ts
KnorpelSenf: We don't do those yet. We can consider adding new test setups in future PQs.

deno.jsonc
```diff
@@ -5,36 +5,30 @@
...
-                "./bundling/bundles"
-            ]
-        }
+        "exclude": [
+            "./deno_cache/",
+            "./node_modules/",
+            "./out/",
+            "./package-lock.json",
+            "./bundling/bundles"
+        ]
```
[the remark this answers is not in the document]
KnorpelSenf: Does Deno support global exclude now?

src/convenience/keyboard.ts
EdJoPaTo: when I first saw it I thought its overly complex and `bottom` is probably never needed :D so having `top` the default is better than specifying it every time

KnorpelSenf: Are you suggesting the value is bad or that we shouldn't have a default at all?

src/convenience/keyboard.ts
KnorpelSenf: So that you can copy all config only

src/convenience/keyboard.ts
```diff
@@ -265,8 +497,38 @@ export class Keyboard {
...
+        } else {
+            source = new Keyboard(source.map((row) => row.map(toButton)));
+        }
+        return Object.assign(source, options);
+    }
 }
 
+type InlineKeyboardButtonSource =
+    | string
+    | [text: string, data?: string]
```
Loskir, line 526: i'd prefer we don't introduce another shortcut

src/convenience/keyboard.ts
Loskir: i think that wrap: top/bottom is not intuitive by itself. you can't tell what it does before you look at the examples. it could very well be the opposite behavior

src/convenience/keyboard.ts
KnorpelSenf: Drop the pairs? I'm fine with that, they're a bit of a hack anyway. So this?
- custom
  - string
  - button
- inline
  - string with data=label
  - button

Both as 1D/2D

src/convenience/keyboard.ts
wojpawlik: I think `columns` shouldn't have a default

src/convenience/keyboard.ts
Loskir: i can't see any use for string with data=label but it's fine, at least no one uses them

src/convenience/keyboard.ts
KnorpelSenf: > I think `columns` shouldn't have a default

Fine by me

> top/bottom is not intuitive

Which names are better?

src/convenience/keyboard.ts
KnorpelSenf: Resized etc

src/convenience/keyboard.ts
KnorpelSenf: I've been wanting to drop the default for data for 2.0. But it makes no sense to support it in one place and not support in in the other place. How about we only support buttons here and then indeed remove the default data for 2.0?

deno.jsonc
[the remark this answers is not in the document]
KnorpelSenf: @dcdunkan do you want to create follow-ups for all your PQs? 🫠

src/convenience/keyboard.ts
wojpawlik: `"bottom"` feels useless.

src/convenience/keyboard.ts
EdJoPaTo: having another array isn't that much better than `{text, data}` so… Having it on my own I wouldn't use the type so… guess it should be just removed.

deno.jsonc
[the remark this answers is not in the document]
dcdunkan: > Does Deno support global exclude now?

I don't think it does, does it? I can't see anything about global exclude here: https://deno.com/manual@v1.34.3/getting_started/configuration_file.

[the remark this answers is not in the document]
KnorpelSenf: So it's just @wojpawlik spreading fake news. Tss tss tss.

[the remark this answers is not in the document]
roj1512: To clarify, his washing machine didn’t work thank to our all-in-one natural intelligence-powered fact checker, Dunkan 1.0 Pro.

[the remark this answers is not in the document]
wojpawlik: https://github.com/denoland/deno/pull/17778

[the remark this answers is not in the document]
dcdunkan: Interesting. I wonder why they haven't documented it yet.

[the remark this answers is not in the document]
KnorpelSenf: Nice! I thought it was too good to be true and now it's true?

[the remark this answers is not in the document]
roj1512: Need to add more workforce to the corp for Dunkan 1.2!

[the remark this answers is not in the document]
KnorpelSenf: Can happen in a future PQ.

src/convenience/keyboard.ts
KnorpelSenf: Used here, needed anyway: https://github.com/grammyjs/grammY/blob/e8eb5041f2659d9c4dcf5a7733a71c3b84c83fcc/src/convenience/keyboard.ts#L410

src/convenience/keyboard.ts
KnorpelSenf: Addressed.

src/convenience/keyboard.ts
KnorpelSenf: Addressed all.

deno.jsonc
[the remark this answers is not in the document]
wojpawlik: No reason to delay the change, the file is changed anyway.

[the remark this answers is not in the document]
KnorpelSenf: Sure, if you consider this blocking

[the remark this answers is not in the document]
KnorpelSenf: Turns out, the VSCode extension does not pick this config file format yet. We will have to wait.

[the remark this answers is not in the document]
roj1512: Uhm, yes. It doesn’t even support some fmt options like lineWidth.

dcdunkan (approved):

src/convenience/keyboard.ts
```diff
@@ -1041,13 +1027,17 @@ function transpose<T>(grid: T[][]): T[][] {
     }
     return transposed;
 }
+interface WrapOptions {
+    /** Set to `true` to completely fill up the last row */
+    fillLastRow?: boolean;
```
EdJoPaTo, line 1032: I think i liked the third argument better than another object. You could just see the default value, its way more speaking to use options rather than a bool and I still need to read the documentation (or try it) as `fillLastRow` or even the comment here is just not that helpful in my opinion.

Having the object there is helpful for future improvements so its not a breaking change to add something while keeping it easy to use but still…

KnorpelSenf: What's your suggestion for naming it then? I am not willing to drop this feature for reasons of completeness, but I also can't address the feedback of you guys if you dislike all suggestions without any constructive criticism 🫠
The idea is very easy: make sure to fill up the bottom row with buttons, rather than the top row. Now, how to call it? :)

src/convenience/keyboard.ts
```diff
@@ -284,6 +539,29 @@ export class Keyboard {
...
+ *
+ * If you want to create callback buttons only, you can directly use a
+ * two-dimensional array of strings (or string pairs if label and callback data
+ * are different) and turn it into an inline keyboard.
+ *
+ * ```ts
+ * const keyboard = InlineKeyboard.from([[
+ *     'one',
+ *     'two',
+ *     ['three', 'callback'],
```
Loskir, line 561: outdated

Loskir (approved):
LGTM 🎉

PR #421, opened 2023-05-31
feat: support supplier functions for input files

### LINKED ISSUE #106: feat: InputFile from supplier functions
Passing an iterator to `InputFile` means that the respective API call cannot be repeated. This is because the iterator may already be (partially) consumed, hence corrupting the file data.

Supplier functions could come to the rescue:

```ts
new InputFile(() => createNewIterator())
```
would allow users to pass a function that can return any of the values that are currently supported. This function will be invoked once per API call attempt.

KnorpelSenf:
Not only allows `new InputFile(X)` but also `new InputFile(() => X)`.

This is very useful if we need to fetch the data from some place before it can be sent. Currently, API requests cannot be repeated by the auto-retry plugin if they send response streams or file streams. We are effectively limited to the things that `InputFile` can fetch itself internally. This PQ liberates us from that, so we support arbitrary data sources for retries.

Closes #106.


src/types.node.ts
```diff
@@ -95,11 +98,24 @@ export class InputFile {
...
+    > {
+        return InputFile.toRaw(this);
+    }
+    /**
+     * Internal method. Do not use.
+     *
+     * Converts this instance into a binary representation that can be sent to
+     * the Bot API server in the request body.
+     */
+    static async toRaw(
```
roj1512, line 112: What is the point of this function being static?

wojpawlik (approved):

KnorpelSenf:
I need to add tests

roj1512 (approved):

wojpawlik (approved):

PR #422, opened 2023-05-31
feat: add protection against missing allowed_updates

### LINKED ISSUE #107: feat: Log warning if starting the bot without specifying all update types that are expected based on bot.on()
It happens very often that people forget to register `chat_member` updates.

KnorpelSenf:
People often forget to specify some updates which they listen for. This PQ fixes that by logging a warning in debug mode in such cases.

Closes #107.


dcdunkan (approved):
LGTM

KnorpelSenf:
I never ran this code … I'll test it some more before merging

roj1512 (commented):
Is debug mode mentioned anywhere in the docs?

KnorpelSenf:
Yes, in the getting started guide.

roj1512:
I haven’t seen that for years. Is it anchor-able?

KnorpelSenf:
No, but it is screenshottable:
IMG_20230531_195932_590

roziscoding (approved):

KnorpelSenf:
Works well actually. Will be much more useful if #424 gets merged.

src/bot.ts
```diff
@@ -1,17 +1,35 @@
...
 } from "./core/client.ts";
 import { GrammyError, HttpError } from "./core/error.ts";
+import { type Filter, type FilterQuery, parse, preprocess } from "./filter.ts";
 import { debug as d } from "./platform.deno.ts";
 import { type Update, type UserFromGetMe } from "./types.ts";
 const debug = d("grammy:bot");
+const debugWarn = d("grammy:warn");
 const debugErr = d("grammy:error");
 
+export const DEFAULT_UPDATE_TYPES = [
```
KnightNiwrem, line 17: All update types appear to be available in `filter.ts` as `UPDATE_KEYS` already. Should we prefer to do

```ts
export const ALL_UPDATE_TYPES = Object.keys(UPDATE_KEYS);
export const DEFAULT_UPDATE_TYPES = ALL_UPDATE_TYPES.filter(u => u !== 'chat_member');
```

To keep a single source of truth for update types in grammY's source?

KnorpelSenf: I considered this, but they're semantically unrelated because the supported filter queries need not be the same as the Bot API specs. That's why I didn't follow DRY.

KnightNiwrem: Is it completely semantically unrelated, or perhaps just that it is a superset of Bot API specs? I'd presume you always want to support update types from Bot API specs, but want room to also conjure L1s of your own.

If so, we can probably do TG_UPDATE_KEYS and CUSTOM_UPDATE_KEYS (if and when it is needed); where filter.ts exports the former, and filters on the set union of the above?

KnorpelSenf: > Is it completely semantically unrelated

Yes. I don't want to mix those two things. One is a mirror of the specs, and the other is the base for filter query validation. We don't know if we want to have filter queries for all fields in updates, or if we will have excessive queries that are not found in updates (shortcuts that aren't computed perhaps).

KnightNiwrem (approved):

PR #424, opened 2023-06-01
feat: reset allowed_updates if unspecified

KnorpelSenf:
Should we consider this a breaking change? If yes, we cannot merge it before 2.0. If no, then I am wondering why we did not do this earlier.


KnightNiwrem:
Tangentially related: https://t.me/grammyjs/135849

dcdunkan (approved):

KnightNiwrem (approved):

PR #434, opened 2023-06-14
feat: support response objects in InputFile

### LINKED ISSUE #426: feat: Add support for Response objects in InputFile
Currently the URL is fetched again when a `Response` object is passed.

KnorpelSenf:
Previously, it would not be a type error on Deno to do `new InputFile(response)` because `Response` is assignable to `URLLike`.

However, this would lead to unexpected behaviour because it would make grammY fetch the URL of the Response object again when sending the file, hence ignoring the response object's data.

This PQ fixes that by adding explicit support for Response objects and catching this case.

Close #426.


src/types.deno.ts
```diff
@@ -115,6 +116,11 @@ export class InputFile {
         }
         if (data instanceof Blob) return data.stream();
         if (isDenoFile(data)) return iterateReader(data);
+        // Handle Reponse objects
```
KnorpelSenf, line 119:
```suggestion
        // Handle Response objects
```

dcdunkan (approved):
LGTM

wojpawlik:
`URLLike` was a mistake. It should be removed or changed to `Pick<URL, "href">` in 2.0.

KnorpelSenf:
> `URLLike` was a mistake.

Why is that?

wojpawlik:
Because it forced `Response` support, and `{ href: fileUrl }` still doesn't work: https://github.com/grammyjs/grammY/blob/b78c72ebec625ddf2d23eb27a9eed895710ac3c5/src/types.node.ts#L111-L116

PR #438, opened 2023-06-18
fix: add `"node"` condition to `"."` entry point

### LINKED ISSUE #372: feat: grammy/web builds for all plugins and storage adapters
**Now**

All Storage Adapters (and Session as it is dependent on them) do not support working without Node API. So Grammy can't be used with the sessions plugin on Cloudflare Workers (and similar Edge platforms).

**Expected**

Some/all Storage Adapters can be imported without Node API dependency.

wojpawlik:
Towards #372

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "exports": {
         ".": {
             "types": "./out/mod.d.ts",
-            "default": "./out/mod.js"
+            "node": "./out/mod.js",
+            "default": "./out/web.mjs"
         },
         "./types": {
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
@roziscoding could you run the tests again which you performed for #339?

roziscoding:
> @roziscoding could you run the tests again which you performed for #339?

In a few hours probably :)

KnorpelSenf (approved):
Thanks!

PR #439, opened 2023-06-19
fix reference link for `answerWebAppQuery`

waptik:
Replaced wrong link for `answerWebAppQuery` api


KnorpelSenf (approved):
Whoopsie daisy, thank you so much

PR #445, opened 2023-06-23
refactor: use never record type instead of empty object

Nazar-Ant:
https://github.com/grammyjs/emoji/pull/19#discussion_r1239375175

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -28,7 +28,9 @@ import {
 
 // === Util types
 export type MaybeArray<T> = T | T[];
-export type StringWithSuggestions<S extends string> = (string & Record<never, never>) | S; // permits `string` but gives hints
+export type StringWithSuggestions<S extends string> =
+    | (string & Record<never, never>)
+    | S; // permits `string` but gives hints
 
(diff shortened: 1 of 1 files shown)


KnorpelSenf (approved):
TYSM!

dcdunkan:
Also applicable for https://github.com/grammyjs/conversations/blob/4cf57239b92884eea67841541f9539c4e2418ef4/src/conversation.ts#L34 btw.

Nazar-Ant:
@dcdunkan Should I fix the type there?

dcdunkan:
@Nazar-Ant Why not, I think @KnorpelSenf would appreciate that.

KnorpelSenf:
> @dcdunkan Should I fix the type there?

Yes please!

KnorpelSenf (changes requested):
Also, please fix formatting here

KnorpelSenf (approved):
Thanks a bunch!

KnorpelSenf:
Consider signing your commits in the future. I will overrule now and merge anyway.

PR #454, opened 2023-06-29
perf: perform setup requests in parallel

### LINKED ISSUE #449: perf: parallel requests in setup
Why again do we perform those two API calls in sequence?

KnorpelSenf:
Closes #449.


src/bot.ts
```diff
@@ -392,23 +392,24 @@ a known bot info object.",
      */
     async start(options?: PollingOptions) {
         // Perform setup
+        const setup: Promise<void>[] = [];
```
roj1512, line 395: What about this?

```suggestion
        const setup = [];
```

Or this?

```suggestion
        const setup = new Array<Promise<void>>();
```

src/bot.ts
```diff
@@ -392,23 +392,24 @@ a known bot info object.",
...
         if (!this.isInited()) {
-            await this.init(this.pollingAbortController?.signal);
+            setup.push(this.init(this.pollingAbortController?.signal));
         }
         if (this.pollingRunning) {
+            await Promise.all(setup);
```
roj1512, line 400: Is this line reached with `setup` not being empty?

src/bot.ts
KnorpelSenf: The former does not type-check. The latter has a `new` too many. If anything, we can do
```suggestion
        const setup = Array<Promise<void>>();
```
but I prefer the current syntax because the code in the build output is more intuitive.

src/bot.ts
KnorpelSenf: Yes, if the bot is used in a weird way. The bot info field can be cleared.

roj1512 (approved):

PR #456, opened 2023-06-29
perf: implement filter query caching

### LINKED ISSUE #447: perf: filter query caching
Currently, the filter query logic performs most of its work upfront in order to generate an efficient predicate function which can test updates. However, this is based on the assumption that inline queries are only computed once during setup.

It turns out that some plugins like the menu plugin dynamically create composers upon request, so the filter queries are recreated every time. This is superfluous work, so we should just cache the values.

KnorpelSenf:
```ts
// src/perf.ts
import { matchFilter } from "https://deno.land/x/grammy@v1.17.1/mod.ts";
import { matchFilter as matchFilter2 } from "./mod.ts";

Deno.bench({
    name: "no-cache",
    baseline: true,
    fn: () => {
        matchFilter("::bold");
        matchFilter("message:dice");
        matchFilter("channel_post");
    },
});

Deno.bench({
    name: "cache",
    fn: () => {
        matchFilter2("::bold");
        matchFilter2("message:dice");
        matchFilter2("channel_post");
    },
});

```
gives
```bash
$ deno bench src/perf.ts 
cpu: Intel(R) Core(TM) i5-4200M CPU @ 2.50GHz
runtime: deno 1.34.3 (x86_64-unknown-linux-gnu)

file:///home/steffen/git/grammy/src/perf.ts
benchmark      time (avg)             (min … max)       p75       p99      p995
------------------------------------------------- -----------------------------
no-cache       15.11 µs/iter    (9.97 µs … 10.99 ms)  10.61 µs  39.51 µs  93.39 µs
cache         234.86 ns/iter (181.11 ns … 794.87 ns) 227.01 ns 490.52 ns 497.15 ns

summary
  no-cache
   64.33x slower than cache
```
Closes #447.


roj1512 (approved):
Perfy!

PR #464, opened 2023-07-07
fix: correct case usage for X-Telegram-Bot-Api-Secret-Token header in…

lejovaar7:
This update modifies the handling of the 'X-Telegram-Bot-Api-Secret-Token' header in both the synchronous and asynchronous AWS Lambda functions. 

In the previous implementation, the `SECRET_HEADER_LOWERCASE` constant was being used to retrieve the header from the event. This caused complications since AWS Lambda does not interpret headers in a case-insensitive way, leading to potential mismatches if the header was not supplied in the exact expected case.

With this fix, the `SECRET_HEADER` constant is used, which matches the exact casing of the 'X-Telegram-Bot-Api-Secret-Token' header as it appears in the event. This adjustment should eliminate any issues caused by case sensitivity.

### DIFF
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -181,7 +181,7 @@ const http: FrameworkAdapter = (req, res) => {
 /** AWS lambda serverless functions */
 const awsLambda: FrameworkAdapter = (event, _context, callback) => ({
     update: JSON.parse(event.body),
-    header: event.headers[SECRET_HEADER_LOWERCASE],
+    header: event.headers[SECRET_HEADER],
     end: () => callback(null, { statusCode: 200 }),
     respond: (json) =>
         callback(null, {
@@ -198,7 +198,7 @@ const awsLambdaAsync: FrameworkAdapter = (event, _context) => {
 
     return {
         update: JSON.parse(event.body),
-        header: event.headers[SECRET_HEADER_LOWERCASE],
+        header: event.headers[SECRET_HEADER],
         end: () => resolveResponse({ statusCode: 200 }),
         respond: (json) =>
             resolveResponse({


KnorpelSenf (approved):
LGTM

KnorpelSenf:
@chrisandrewcl how come you used the lowercase version in #418? I didn't try this but it seems to me like the AWS docs imply that headers are not lowercased. Can you enlighten us?

chrisandrewcl:
Expanding on what I said [here](https://github.com/grammyjs/grammY/pull/418#issuecomment-1567660003): while I have seen headers in the canonical form on lambdas before, I've not encountered telegram's secret header in canonical form while working with grammy for my project, just in its lowercase form. Given reports from other developers, it seems it doesn't mean it will always be lowercase, just that in this case it was lowercase. 

Similar reports highlighting this behavior: 
https://github.com/aws/aws-sam-cli/issues/3083#issuecomment-887632702 
https://github.com/aws/aws-sam-cli/issues/1860#issuecomment-888800663

Given the inconsistencies, I was not confident opting for one or another but that was what ultimately was decided for the PR I submitted. 

To avoid this issue, you could consider normalizing the headers, checking for both formats or adding a note to the docs.

I am happy either way, since I ended up using the adapter just as a reference.

KnorpelSenf:
Telegram always delivers headers with proper casing: https://github.com/tdlib/telegram-bot-api/blob/251589221708a8280ffcad32fcdc5348d014a6ae/telegram-bot-api/WebhookActor.cpp#L561

Since it seems like we all agree that AWS doesn't lowercase the headers (side note: why does any software mess with header casing even???), it seems to be the reasonable thing to merge this.

If we find a repro where this breaks anything, we can restart the discussion on the matter.

KnorpelSenf:
CI will be fixed as soon as #465 is merged. The linting rules were updated, so we had to fix the code.

KnorpelSenf:
@lejovaar7 please consider signing your commits in the future. I will overrule now and merge anyway, but we usually require this from contributors :)

KnorpelSenf:
@all-contributors add @lejovaar7 for bug and code

KnorpelSenf:
@all-contributors add @chrisandrewcl for review!

PR #465, opened 2023-07-08
refactor: inline SomeObject type alias

KnorpelSenf:
`object` was now finally allowed by `deno lint` so we can get rid of that ugly helper type alias.


dcdunkan (approved):
LGTM

PR #468, opened 2023-07-09
Fixed use of deprecated express method

PonomareVlad:
Tested by [this](https://github.com/PonomareVlad/grammYExportsTest/blob/express/index.mjs) example 🙂


KnorpelSenf (approved):
Epic stuff

PR #473, opened 2023-08-15
chore: update deps

roj1512:
Also replaces std/testing/asserts imports with std/assert.


KnorpelSenf (approved):
Excellent, thanks. Should've done this before releaseing, will patch.

Issue #275, opened 2022-09-02
Case-insensitive command names


roj1512:
I think there should be an option to make command names case-insensitive, or do it by default.

## Use Case

There are many use cases for this, and here's an example. Imagine we have a bot that configures PC builds, and includes a footer like this in one of its messages:

> See /available_CPUs.

This cannot be handled if the command name passed to the filter doesn't look like the one in the message. And once it's handled the way it is displayed to the user, it can't be handled when the cases are changed, say all lower like `/available_cpus`.

It is case-insensitive by default in most of the other bot frameworks.

KnorpelSenf:
We should ask the community if this is desired by default. I don't think so, but it's best to hear the opinion of many people here.

I would like to keep the command handler in the core package free of configuration options. Custom behaviour for case sensitivity can be added in #108.

PR #478, opened 2023-08-20
fix: case insensitive bot commands

shaunnope:
## Issue
Bot usernames are not case sensitive. Hence, commands such as `/foo@this_bot`, `/foo@this_Bot`, and `/foo@tHis_bOt` would all be sent to a bot with username `@this_bot`.

The existing implementation for bot command checking _is_ case sensitive, causing bot commands with non-standard username capitalisation to not be handled by a grammY bot.

### Example
```javascript
import { Bot } from "grammy";

const bot = new Bot("<telegram token>");

bot.command('foo', ctx => ctx.reply("bar"))

// catch unhandled messages
bot.use(ctx => ctx.chat && ctx.reply("unhandled"))

bot.start();
```

## Proposed Solution
Convert both `atTarget` and `ctx.me.username` to lower case before comparison.

## Notes
I note the intention to configure case insensitivity using the `commands` plugin, as mentioned in #108 and #275. However, I believe this is a separate issue, since bot usernames are inherently case insensitive.


roj1512:
Possibly relatable: https://github.com/grammyjs/grammY/issues/275

KnorpelSenf (approved):
LGTM, thanks!

KnorpelSenf:
@all-contributors add @shaunnope for bug and code!

KnorpelSenf:
Please fix CI

KnorpelSenf:
Thanks. We usually require all contributors to sign their commits. I will ignore that now and merge anyway, but please consider setting this up for future contributions :) (and cuz it's good practice)

shaunnope:
> Thanks. We usually require all contributors to sign their commits. I will ignore that now and merge anyway, but please consider setting this up for future contributions :) (and cuz it's good practice)

Oh thanks! I'll bear that in mind :)

PR #480, opened 2023-08-28
refactor: fix some eslint hints

EdJoPaTo:
I ran [xo](https://github.com/xojs/xo) on this repo and fixed some lints that came up.

### DIFF
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -160,7 +160,7 @@ const http: FrameworkAdapter = (req, res) => {
             const chunks: Chunk[] = [];
             req.on("data", (chunk: Chunk) => chunks.push(chunk))
                 .once("end", () => {
-                    // @ts-ignore `Buffer` is Node-only
+                    // @ts-expect-error `Buffer` is Node-only
                     const raw = Buffer.concat(chunks).toString("utf-8");
                     resolve(JSON.parse(raw));
                 })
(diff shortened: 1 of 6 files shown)


src/convenience/frameworks.ts
```diff
@@ -160,7 +160,7 @@ const http: FrameworkAdapter = (req, res) => {
             const chunks: Chunk[] = [];
             req.on("data", (chunk: Chunk) => chunks.push(chunk))
                 .once("end", () => {
-                    // @ts-ignore `Buffer` is Node-only
+                    // @ts-expect-error `Buffer` is Node-only
```
EdJoPaTo, line 163: backport fails with this line as its no error in this case. Maybe revert this one back to ts-ignore and keep the other ones at ts-expect-error?

src/core/error.ts
```diff
@@ -13,7 +13,7 @@ import { type ApiError, type ResponseParameters } from "../types.ts";
  */
 export class GrammyError extends Error implements ApiError {
     /** Flag that this request was unsuccessful. Always `false`. */
-    public readonly ok: false = false;
+    public readonly ok = false;
```
KnorpelSenf, line 16: This is confusing because one needs to check the type hierarchy in order to understand that type widening is being prevented. Please revert.

KnorpelSenf (changes requested):

src/convenience/frameworks.ts
KnorpelSenf: Yes please

KnorpelSenf (approved):
Thanks! <3

KnorpelSenf:
Remember to sign your commits. I will overrule now and merge anyway.

PR #482, opened 2023-09-09
refactor: reduce Deno deps

### LINKED ISSUE #481: Excess dependencies
In `types.deno.ts` there an import of the entire `path` module which causes unnesessary deps in deno.lock:

```
import { basename } from "https://deno.land/std@0.198.0/path/mod.ts";
```
It would probably be better to import from `https://deno.land/std@0.198.0/path/basename.ts`

KnorpelSenf:
Closes #481.


sergeysolovev (approved):
Looks good to me

KnorpelSenf:
@all-contributors add @sergeysolovev for the idea and the review

PR #484, opened 2023-09-09
build: update deps

KnorpelSenf:
Updates all deps including the Bot API types

### DIFF
--- a/src/platform.web.ts
+++ b/src/platform.web.ts
@@ -3,7 +3,8 @@ export { d as debug };
 
 // === Export system-specific operations
 // Turn an AsyncIterable<Uint8Array> into a stream
-export { readableStreamFromIterable as itrToStream } from "https://deno.land/std@0.198.0/streams/readable_stream_from_iterable.ts";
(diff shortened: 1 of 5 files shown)


dcdunkan (approved):

src/platform.web.ts
```diff
@@ -3,7 +3,8 @@ export { d as debug };
 
 // === Export system-specific operations
 // Turn an AsyncIterable<Uint8Array> into a stream
-export { readableStreamFromIterable as itrToStream } from "https://deno.land/std@0.198.0/streams/readable_stream_from_iterable.ts";
+export const itrToStream = (itr: AsyncIterable<Uint8Array>) =>
+    ReadableStream.from(itr);
```
PonomareVlad, line 7: Since this change, passing blob to `new InputFile()` lead to this error on **Vercel Edge Runtime**:

`TypeError: ReadableStream.from is not a function`

Example code:

```js
const response = await fetch(url, {headers, method, body});

const file = new InputFile(await response.blob(), filename);

await ctx.replyWithDocument(file);
```

Снимок экрана 2023-10-07 в 19 12 07

PR #488, opened 2023-09-25
feat: prevent incorrect signal usage

KnorpelSenf:
In https://t.me/grammyjs/168114 and so many other times, people who do not use TS (and who do not read docs) use the context shortcuts and other methods incorrectly. This PQ attempt to explain to them how to fix their code.

### DIFF
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -426,7 +426,7 @@ function validateSignal(
 ) {
     // We use a very simple heuristic to check for AbortSignal instances
     // in order to avoid doing a runtime-specific version of `instanceof`.
-    if (typeof signal.addEventListener === "function") {
+    if (typeof signal?.addEventListener === "function") {
         return;
     }
 
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -442,7 +442,8 @@ function validateSignal(
         `Incorrect abort signal instance found! \
 You passed two payloads to '${method}' but you should merge \
 the second one containing '${payload1}' into the first one \
-containing '${payload0}'!
+containing '${payload0}'! If you are using context shortcuts, \
+you may want to use a method on 'ctx.api' instead.
 
 If you want to prevent such mistakes in the future, \
 consider using TypeScript. https://www.typescriptlang.org/`,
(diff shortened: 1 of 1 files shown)


KnightNiwrem (approved):

src/core/client.ts
```diff
@@ -244,6 +244,7 @@ class ApiClient<R extends RawApi> {
     ) => {
         const payload = p ?? {};
         debug(`Calling ${method}`);
+        if (signal !== undefined) validateSignal(method, payload, signal);
```
KnightNiwrem, line 247: Null check could be added too

roziscoding (approved):

src/core/client.ts
KnorpelSenf: It would be a type error to pass `null` as a signal, so why not make it a runtime error, too?

KnightNiwrem: Wait no, I highlighted the wrong line

src/core/client.ts
```diff
@@ -417,3 +418,34 @@ function createAbortControllerFromSignal(signal?: AbortSignal) {
...
 }
+
+function validateSignal(
+    method: string,
+    payload: Record<string, unknown>,
+    signal: AbortSignal,
+) {
+    // We use a very simple heuristic to check for AbortSignal instances
+    // in order to avoid doing a runtime-specific version of `instanceof`.
+    if (typeof signal.addEventListener === "function") {
```
KnightNiwrem, line 429: Here, if null is given as signal, it would throw a runtime error when we try to access the property.

Do we want to throw the default error message, or would we prefer to also provide the custom error message below for null signals?

KnorpelSenf: Oh I see. I will adjust this to handle `null`. Good idea, thanks!

src/core/client.ts
```diff
@@ -417,3 +418,34 @@ function createAbortControllerFromSignal(signal?: AbortSignal) {
...
+    }
+    throw new Error(
+        `Incorrect abort signal instance found! \
+You passed two payloads to '${method}' but you should merge \
+the second one containing '${payload1}' into the first one \
+containing '${payload0}'! If you are using context shortcuts, \
+you may want to use a method on 'ctx.api' instead.
+
+If you want to prevent such mistakes in the future, \
+consider using TypeScript. https://www.typescriptlang.org/`,
```
PonomareVlad, line 449: Good try 😀

KnorpelSenf: I hope it will help someone

PR #492, opened 2023-10-07
fix(web): do not rely on ReadableStream.from

### LINKED ISSUE #491: bug: passing blob to `new InputFile()` lead to this error on **Vercel Edge Runtime**:
Since this change, passing blob to `new InputFile()` lead to this error on **Vercel Edge Runtime**:

`TypeError: ReadableStream.from is not a function`

Example code:

```js
const response = await fetch(url, {headers, method, body});

const file = new InputFile(await response.blob(), filename);

await ctx.replyWithDocument(file);
```

Снимок экрана 2023-10-07 в 19 12 07

_Originally posted by @PonomareVlad in https://github.com/grammyjs/grammY/pull/484#discussion_r1349527155_

KnorpelSenf:
Fixes #491

### DIFF
--- a/src/platform.web.ts
+++ b/src/platform.web.ts
@@ -3,8 +3,17 @@ export { d as debug };
 
 // === Export system-specific operations
 // Turn an AsyncIterable<Uint8Array> into a stream
-export const itrToStream = (itr: AsyncIterable<Uint8Array>) =>
-    ReadableStream.from(itr);
+export const itrToStream = (itr: AsyncIterable<Uint8Array>) => {
+    // do not assume ReadableStream.from to exist yet
+    const it = itr[Symbol.asyncIterator]();
+    return new ReadableStream({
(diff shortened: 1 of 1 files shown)


MKRhere (approved):

KnorpelSenf:
Apparently CFW does not support `new ReadableStream`. Why???

KnorpelSenf:
https://developers.cloudflare.com/workers/runtime-apis/streams/readablestream/

roj1512:
> Apparently CFW does not support new ReadableStream. Why???

Because they didn’t know of @KnightNiwrem when they decided to not support it.

PonomareVlad (approved):

KnorpelSenf:
It does support it, but you have to set a flag: https://github.com/grammyjs/grammY/pull/493#issuecomment-1751945395

PR #493, opened 2023-10-07
fix(web): do not rely on ReadableStream.from or new ReadableStream

KnightNiwrem:
For #492


KnightNiwrem:
Please feel free to move `writer.close()` out of the async function, and perhaps into a `.finally(...)` block instead

PonomareVlad (approved):

KnightNiwrem:
CFW should raise the compatibility date in their config, or set the feature flag: https://developers.cloudflare.com/workers/configuration/compatibility-dates/#streams-constructors

PR #494, opened 2023-10-12
support `nhttp` for webhook handler

shevernitskiy:
This PR adds support for [nhttp](https://nhttp.deno.dev/) for webhook handler.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -330,7 +330,7 @@ export class Bot<
     ) {
         if (this.me === undefined) {
             throw new Error(
-"Bot not initialized! Either call `await bot.init()`, \
+                "Bot not initialized! Either call `await bot.init()`, \
 or directly set the `botInfo` option in the `Bot` constructor to specify \
 a known bot info object.",
             );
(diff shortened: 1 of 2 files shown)


shevernitskiy:
> Does this library support returning responses from a `Response` instance?

Handler signature assume the only `RequestEvent` as arg. `Response` object could be returned as well as plenty other types.

roj1512:
We already have a `Request` adapter. Does that suffice?

shevernitskiy:
> We already have a `Request`-`Response` adapter. Does that suffice?

nhttp `RequestEvent` has a source request object, but with consumed body. Also it has `newRequest` - the clone of the original one with no consumed body (this one match for basic Req => Res). But it is a clone - no good for perf and memory, obviously. So thats why I made this PR.

KnorpelSenf (approved):
Thanks!

KnorpelSenf:
@all-contributors add @shevernitskiy for their code

KnorpelSenf:
@shevernitskiy we usually require all contributors to sign their commits. I overruled this now and merged anyway, but please consider signing your commits in the future.

PR #509, opened 2023-12-24
feat: simplify using Test DC

KnorpelSenf:
Requested in https://t.me/grammyjs/190672

Use it like so:

```ts
const bot = new Bot("token", {
    client: { environment: "test" },
});
```

### DIFF
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -116,8 +116,8 @@ export interface ApiClientOptions {
      */
     apiRoot?: string;
     /**
-     * Specifies whether to use the [testing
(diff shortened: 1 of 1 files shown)


Loskir (changes requested):
Maybe rename dc to environment? Data center is a different entity
This update is also not backwards compatible for those who use buildUrl

KnorpelSenf:
> Maybe rename dc to environment?

Hmm yes that's a better name, let me rename it.

> This update is also not backwards compatible for those who use buildUrl

That is not correct, you can always omit the parameter when specifying a function. `() => ""` is assignable to `(a: number, b: number, c: number) => string`.

Loskir (approved):

PR #517, opened 2024-01-01
fix: properly define the subpath ./types in exports

Satont:
Should fix

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './types' is not defined by "exports"
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
         },
         "./types": {
             "types": "./out/types.d.ts",
-            "node": "./out/types.node.js",
             "default": "./out/types.js"
         },
         "./web": {
--- a/package.json
+++ b/package.json
@@ -41,7 +41,9 @@
             "default": "./out/web.mjs"
         },
         "./types": {
-            "types": "./out/types.d.ts"
+            "types": "./out/types.d.ts",
+            "node": "./out/types.node.js",
+            "default": "./out/types.js"
         },
         "./web": {
             "types": "./out/web.d.ts",


package.json
```diff
@@ -41,7 +41,9 @@
             "default": "./out/web.mjs"
         },
         "./types": {
-            "types": "./out/types.d.ts"
+            "types": "./out/types.d.ts",
+            "node": "./out/types.node.js",
+            "default": "./out/types.js"
```
KnorpelSenf, line 46: In the npm package, `types.js` simply re-exports everything from `types.node.js`. Can we perform the following change and still have it work?
```suggestion
            "node": "./out/types.js",
            "default": "./out/types.js"
```

KnorpelSenf: And if yes, then it should also be possible to drop the node-specific export entirely, right?

KnorpelSenf: In fact, @kiki-kanri suggested the same thing in https://github.com/grammyjs/hydrate/issues/24#issuecomment-1873433145

KnorpelSenf:
```suggestion
            "default": "./out/types.js"
```

KnorpelSenf:
This PQ made me remember what is going on. Thanks!

We should only allow type imports via `/types`. This PQ is not going in the right direction.

Instead, the plugin needs to be fixed. It crashes because it uses `export * from` while it should only use `export type * from`. https://github.com/grammyjs/hydrate/pull/25 Fixes that.

wojpawlik (approved):

PR #545, opened 2024-03-22
fix: avoid top-level await

KnorpelSenf:
Dropping the only instance of TLA

### DIFF
--- a/src/platform.deno.ts
+++ b/src/platform.deno.ts
@@ -8,7 +8,7 @@ const DEBUG = "DEBUG";
 if (isDeno) {
     debug.useColors = () => !Deno.noColor;
     const env = { name: "env", variable: DEBUG } as const;
-    const res = await Deno.permissions.query(env);
+    const res = Deno.permissions.querySync(env);
     let namespace: string | undefined = undefined;
     if (res.state === "granted") namespace = Deno.env.get(DEBUG);
     if (namespace) debug.enable(namespace);


KnightNiwrem (commented):
Pending

https://t.me/grammyjs/225486
https://github.com/denoland/deploy_feedback/issues/527

src/platform.deno.ts
```diff
@@ -8,7 +8,7 @@ const DEBUG = "DEBUG";
 if (isDeno) {
     debug.useColors = () => !Deno.noColor;
     const env = { name: "env", variable: DEBUG } as const;
-    const res = await Deno.permissions.query(env);
+    const res = Deno.permissions.querySync(env);
     let namespace: string | undefined = undefined;
     if (res.state === "granted") namespace = Deno.env.get(DEBUG);
```
wojpawlik, line 13:
```suggestion
    const res = Deno.permissions.querySync?.(env);
    let namespace: string | undefined = undefined;
    if (res?.state === "granted") namespace = Deno.env.get(DEBUG);
```

to accomodate Deno Deploy

KnorpelSenf:
This is more complicated than I thought, and the benefits are small. I will abandon this.

KnightNiwrem:
What happened?

KnorpelSenf:
TLA is not supported on all platforms, `querySync` is not supported on all platforms, and if we migrate to JSR, we will have to unify all the platform-specific code anyway. I don't see a lot of value in figuring out a good short-term solution here.

Issue #543, opened 2024-03-19
TypeScript compiles too long when using omitted filter or shortcut for it


SecondThundeR:
> Couldn't find a close example for this in other issues, so I think it would be fine to post this. If this has already been discussed, then this issue can be closed 🤔 

For some reason, tsc taking too long to compile simple example where not full form of filter notation was used, e.g. `.on(":text", ...)`

Steps to reproduce:
1. Head to https://grammy.dev/guide/getting-started#getting-started-on-node-js to create simple project for NodeJS with TypeScript
2. Insert this snippet to bot.ts
	```ts
	import { Bot } from "grammy";
	
	// Create an instance of the `Bot` class and pass your bot token to it.
	const bot = new Bot(""); // <-- put your bot token between the ""
	
	// You can now register listeners on your bot object `bot`.
	// grammY will call the listeners when users send messages to your bot.
	
	// Handle the /start command.
	bot.command("start", (ctx) => ctx.reply("Welcome! Up and running."));
	// Handle other messages.
	bot.on(":text", (ctx) => ctx.reply("Got another message!"));
	
	// Now that you specified how to handle messages, you can start your bot.
	// This will connect to the Telegram servers and wait for messages.
	
	// Start the bot.
	bot.start();
	```
3. Run `tsc` or `tsc --extendedDiagnostics`

**Expected behavior:** TypeScript compiles snippet not longer than regular example or at least not for so long

**Actual behavior:** TypeScript taking 30-35 seconds to compile simple example with omitted filters/shortcut (was getting the same time with `:contact` at my project)

#### Diagnostics for `.on("message:text", ...)`

```shell
Files:                         53
Lines of Library:           36442
Lines of Definitions:       11507
Lines of TypeScript:           19
Lines of JavaScript:            0
Lines of JSON:                  0
Lines of Other:                 0
Identifiers:                50799
Symbols:                    36919
Types:                       4892
Instantiations:             65701
Memory used:               77521K
Assignability cache size:    2630
Identity cache size:           11
Subtype cache size:             0
Strict subtype cache size:      8
I/O Read time:              0.01s
Parse time:                 0.11s
ResolveModule time:         0.01s
ResolveTypeReference time:  0.00s
ResolveLibrary time:        0.00s
Program time:               0.14s
Bind time:                  0.06s
Check time:                 0.12s
transformTime time:         0.00s
commentTime time:           0.00s
I/O Write time:             0.00s
printTime time:             0.01s
[...]

PR #553, opened 2024-04-01
fix: avoid mapped type during filter query type narrowing

### LINKED ISSUE #551: (bug) memory leak
```typescript 
bot.on(":someEvent", ctx => {});
```
After this, a memory leak occurs. This happens at the "context" type calculation stage. At this point, the RAM is loaded by 2 GB. At the same time, the tsc may hang until the end of the calculations, and if there is not enough memory, the program will crash with out of memory error. Until you add `ctx => ...`, everything is fine.

**This only works with events starting with a colon and shortcuts.**

This bug occurs when using the "[Combine With OR](https://grammy.dev/guide/filter-queries#combining-multiple-queries)" method. Therefore, the bug is with `:text`, because `:text` is an alias to `msg:text`, and `msg` is the same as `["message", "channel_post"]`, and this, in turn, is combine.

```typescript 
bot.on("msg"); // still fine
```

```typescript 
bot.on("msg", ctx => {}); // memory leak
```

```typescript 
bot.on("message", ctx => {}); // still fine
```

```typescript 
bot.on(["message", "channel_post"], ctx => {}); // memory leak
```

KnorpelSenf:
The mapped type `AliasProps` prevents us from leveraging the improvements in https://github.com/microsoft/TypeScript/pull/57871. Hence, even though https://github.com/microsoft/TypeScript/issues/57863 was fixed as a consequence of #543, we are still seeing problems like #551 appear using `typescript@next`.

This PQ removes the helper type and inlines its functionality into the existing collection of `Shortcuts`. In addition to fixing #551, this removes complexity and improves consistency. However, it reduces code reuse a little bit, which is acceptable.

You can only test these changes on Node, since Deno does not ship with `typescript@next` yet, which you need to install. You can try out these changes with `npm install github:grammyjs/grammY#avoid-mapped-type` as usual.

Closes #551.


kolay-v (approved):
pq is nice

KnorpelSenf:
@all-contributors add @kolay-v for review

KnorpelSenf:
Oh

KnorpelSenf:
Awesome

test/filter.test.ts
```diff
@@ -112,4 +118,29 @@ describe("matchFilter", () => {
...
+            undefined as any,
+            // deno-lint-ignore no-explicit-any
+            undefined as any,
+        );
+        const pred = matchFilter([":text", "callback_query:data"]);
+        if (pred(ctx)) {
+            if (ctx.callbackQuery) {
+                const s: string = ctx.update.callback_query.data;
+                assert(s);
+                throw "never";
```
roj1512, line 137: Perhaps you could replace this and other instances of it with https://deno.land/std@0.224.0/assert/unreachable.ts.

PR #560, opened 2024-04-10
refactor: prefer FsFile#readable over std/iterateReader

fwqaaq:
1. `iterateReader` was deprecated. See: <https://deno.land/std@0.211.0/streams/iterate_reader.ts?s=iterateReader>.
2. `Deno.open` need to release resources. Deno provided `using` keywords for this purpose. Like this: <https://github.com/denoland/deno/blob/08f46ac4467968f53400c4cf9db95f62b424a161/cli/tsc/dts/lib.deno.ns.d.ts#L1891>
3. `importMaps`, it would be more convenient, if `grammy` and `grammy_type` were to be published JSR. (Like: `deno add`)

### DIFF
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -30,10 +30,5 @@
             "./package-lock.json",
             "./bundling/bundles"
         ]
-    },
-    "imports": {
-        "@std/path": "jsr:@std/path@^0.221.0",
-        "grammy_types": "https://deno.land/x/grammy_types@v3.6.2/mod.ts",
-        "debug": "https://cdn.skypack.dev/debug@4.3.4"
     }
 }
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -7,7 +7,6 @@
         "dev": "deno fmt && deno lint && deno task test && deno task check",
         "coverage": "rm -rf ./test/cov_profile && deno task test --coverage=./test/cov_profile && deno coverage --lcov --output=./coverage.lcov ./test/cov_profile",
         "report": "genhtml ./coverage.lcov --output-directory ./test/coverage/ && echo 'Point your browser to test/coverage/index.html to see the test coverage report.'",
-        "bundle": "cd bundling && ./bundle-all.sh",
         "bundle-web": "mkdir -p out deno_cache && cd bundling && DENO_DIR=$PWD/../deno_cache deno run --unstable --quiet --allow-net --allow-read --allow-env=DENO_DIR,XDG_CACHE_HOME,HOME,DENO_AUTH_TOKENS --allow-write=../out,$PWD/../deno_cache bundle-web.ts dev ../src/mod.ts",
         "contribs": "deno run --allow-env --allow-read --allow-write=. --allow-net=api.github.com --allow-sys npm:all-contributors-cli"
     },
@@ -31,5 +30,10 @@
             "./package-lock.json",
             "./bundling/bundles"
         ]
+    },
+    "imports": {
+        "@std/path": "jsr:@std/path@^0.221.0",
+        "grammy_types": "https://deno.land/x/grammy_types@v3.6.2/mod.ts",
+        "debug": "https://cdn.skypack.dev/debug@4.3.4"
     }
 }
--- a/src/types.deno.ts
+++ b/src/types.deno.ts
@@ -115,7 +115,7 @@ export class InputFile {
                 );
             }
             const file = await Deno.open(data);
-            return file.readable;
+            return file.readable[Symbol.asyncIterator]();
         }
         if (data instanceof Blob) return data.stream();
         if (isDenoFile(data)) return data.readable;
--- a/src/types.deno.ts
+++ b/src/types.deno.ts
@@ -1,6 +1,5 @@
 // === Needed imports
 import { basename } from "https://deno.land/std@0.211.0/path/basename.ts";
-import { iterateReader } from "https://deno.land/std@0.211.0/io/iterate_reader.ts";
 
 import {
     type ApiMethods as ApiMethodsF,
@@ -116,10 +115,10 @@ export class InputFile {
                 );
             }
             using file = await Deno.open(data);
-            return iterateReader(file);
+            return file.readable;
         }
         if (data instanceof Blob) return data.stream();
-        if (isDenoFile(data)) return iterateReader(data);
+        if (isDenoFile(data)) return data.readable;
         // Handle Response objects
         if (data instanceof Response) {
             if (data.body === null) throw new Error(`No response body!`);
--- a/src/types.deno.ts
+++ b/src/types.deno.ts
@@ -146,7 +146,7 @@ async function fetchFile(
     if (body === null) {
         throw new Error(`Download failed, no response body from '${url}'`);
     }
-    return body;
+    return body[Symbol.asyncIterator]();
 }
 function isDenoFile(data: unknown): data is Deno.FsFile {
     return isDeno && data instanceof Deno.FsFile;
--- a/src/types.deno.ts
+++ b/src/types.deno.ts
@@ -1,6 +1,6 @@
 // === Needed imports
-import { basename } from "jsr:@std/path@^0.221.0";
-import { iterateReader } from "jsr:@std/io@^0.221.0";
+import { basename } from "https://deno.land/std@0.211.0/path/basename.ts";
+import { iterateReader } from "https://deno.land/std@0.211.0/io/iterate_reader.ts";
 
 import {
     type ApiMethods as ApiMethodsF,
--- a/src/types.deno.ts
+++ b/src/types.deno.ts
@@ -114,7 +114,7 @@ export class InputFile {
                     "Reading files by path requires a Deno environment",
                 );
             }
-            using file = await Deno.open(data);
+            const file = await Deno.open(data);
             return file.readable;
         }
         if (data instanceof Blob) return data.stream();
--- a/src/types.deno.ts
+++ b/src/types.deno.ts
@@ -118,7 +118,7 @@ export class InputFile {
             return file.readable[Symbol.asyncIterator]();
         }
         if (data instanceof Blob) return data.stream();
-        if (isDenoFile(data)) return data.readable;
+        if (isDenoFile(data)) return data.readable[Symbol.asyncIterator]();
         // Handle Response objects
         if (data instanceof Response) {
             if (data.body === null) throw new Error(`No response body!`);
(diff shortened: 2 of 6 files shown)


roj1512 (changes requested):
`iterateReader()` is not deprecated, it is rather moved from std/streams to std/io.

https://deno.land/std@0.221.0/io/iterate_reader.ts?s=iterateReader

deno.jsonc
```diff
@@ -7,7 +7,6 @@
         "dev": "deno fmt && deno lint && deno task test && deno task check",
         "coverage": "rm -rf ./test/cov_profile && deno task test --coverage=./test/cov_profile && deno coverage --lcov --output=./coverage.lcov ./test/cov_profile",
         "report": "genhtml ./coverage.lcov --output-directory ./test/coverage/ && echo 'Point your browser to test/coverage/index.html to see the test coverage report.'",
-        "bundle": "cd bundling && ./bundle-all.sh",
```
KnorpelSenf, line 10: Why did you remove the bundling?

deno.jsonc
```diff
@@ -31,5 +30,10 @@
             "./package-lock.json",
             "./bundling/bundles"
         ]
+    },
+    "imports": {
+        "@std/path": "jsr:@std/path@^0.221.0",
+        "grammy_types": "https://deno.land/x/grammy_types@v3.6.2/mod.ts",
+        "debug": "https://cdn.skypack.dev/debug@4.3.4"
```
KnorpelSenf, line 37: grammY is a library. You cannot use import maps with libraries. Please revert.

You can technically use them with JSR, but this is a feature of JSR that can rewrite parts of the library as a build step that is based on import maps. The actual library that people later import from JSR does not contain an import map anymore.

fwqaaq: Maybe like this: <https://github.com/oakserver/oak/blob/c8e535fdec9bb8155cdbf92c2ad8172ac1b9e893/examples/closeServer.ts#L6>

deno.jsonc
KnorpelSenf: You're right, I misread, I was thinking about `bundle-web`. This can indeed be removed, thanks!

deno.jsonc
KnorpelSenf: That does not work on /x yet.

src/types.deno.ts
```diff
@@ -1,5 +1,7 @@
 // === Needed imports
-import { basename } from "@std/path";
+import { basename } from "jsr:@std/path@^0.221.0";
+import { iterateReader } from "jsr:@std/io@^0.221.0";
```
roj1512, line 3: I don’t have any problems specifically with using JSR to import std modules, but I think it doesn’t make sense, at least now, since the versions of all modules there are the same.

KnorpelSenf: To be completely honest, I'm also not too comfortable with depending on yet another platform. JSR is very cool and we are going to adopt it at some point, but we currently don't use it anywhere and I would ideally like to avoid a weird transitional period where some parts use JSR and some parts don't and you never really know what's going on. This is just needless complexity. Let's do it once and do it right.

Also, we cannot use `^` because `@std` is still 0.x, so there are constantly breaking changes. We have to pin exact versions anyway.

fwqaaq: Yeah, you're right. I think if we use `JSR`, there's absolutely no need to publish `/x`, as `deno add` can install these packages.

fwqaaq:
```suggestion
import { basename } from "https://deno.land/std@0.211.0/path/basename.ts";
import { iterateReader } from "https://deno.land/std@0.211.0/io/iterate_reader.ts";
```

fwqaaq:
Hey, @roj1512 and @KnorpelSenf, I just tried it out. Using Deno std throudh JSR works perfectly fine, and VSCode Deno plugin also support it.

9AF7D371-08D6-4C98-815A-884C31A3EBE1-82762-0004F95679DE0A5B

src/types.deno.ts
KnorpelSenf: Thanks!

src/types.deno.ts
```diff
@@ -114,11 +114,11 @@ export class InputFile {
                     "Reading files by path requires a Deno environment",
                 );
             }
-            const file = await Deno.open(data);
-            return iterateReader(file);
+            using file = await Deno.open(data);
+            return file.readable;
```
fwqaaq, line 118: I think there's no need to use the standard library; perhaps using ReadableStream would be better.

KnorpelSenf (approved):
Thanks!

KnorpelSenf:
@fwqaaq the tests are failing, something is broken with `using`. Can you fix this up?

We use other deprecated methods in the tests that cover input files. Do you want to adjust the tests in the same pull request, or should this be done at a later point?

test/types.test.ts
```diff
@@ -187,8 +167,9 @@ Deno.test({
         const file = new InputFile({ url: "https://grammy.dev" });
         const data = await file.toRaw();
         if (data instanceof Uint8Array) throw new Error("no itr");
+
         assertRejects(
-            () => readAll(readerFromIterable(data)),
+            () => convertToUint8Array(data),
```
fwqaaq, line 172: Hey, bro, here no exception will be thrown, I don't know what the issue is,  perhaps it's the ReadableStream.from above?

src/types.deno.ts
```diff
@@ -139,12 +139,14 @@ export class InputFile {
...
-async function* fetchFile(url: string | URL): AsyncIterable<Uint8Array> {
+async function fetchFile(
+    url: string | URL,
+): Promise<AsyncIterable<Uint8Array>> {
     const { body } = await fetch(url);
     if (body === null) {
         throw new Error(`Download failed, no response body from '${url}'`);
     }
-    yield* body;
+    return body;
```
fwqaaq, line 149: Rewrite `fetchFile` methods, because `ReadableStream` extends `AsyncIterator` function allowing you to use `yield * ReadableStream` in other methods.

KnorpelSenf: very cool

test/types.test.ts
```diff
@@ -101,34 +104,11 @@ Deno.test({
...
-        const data = await file.toRaw();
-        if (data instanceof Uint8Array) throw new Error("no itr");
-        const values = await readAll(readerFromIterable(data));
-        const expected = new Uint8Array(count);
-        for (let i = 0; i < count; i++) expected[i] = i + 65;
-        assertEquals(values, expected);
-        source.restore();
-    },
-});
-
```
KnorpelSenf, line 131: Why did you remove this test? I understand that the same code may cover this use case, but it would still be nice to be sure that `Deno.FsFile` is supported, even if it's just a check on the type level and the implementation is redundant.

fwqaaq: But the API has been deprecated in Deno in favor of using `Deno.open`.

KnorpelSenf: Fair enough, so we'd have to find a new way to stub it. I don't currently see how to do that, though. I guess I'll just have to accept this one.

src/types.deno.ts
```diff
@@ -114,19 +114,19 @@ export class InputFile {
                     "Reading files by path requires a Deno environment",
                 );
             }
-            const file = await Deno.open(data);
-            return iterateReader(file);
+            using file = await Deno.open(data);
```
KnorpelSenf, line 117: I have one last question about this. My understanding is that `using` causes the file to be closed as soon as the function returns (end of scope), not when the file contents are fully read. How am I wrong about this?

fwqaaq: Your understanding is correct. Using `Deno.open` opens a file and returns a file descriptor, which is automatically closed when the scope ends, not necessarily when the file contents are fully read.

fwqaaq: But returned value is `file.readable` that is a Objejct (like **pointer**), gc will not immediately release these resource.

fwqaaq: Yeah, you're right. But the design of JavaScript sucks ass (The ownership is a complete mess. Objects are passed around carelessly, and no one knows where to release them.). But it does work 🤡:)

KnorpelSenf: Rustacean spotted

fwqaaq: * When I update to the latest version of Deno, writing it like this does cause issues. 
* I will update the PR. :)

src/types.deno.ts
```diff
@@ -114,7 +114,7 @@ export class InputFile {
                     "Reading files by path requires a Deno environment",
                 );
             }
-            using file = await Deno.open(data);
+            const file = await Deno.open(data);
             return file.readable;
```
fwqaaq, line 118: Resource Leak, when I have time, I will try to fix it. (My idea is to use `AsyncIterator` along with `[Symbol.dispose]`)

src/types.deno.ts
```diff
@@ -115,18 +115,18 @@ export class InputFile {
                 );
             }
             const file = await Deno.open(data);
-            return iterateReader(file);
+            return file.readable[Symbol.asyncIterator]();
```
fwqaaq, line 118: Deno has already closed the file when the reader returns `{ done: true }`.

Maybe this is better :)

src/types.deno.ts
fwqaaq, line 149:
```suggestion
    return body[Symbol.asyncIterator]();
```

src/types.deno.ts
```diff
@@ -115,18 +115,18 @@ export class InputFile {
...
         }
         if (data instanceof Blob) return data.stream();
-        if (isDenoFile(data)) return iterateReader(data);
+        if (isDenoFile(data)) return data.readable;
```
fwqaaq, line 121:
```suggestion
        if (isDenoFile(data)) return data.readable[Symbol.asyncIterator]();
```

test/deps.test.ts
```diff
@@ -25,3 +21,15 @@ export {
     type Stub,
     stub,
 } from "https://deno.land/std@0.211.0/testing/mock.ts";
+
+export async function convertToUint8Array(
+    data: Iterable<Uint8Array> | AsyncIterable<Uint8Array>,
+) {
+    const values = [] as number[];
```
KnorpelSenf, line 28: smallest nit of all times
```suggestion
    const values: number[] = [];
```

fwqaaq:
Thanks bro, I remember it.😉

KnorpelSenf:
@all-contributors add @fwqaaq for idea, code

PR #565, opened 2024-05-06
feat: allow `webhookCallback` to infer adapter types

winstxnhdw:
## Background

Traditionally, `webhookCallback` has never been strictly typed. It returns a handler that takes in arguments of `any[]` and returns `any`. This can be confusing to new users, forcing many attempts at different configurations before `webhookCallback` is able to work correctly.

## Objective

This PR will overload `webhookCallback` with typings for all adapters, allowing TypeScript to resolve the appropriate handler parameters and return types. We also add tests to track any incompatibility between our supported web frameworks and the respective adapter implementation.

With the newly resolved types introduced with structural typing, our intellisense can now emit an error when an incorrect configuration is entered.


winstxnhdw:
I am not exactly satisfied with the current implementation. I understand that we want to reduce dependencies, so here's a compromise:

Let's write incomplete type stubs for each adapters. For example, although this `hono` adapter consumes a `ctx`, we can write a contract for only what we need.

```ts
const hono: FrameworkAdapter = (ctx) => {
    let resolveResponse: (res: Response) => void;
    return {
        update: ctx.req.json(),
        header: ctx.req.header(SECRET_HEADER),
        end: () => {
            resolveResponse(ctx.body());
        },
        respond: (json) => {
            resolveResponse(ctx.json(json));
        },
        unauthorized: () => {
            ctx.status(401);
            ctx.statusText(WRONG_TOKEN_ERROR);
            resolveResponse(ctx.body());
        },
        handlerReturn: new Promise<Response>((resolve) => {
            resolveResponse = resolve;
        }),
    };
};
```

In this case, we only need `req`, `req.json`, `req.headers`, `body`, `json`, `status` and `statusText`. So we can write an interface like the following:

```ts
export interface HonoContext {
    req: {
        json: () => Promise<Update>;
        header: (header: string) => string | undefined;
    };
    body: () => Response;
    status: (status: number) => void;
    statusText: (statusText: string) => void;
    json: (json: string) => Response;
}
```

Then our `webhookCallback` can be:

```ts
export function webhookCallback<C extends Context = Context>(
    bot: Bot<C>,
    adapter?: "hono",
    webhookOptions?: WebhookOptions,
): (request: HonoContext) => Promise<Response>;
```

This is a good idea because we use the following properties in our implementation: `req`, `req.json`, `req.headers`, `body`, `json`, `status` and `statusText`. It is not a bad thing to have them typed. If there are any breaking changes on the adapter side, we would have to fix it on our end regardless of whether they are typed or not. 

In the long run, there may be the added benefit of easier maintenance. Do let me know what you think.

winstxnhdw:
Looks like `Deno.requestEvent` can't be backported :(

KnorpelSenf:
That is what I suggested in https://t.me/grammyjs/235629. We could have done this from the inception of the library, but I wanted to move fast back then and couldn't be bothered to figure out the correct types for each adapter. I focused on adding more adapters instead. Since the concept of framework adapters is mature now, we can start adding the types exactly like you suggested, so if you feel like spending time on getting these things done, I'd be more than happy to merge such changes.

The approach outlined above will even work for all adapters, not just the ones with globally available types.

winstxnhdw:
I wonder if we still need this method here. Seems a bit redundant.

```ts
export function webhookCallback<C extends Context = Context>(
    bot: Bot<C>,
    adapter?: SupportedFrameworks | FrameworkAdapter,
    onTimeout?: WebhookOptions["onTimeout"],
    timeoutMilliseconds?: WebhookOptions["timeoutMilliseconds"],
    secretToken?: WebhookOptions["secretToken"],
): (...args: any[]) => any;
```

KnorpelSenf:
It is only there for backward compatibility and will be removed for the next major version. It used to be the only overload.

winstxnhdw:
Maybe we should mark as deprecated?

KnorpelSenf:
Sure, if you want :)

src/convenience/webhook.ts
```diff
@@ -59,11 +81,166 @@ export function webhookCallback<C extends Context = Context>(
...
     secretToken?: WebhookOptions["secretToken"],
 ): (...args: any[]) => any;
+
 export function webhookCallback<C extends Context = Context>(
     bot: Bot<C>,
     adapter?: SupportedFrameworks | FrameworkAdapter,
     webhookOptions?: WebhookOptions,
 ): (...args: any[]) => any;
+
+export function webhookCallback<
```
KnorpelSenf, line 91: Do you think that there is a way to implement this without an overload per adapter? Ideally, I would like to have a union type here that gets narrowed down based on the adapter name or something. That way, we don't have to duplicate the list of adapters to several places.

winstxnhdw: I have been trying to do this for a several hours now but I can't get TypeScript to properly narrow down the return type. It can narrow down the parameters fine.

```ts
export function webhookCallback<C extends Context = Context>(
    bot: Bot<C>,
    adapter: keyof typeof nativeAdapters,
    webhookOptions?: WebhookOptions,
): (
    ...args: Parameters<typeof nativeAdapters[typeof adapter]>
) => ReturnType<
    typeof nativeAdapters[typeof adapter]
>["handlerReturn"];
```

Any ideas?

src/convenience/frameworks.ts
```diff
@@ -231,80 +405,86 @@ const http: FrameworkAdapter = (req, res) => {
...
     },
     respond: (json) => {
-        ctx.response.type = "json";
-        ctx.response.body = json;
+        context.response.type = "json";
+        context.response.body = json;
     },
     unauthorized: () => {
-        ctx.response.status = 401;
+        context.response.status = 401;
```
KnorpelSenf, line 467: Is there a reason for making so many unrelated renames? Looks to me like you're deviating from the naming conventions of the respective frameworks, and from their code examples 🤔

KnorpelSenf: Just to be clear, this isn't blocking, I was just curious why you think that rolling with non-standard names is better

winstxnhdw: I thought it’d be best to stick to single naming convention, but I can revert this.

KnorpelSenf: Ah yeah I think it would be best if every adapter shows idiomatic usage of the respective web framework, you are right that hono did not do this right thus far

winstxnhdw: Done. If I seem to 'miss' some, it's actually because that's actually the naming convention. Only one I am not sure about is Azure.

winstxnhdw:
I’d like to suggest that we have a regression test for every adapter. This way, we can ensure that our adapters are always compatible and still keep the adapters as devDependencies. What do you think?

KnorpelSenf:
Amazing plan. Another thing that I was just too lazy to do so far. 

@PonomareVlad you're gonna like to see this PQ

KnorpelSenf:
The next minor version will be released now (without these changes) but I am happy to follow up with a patch release as soon as this is done :)

winstxnhdw:
Thanks! Will get this done soon. Just need to figure out how we can test TypeScript in Deno by compile-time.

winstxnhdw:
It seems that the implementation for Azure function's callback is not done correctly and I am not sure how to add the types for Express to Deno since they live in `@types/express`

src/convenience/frameworks.ts
```diff
@@ -26,7 +26,7 @@ export type SupportedFrameworks = keyof typeof adapters;
  * Abstraction over a request-response cycle, providing access to the update, as
  * well as a mechanism for responding to the request and to end it.
  */
-export interface ReqResHandler {
+export interface ReqResHandler<T = undefined> {
```
KnorpelSenf, line 29: I'm not sure if
```suggestion
export interface ReqResHandler<T = void> {
```
is possible, could you try?

KnorpelSenf:
> It seems that the implementation for Azure function's callback is not done correctly

That is entirely possible, it isn't very popular so feedback is very rare. 

> I am not sure how to add the types for Express to Deno since they live in `@types/express`

Wouldn't this be done the same way as with all the other adapters, i.e. by defining a minimal interface in the types and do it in a way that is compatible with express? Why would it be a problem that express is on npm?

winstxnhdw:
> Wouldn't this be done the same way as with all the other adapters, i.e. by defining a minimal interface in the types and doing it in a way that is compatible with express? Why would it be a problem that express is on npm?

Because express doesn't come with types. I am not sure how Deno can resolve `@types/express`. Do I just do this?

```ts
import * from 'npm:@types/express'
```

src/convenience/frameworks.ts
winstxnhdw: Seems to work.

KnorpelSenf:
I see. No, you have to add a comment. See https://docs.deno.com/runtime/manual/node/npm_specifiers#typescript-types.

You can also just test it on node.

winstxnhdw:
I have added some Deno permission flags when running the tests. This is because `try-catch` doesn't work on Node modules. Not sure if this will be a security risk. Do check.

winstxnhdw:
I am unsure how to test sveltekit's. As for `worktop`, I think it's best if we wait for Deno compatibility, which is coming very soon.

KnorpelSenf:
I guess I don't understand why the extra permissions are required then

winstxnhdw:
> I guess I don't understand why the extra permissions are required then

Just to not fail the tests.

KnorpelSenf:
@winstxnhdw I fixed up the permission errors, please TAL at the commit message

test/convenience/webhook.test.ts
```diff
@@ -0,0 +1,105 @@
...
+import type { NextApiRequest, NextApiResponse } from "npm:next";
+import { Bot, webhookCallback } from "../../src/mod.ts";
+import { describe, it } from "../deps.test.ts";
+
+import { type Express } from "npm:@types/express@^4.17";
+import type Koa from "npm:@types/koa@^2.15";
+import type { UserFromGetMe } from "../../src/types.ts";
+
+describe("webhook", () => {
+    const bot = new Bot("dummy", { botInfo: {} as unknown as UserFromGetMe });
```
winstxnhdw, line 21: Just wondering why we need to set `botInfo` here.

winstxnhdw:
> @winstxnhdw I fixed up the permission errors, please TAL at the commit message

This is great but I purposefully exposed the parameters for easier maintenance. If there's a breaking change in the future, it will be difficult to identify what's the offending property.

test/convenience/webhook.test.ts
KnorpelSenf: This avoids the call to `getMe` when updates are handled. Usually, grammY initialises the bot as soon as the first update arrives.

KnorpelSenf:
> the parameters

Which ones? I can't quite follow rn

winstxnhdw:
> Which ones? I can't quite follow rn

For example, instead of

```ts
    it("Express should be compatible with grammY adapter", () => {
        const app = { post: () => {} } as unknown as Express;
        app.post("/", webhookCallback(bot, "express"));
    });
```

Maybe it's better to do

```ts
    it("Express should be compatible with grammY adapter", () => {
        const app = { post: () => {} } as unknown as Express;
        app.post("/", (req, res) => { 
          webhookCallback(bot, "express")(req, res) 
        });
    });
```

KnorpelSenf:
Oh. I see. That is not how I would approach debugging that, so I did not even think of that. I mostly inlined it to illustrate idiomatic usage of `webhookCallback` to whoever may read the tests as documentation (and because it is shorter).

But I also don't have too strong opinions about it, so if you feel like the code is more maintainable, we can just as well turn the things into lambdas again. Would you like me to revert?

winstxnhdw:
Oh, how else would you be debugging this? If there's a better way that is not reading the error messages, we can stick with this. Otherwise, I think it's best to turn them back into lambdas again.

test/convenience/webhook.test.ts
winstxnhdw: But in testing code, the first update would never arrive. Is this for when we eventually decide to test runtime code?

KnorpelSenf: The first update does arrive in the tests where we supply the fake updates, such as for NextJS.

src/convenience/frameworks.ts
```diff
@@ -241,8 +410,8 @@ const http: FrameworkAdapter = (req, res) => {
 };
 
 /** koa web framework */
-const koa: FrameworkAdapter = (ctx) => ({
-    update: Promise.resolve(ctx.request.body),
+const koa: KoaAdapter = (ctx) => ({
+    update: Promise.resolve(JSON.parse(ctx.body)),
```
KnorpelSenf, line 414: Are these changes actually correct? It seems to me from the docs that `ctx.body` is a response alias, so obtaining the request body should still be done via `ctx.request.body` after applying a `bodyParser`.

```suggestion
const koa: KoaAdapter = (ctx) => ({
    update: Promise.resolve(ctx.request.body),
```

Is there something I'm missing?

src/convenience/webhook.ts
```diff
@@ -21,6 +21,7 @@ const callbackAdapter: FrameworkAdapter = (
     respond: callback,
     header,
     unauthorized,
+    handlerReturn: undefined,
```
KnorpelSenf, line 24: No longer needed.
```suggestion
```

src/convenience/webhook.ts
```diff
@@ -54,16 +55,28 @@ export interface WebhookOptions {
  */
 export function webhookCallback<C extends Context = Context>(
     bot: Bot<C>,
-    adapter?: SupportedFrameworks | FrameworkAdapter,
```
KnorpelSenf, line 57: Perhaps we should remove the definition of `SupportedFrameworks` entirely?

KnorpelSenf (changes requested):

src/convenience/webhook.ts
winstxnhdw: Yeap.

src/convenience/frameworks.ts
winstxnhdw: There doesn't seem to be a body property in the types.

winstxnhdw: Are you referring to using the `bodyparser` plugin?  https://github.com/koajs/bodyparser

KnorpelSenf: Yes

KnorpelSenf:
@winstxnhdw would you like to perform https://t.me/grammyjs/237669 in the same PQ, or rather get this merged and follow up with more changes?

winstxnhdw:
It's a bit out of the scope of this PR. I think that should be a in another PR.

KnorpelSenf:
Alright, agreed. Regarding https://github.com/grammyjs/grammY/pull/565/commits/aae3884ad95b4fa1938cb0cd31b4110959a8444c, I am no longer comfortable with option 1. We may do option 2 or also just leave it and merge this as-is. What do you prefer?

winstxnhdw:
I want to try and keep our PRs cleaner and with a single objective. Let’s leave that for the next PR. Also, I am not exactly sure how we want to test the runtime. We should discuss this more in the chats.

KnorpelSenf:
This PQ already introduces runtime tests for some of the adapters, but none of the others. But we've iterated long enough on this one, I'm more than happy to merge.

KnorpelSenf (approved):

KnorpelSenf:
Please consider signing your commits in the future :)

KnorpelSenf:
@all-contributors add @winstxnhdw for idea and code and review and test

PR #586, opened 2024-05-24
chore(ci): type-check library in CI

KnorpelSenf:
How did we never type-check grammY in CI? It currently fails and I only discovered this locally.

### DIFF
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -59,7 +59,7 @@ jobs:
                   deno-version: v1.x
 
             - name: Cache Dependencies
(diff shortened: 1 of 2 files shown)


winstxnhdw:
Do we have to add `/// <reference types="npm:@types/node" />` in here as well?

KnorpelSenf:
I assumed the same thing. It might only check test files, not all modules under test.

I don't think we can use triple-slash directives here because that would pollute the global types at runtime. We have to find a different way.

winstxnhdw:
Ok, let me handle this.

KnightNiwrem (approved):

PR #594, opened 2024-06-02
feat: allow cloning `Api` instances

KnorpelSenf:
This exposes `api.token` and `api.options`. This is useful because it lets us create new `Api` instances from existing ones, optionally reconfiguring them on the fly.

The main use case for this is the second major version of the conversations plugin which no longer shares the same `Api` instance across context objects—something that was a big hack that mostly worked for v1.

### DIFF
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -97,17 +97,17 @@ export class Api<R extends RawApi = RawApi> {
      * if transformers.
      *
      * @param token Bot API token obtained from [@BotFather](https://t.me/BotFather)
-     * @param client Optional API client options for the underlying client instance
+     * @param options Optional API client options for the underlying client instance
      * @param webhookReplyEnvelope Optional envelope to handle webhook replies
      */
     constructor(
         public readonly token: string,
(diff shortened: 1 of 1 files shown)


src/core/api.ts
```diff
@@ -91,14 +91,23 @@ export class Api<R extends RawApi = RawApi> {
...
+     *
+     * @param token Bot API token obtained from [@BotFather](https://t.me/BotFather)
+     * @param options Optional API client options for the underlying client instance
+     * @param webhookReplyEnvelope Optional envelope to handle webhook replies
+     */
     constructor(
-        token: string,
-        config?: ApiClientOptions,
+        public readonly token: string,
+        public readonly options?: ApiClientOptions,
```
KnorpelSenf, line 105: Wer should probably do
```suggestion
        public readonly options?: Readonly<ApiClientOptions>,
```

KnightNiwrem (approved):

src/core/api.ts
KnorpelSenf: I'm naïve enough to believe that this isn't needed

PR #603, opened 2024-06-21
fix: resolve type conflict of ctx.chatId for callback queries

KnorpelSenf:
First reported in https://t.me/grammyjs/248339


KnorpelSenf:
Approved in https://t.me/grammyjs/248847

PR #604, opened 2024-06-21
fix: unify webhookCallback overloads

KnorpelSenf:
First reported in https://t.me/grammyjs/248742


KnorpelSenf:
Approved in https://t.me/grammyjs/248849

PR #610, opened 2024-07-03
feat: suggest "privacy" commands

KnorpelSenf:
In response to https://t.me/BotNews/96, the right thing to do is to suggest developers to add a section about privacy.

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -11,7 +11,7 @@ import {
     type PreCheckoutQueryContext,
     type ReactionContext,
     type ShippingQueryContext,
-    type StringWithSuggestions,
+    type StringWithCommandSuggestions,
 } from "./context.ts";
 import { type Filter, type FilterQuery } from "./filter.ts";
 import {
@@ -380,10 +380,8 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
      * @param command The command to look for
      * @param middleware The middleware to register
      */
-    command<S extends string>(
-        command: MaybeArray<
-            StringWithSuggestions<S | "start" | "help" | "settings">
-        >,
+    command(
+        command: MaybeArray<StringWithCommandSuggestions>,
         ...middleware: Array<CommandMiddleware<C>>
     ): Composer<CommandContext<C>> {
         return this.filter(Context.has.command(command), ...middleware);
--- a/src/context.ts
+++ b/src/context.ts
@@ -32,9 +32,13 @@ import {
 
 // === Util types
 export type MaybeArray<T> = T | T[];
-export type StringWithSuggestions<S extends string> =
+/** permits `string` but gives hints */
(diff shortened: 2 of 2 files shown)


HeeroML:
I actually disagree to this. But it's not a big change. In the docs it's the right place for it. I get where you come from with this pr, but we actually don't know enough to make it such a recommendation, not every country is required to have such a policy in place or they have different regulations.

roziscoding:
> I actually disagree to this. But it's not a big change. In the docs is the right place for it. I get where you come from with this pr, but we actually don't know enough to make it such a recommendation, not every country is required to have such a policy in place or they have different regulations. 

Tha announcement says every bot should respond to the command. It is a "standardized" command as much as `/help` or `/start` IMO

KnorpelSenf:
That is incorrect. According to the GDPR, every service that is accessible by citizens of an EU country must comply with it. Every bot is available in the EU, which means that every bot must comply.

Legally, there are only two types of bots: those with a privacy notice and those who break (at least) EU law. In practice, this means that nearly every new bot will implement a privacy notice.

Whether or not we should add an auto-complete suggestion for this is a different question, but my main motivation for adding this command is actually not a legal one—it is that BotNews encourages every bot developer to implement `/privacy`. Similarly, native apps have special handling for `/start`, `/help`, and `/settings` in their UI. Those commands all are a convention defined by Telegram, and grammY should always stay close to how bots are meant to be written.

HeeroML:
> Tha announcement says every bot should respond to the command. It is a "standardized" command as much as `/help` or `/start` IMO

Your right, i read it multiple times and misread the "must make it"

HeeroML (approved):

KnorpelSenf:
Thanks for being thorough! I appreciate it :)

HeeroML:
> > I actually disagree to this. But it's not a big change. In the docs is the right place for it. I get where you come from with this pr, but we actually don't know enough to make it such a recommendation, not every country is required to have such a policy in place or they have different regulations.
> 
> 
> 
> That is incorrect. According to the GDPR, every service that is accessible by citizens of an EU country must comply with it.



This is actually incorrect too. It needs to target EU Users, websites, speaking chinese and so on but not being blocked in EU, don't need to comply. Or i got something wrong seit GDPR. It's mainly the intent to process EU-Data.

KnorpelSenf:
https://gdpr.eu/companies-outside-of-europe/ seems to explain it fairly well.

You might be referring to targeting EU consumers, and I agree that this part does not apply. However, there is a second part: monitoring. If your bot is recording data about EU citizens (storing any kind of data in your database about them) then GDPR applies.

There seems to be an exception, though. If your bot is neither professional nor commercial, then this could be an exception. At this point I'm not entirely sure anymore.

Adding a short privacy note is trivial, though, so it's probably best to just do it.

PR #622, opened 2024-08-15
feat: suggest /developer_info as a command

EdJoPaTo:
This is suggested in https://telegram.org/privacy-tpa so maybe embrace this a bit more?

See also #610


KnorpelSenf (approved):
Thanks!

vggj21 (approved):

PR #623, opened 2024-08-15
build: move exclude to deno.jsonc top level

EdJoPaTo:
Maybe there is some meaning behind this why it isn't used yet. Backwards compatibility? But the two entries are not equal so moving them to a single place provides a single point of truth there that can't diverge.


KnorpelSenf (approved):
Thank you. This feature did not exist yet when the config file was written.

PR #625, opened 2024-08-16
feat: support Bot API 7.9

KnorpelSenf:
https://core.telegram.org/bots/api#august-14-2024

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -231,10 +231,21 @@ const checker: StaticHas = {
             : (Array.isArray(reaction) ? reaction : [reaction]).map((emoji) =>
                 typeof emoji === "string" ? { type: "emoji", emoji } : emoji
             );
+        const emoji = new Set(
+            normalized.filter((r) => r.type === "emoji")
+                .map((r) => r.emoji),
+        );
+        const customEmoji = new Set(
+            normalized.filter((r) => r.type === "custom_emoji")
+                .map((r) => r.custom_emoji_id),
+        );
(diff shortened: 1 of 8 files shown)


src/context.ts
```diff
@@ -700,6 +711,21 @@ export class Context implements RenamedUpdate {
...
+         * reaction, and `false` otherwise
+         */
+        paid: boolean;
+        /**
+         * `true` if a paid reaction was newly added to this user's reaction,
+         * and `false` otherwise
+         */
+        paidAdded: boolean;
+        /**
+         * `true` if a paid reaction was removed from this user's reaction, and
```
rojvv, line 725: Feel free to ignore.

```suggestion
         * `true` if a paid reaction was just added to this user's reaction,
         * and `false` otherwise
         */
        paidAdded: boolean;
        /**
         * `true` if a paid reaction was just removed from this user's reaction, and
```

rojvv (approved):

src/context.ts
KnorpelSenf: I will. It is inconsistent with the wording above for emoji reactions and custom emoji reactions.

src/context.ts
```diff
@@ -700,6 +711,21 @@ export class Context implements RenamedUpdate {
...
+         * `false` otherwise
+         */
+        paidRemoved: boolean;
```
KnorpelSenf, line 728: Are users actually able to remove star reactions? @rojvv do you know?

KnorpelSenf: The cannot be removed

PR #626, opened 2024-08-21
build: update deno2node, fix docs, add suggestions to sendDice

rayz1065:
The JSDocs for `api.sendDice` and `ctx.replyWithDice` were missing '🎳'  so I brought them up to date with the latest changes on https://core.telegram.org/bots/api#senddice.
I also added suggestions for the parameter `emoji` following the style in `sendChatAction`. It might be reasonable to remove `| (string & Record<never, never>)` but I didn't want to introduce a breaking change.

I also fixed two JSDocs with an incorrect link to the documentation.

### DIFF
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -2,7 +2,7 @@
     "lock": false,
     "tasks": {
         "check": "deno cache --check=all src/mod.ts",
-        "backport": "deno run --no-prompt --allow-read=. --allow-write=. https://deno.land/x/deno2node@v1.9.0/src/cli.ts tsconfig.json",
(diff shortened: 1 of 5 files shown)


KnorpelSenf:
Can you fix the formatting?

KnorpelSenf:
Hang on, it's actually not related to these changes, I'll investigate

KnorpelSenf:
I guess we can just disable the lint until we support `node:http` as an adapter on Deno

KnorpelSenf (approved):
I fixed it up. Thanks!

KnorpelSenf:
@rayz1065 please consider signing your commits in the future. We usually require this from all contributions, but I made an exception now and merged anyway.

@all-contributors add @rayz1065 for the idea and the code!

PR #644, opened 2024-10-16
feat: `bot.isRunning()`

KnorpelSenf:
Adds a way to inspect whether or not the bot is currently running.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -85,6 +85,9 @@ export interface PollingOptions {
      * fetched. The bot information `bot.botInfo` will be available when the
      * function is run. For convenience, the callback function receives the
      * value of `bot.botInfo` as an argument.
+     *
+     * When this function is invoked, the bot already signals that it is
+     * running. In other words, `bot.isRunning()` already returns true.
      */
     onStart?: (botInfo: UserFromGetMe) => void | Promise<void>;
 }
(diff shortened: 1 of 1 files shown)


src/bot.ts
```diff
@@ -495,6 +498,25 @@ a known bot info object.",
...
+     * and false otherwise.
+     *
+     * If this method returns true, it means that `bot.start()` has been called,
+     * and that the bot has neither crashed nor was it stopped via a call to
+     * `bot.stop()`. This also means that you cannot use this method to check if
+     * a webhook server is running, or if grammY runner was started.
+     *
+     * Note that this method will already begin to return true even before the
+     * call to `bot.start()` has completed its initialization phase. During that
+     * period, it is `bot.isRunning() && !bot.isInited()`. By extension, this
```
roziscoding, line 512: > During that period, it is `bot.isRunning() && !bot.isInited()`

I don't get what this is trying to tell me. Does it mean the function returns something else? Or that I should do something else. I think could be a bit more clear here 🤔

KnorpelSenf: It just means that just because the bot is running, does not mean it is inited. I will clarify this.

KnightNiwrem (approved):

KnightNiwrem:
Interesting... so this never worked before:

```ts
// Some transformer to always throw 401 error on getUpdates

try {
  await bot.start();
} catch {}

console.log('abc');
console.log(bot.pollingRunning);

try {
  await bot.start();
} catch {}
```

```txt
grammy:bot Initializing bot +0ms
grammy:core Calling getMe +0ms
grammy:core Calling deleteWebhook +9ms
grammy:bot I am he_exp_bot! +605ms
grammy:bot Starting simple long polling +2ms
grammy:error msg (401: abc) +0ms
abc
true
grammy:bot Simple long polling already running! +3ms
```

KnorpelSenf:
Correct. That was a bug.

PR #645, opened 2024-10-16
chore: update deps inside CI config

KnorpelSenf:
Updates CI to Deno 2, and updates the checkout action to v4. Makes tests compatible with Deno 2.


KnorpelSenf:
thx!

PR #651, opened 2024-10-27
Migrate tests from deno.land to JSR

KnorpelSenf:
This PQ
- fixes up the old context types tests and aligns them with the composer type tests
- migrates the test deps to JSR
- stops depending on various deno.land deps to mock types and uses `npm:` imports instead


test/context.type.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import {
+    assertType,
+    describe,
+    type Has,
+    type IsExact,
+    it,
+} from "./deps.test.ts";
+
+describe(
+    "context.has",
```
KnorpelSenf, line 11: We should probably call it
```suggestion
    "ctx.has* tests",
```

test/context.type.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import {
+    assertType,
+    describe,
+    type Has,
+    type IsExact,
+    it,
+} from "./deps.test.ts";
+
+describe(
+    "ctx.has* tests",
```
KnorpelSenf, line 11: Or this?
```suggestion
    "ctx.has* checks",
```

test/context.type.test.ts
```diff
@@ -0,0 +1,50 @@
...
+                >(true);
+            }
+            if (ctx.hasInlineQuery("123")) {
+                assertType<IsExact<typeof ctx.inlineQuery.id, string>>(
+                    true,
+                );
+            }
+        });
+        c.command("c", (ctx) => {
+            assertType<Has<typeof ctx.match, string>>(true);
```
Loskir, line 47: Here it should be IsExact instead of Has, maybe? Since we don't want it to be RegExpMatchArray

KnorpelSenf: Good point, I'll fix it

KnorpelSenf:
I am trying to contribute `IsMutuallyAssignable` to `@std`: https://github.com/denoland/std/pull/6154

Let's see what they think of it. If they merge, we can remove our own declaration of the type, and rely on `@std` instead.

Loskir (approved):
LGTM 👍

PR #658, opened 2024-11-09
fix: handle abort signal for methods without payload

KnorpelSenf:
As reported in https://t.me/grammyjs/279613


KnorpelSenf:
This will be reverted for 2.0 when we can unify the function signatures of all of `RawApi` in a breaking change. It will therefore not be needed for long to keep these values in sync with the Bot API.

KnightNiwrem (approved):

PR #660, opened 2024-11-10
fix: implement custom fetch handling in ApiClient for Cloudflare compatibility

### LINKED ISSUE #659: After update with big extimol i.e. 1.31.1 cloudflare workers stopped working, I think it was because fetch was added to this ApiClientOptions.
After update with big extimol i.e. 1.31.1 cloudflare workers stopped working, I think it was because fetch was added to this ApiClientOptions.

Error log in Cloidflare workers:
TypeError: Illegal invocation: function called with incorrect `this` reference.

image

_Originally posted by @BilagoNet in https://github.com/grammyjs/grammY/issues/655#issuecomment-2466660778_

IKatsuba:
Improved `fetch` handling in `ApiClient` to ensure compatibility with Cloudflare workers by calling `fetch` without context

Closed #659

### DIFF
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -252,7 +252,14 @@ class ApiClient<R extends RawApi> {
             },
             canUseWebhookReply: options.canUseWebhookReply ?? (() => false),
             sensitiveLogs: options.sensitiveLogs ?? false,
-            fetch: options.fetch ?? fetch,
+            // In an ideal world, `fetch` is independent of the context being called,
+            // but in a Cloudflare worker, any context other than global throws an error.
+            // That is why we need to call custom fetch or fetch without context.
+            fetch: ((...args: Parameters<typeof fetch>) => {
+                const { fetch: customFetch } = options;
+
+                return (customFetch ?? fetch)(...args);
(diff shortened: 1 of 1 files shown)


src/core/client.ts
```diff
@@ -252,7 +252,14 @@ class ApiClient<R extends RawApi> {
...
             canUseWebhookReply: options.canUseWebhookReply ?? (() => false),
             sensitiveLogs: options.sensitiveLogs ?? false,
-            fetch: options.fetch ?? fetch,
+            // In an ideal world, `fetch` is independent of the context being called,
+            // but in a Cloudflare worker, any context other than global throws an error.
+            // That is why we need to call custom fetch or fetch without context.
+            fetch: ((...args: Parameters<typeof fetch>) => {
+                const { fetch: customFetch } = options;
+
+                return (customFetch ?? fetch)(...args);
```
KnorpelSenf, line 261: You're now checking the options object for every single request. That is in the hottest code path of this library. Can you change it so that the option is only checked once, and two different lambdas are created?

IKatsuba: Yes, I can

PR #665, opened 2024-11-13
feat: add custom prefix for session key

KnorpelSenf:
With multi sessions and the conversations plugin, storage adapters are used in many places. It needs to be made simple to store session data under namespaces. This PQ does that by adding a new option `prefix` to session config. The prefix will be prepended to every session key before reading, writing, or deleting data.

### DIFF
--- a/src/convenience/session.ts
+++ b/src/convenience/session.ts
@@ -119,6 +119,15 @@ export interface SessionOptions<S, C extends Context = Context> {
      */
     initial?: () => S;
     /**
+     * A optional prefix to prepend to the session key after it was generated.
+     *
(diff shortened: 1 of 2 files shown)


src/convenience/session.ts
```diff
@@ -118,6 +118,15 @@ export interface SessionOptions<S, C extends Context = Context> {
      * session data.
      */
     initial?: () => S;
+    /**
+     * A optional prefix to prepend to the session key after it was generated.
```
KnorpelSenf, line 122:
```suggestion
     * An optional prefix to prepend to the session key after it was generated.
```

KnightNiwrem (approved):
Test cases not verified by this review

Issue #686, opened 2024-12-15
feat: `bot.command` should work on captions


KnorpelSenf:
Title says it all, see #164

KnorpelSenf:
Done on the `v2` branch in f12bb4f35bc02695881c30c5dad1e590aab49e31

PR #732, opened 2024-12-31
feat: add `ctx.txt`

### LINKED ISSUE #731: feat: add context shortcut `ctx.txt` for `msg.text` and `msg.caption`
[discussion](https://t.me/grammyjs/286914)

Similar thing could be done for `ctx.audio` `ctx.document`, etc...

carafelix:
closes #731

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -522,7 +522,7 @@ export class Context implements RenamedUpdate {
     }
     /**
      * Get the message text from wherever possible. Alias for `this.msg?.text ??
-     * this.msg?.caption
+     * this.msg?.caption`.
      */
     get txt(): string | undefined {
         // Keep in sync with types in `filter.ts`.
--- a/src/context.ts
+++ b/src/context.ts
@@ -521,6 +521,16 @@ export class Context implements RenamedUpdate {
         );
     }
     /**
+     * Get the message text from wherever possible. Alias for `this.msg?.text ??
+     * this.msg?.caption
+     */
+    get txt(): string | undefined {
+        // Keep in sync with types in `filter.ts`.
+        return (
+            this.msg?.text ?? this.msg?.caption
+        );
+    }
+    /**
      * Get the chat object from wherever possible. Alias for `(this.msg ??
      * this.deletedBusinessMessages ?? this.messageReaction ??
      * this.messageReactionCount ?? this.myChatMember ??  this.chatMember ??
--- a/src/context.ts
+++ b/src/context.ts
@@ -526,9 +526,8 @@ export class Context implements RenamedUpdate {
      */
     get txt(): string | undefined {
         // Keep in sync with types in `filter.ts`.
-        return (
-            this.msg?.text ?? this.msg?.caption
-        );
+        const msg = this.msg;
+        return msg?.text ?? msg?.caption;
     }
     /**
      * Get the chat object from wherever possible. Alias for `(this.msg ??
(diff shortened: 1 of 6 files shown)


src/filter.ts
```diff
@@ -601,6 +601,7 @@ interface Shortcuts<U extends Update> {
         : [U["callback_query"]] extends [object]
             ? U["callback_query"]["message"]
         : undefined;
+    txt: [Shortcuts<U>["msg"]] extends [object] ? string : undefined;
```
KnorpelSenf, line 604: Not all messages have text. You need to map this to the actual runtime behaviour.

KnorpelSenf (changes requested):
The hard part is missing :D

src/filter.ts
carafelix: Now its fine, I think(?)

KnorpelSenf (changes requested):
Two more things: 
- please add the respective tests 
- please change it so `ctx.msg` is only computed once

src/filter.ts
KnorpelSenf: I believe so, but I haven't tested it myself

wojpawlik (dismissed):
Please add a test in `composer.type.test.ts` showing that `.txt` is `string` `.on([":text", ":caption"])`.

carafelix:
Test case for `.on([":text", ":caption"])` is failing while `.on(":text")` and `.on(":caption")` separately, are not.

Type definition should be adjusted accordingly. But how?

KnorpelSenf:
Continuing in https://t.me/grammyjs/289375

src/filter.ts
```diff
@@ -601,6 +601,11 @@ interface Shortcuts<U extends Update> {
         : [U["callback_query"]] extends [object]
             ? U["callback_query"]["message"]
         : undefined;
+    txt: [Shortcuts<U>["msg"]] extends [object]
+        ? (Shortcuts<U>["msg"]["text"] extends string ? string
```
KnorpelSenf, line 605:
```suggestion
        ? ([Shortcuts<U>["msg"]["text"]] extends [string] ? string
```

src/filter.ts
```diff
@@ -601,6 +601,11 @@ interface Shortcuts<U extends Update> {
...
+            : Shortcuts<U>["msg"]["caption"] extends string ? string
```
KnorpelSenf, line 606:
```suggestion
            : [Shortcuts<U>["msg"]["caption"]] extends [string] ? string
```

src/context.ts
```diff
@@ -520,6 +520,15 @@ export class Context implements RenamedUpdate {
                 this.callbackQuery?.message
         );
     }
+    /**
+     * Get the message text from wherever possible. Alias for `this.msg?.text ??
+     * this.msg?.caption
```
KnorpelSenf, line 525:
```suggestion
     * this.msg?.caption`.
```

test/filter.test.ts
```diff
@@ -3,7 +3,9 @@ import {
     assert,
     assertEquals,
     assertThrows,
+    assertType,
     describe,
+    IsNullable,
```
KnorpelSenf, line 8:
```suggestion
    type IsNullable,
```

carafelix:
Since this is not breaking, should it point towards v1 instead?

KnorpelSenf:
Nah, no need. Using `ctx.msg.text` is fine for v1. It's only after #686 that `ctx.txt` gets really useful, and that one will also only land in v2.

In addition, I'd like all feature development to go towards v2 directly so that we can design and test these features together. We might end up changing them if they affect each other in unexpected ways. After we have figured this out, we can think about backporting the most important ideas. In contrast, if we were to introduce them to v1 first, we might end up breaking them immediately after, which causes needles disruption.

KnorpelSenf (approved):
Excellent work!

PR #733, opened 2024-12-31
feat(`Composer`): match any command

### LINKED ISSUE #685: feat: `bot.command([])` should match any bot command
This is inconsistent with how empty disjunctions work in logic, but who cares. We need this feature and there is no way of doing it yet. That syntax would be very concise.

Are there better ideas, such as passing no arguments (=making the first param optional)? Maybe both?

wojpawlik:
Happy new year!
Closes #685

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -82,7 +82,7 @@ interface StaticHas {
      * @param command The command to match
      */
     command(
-        command: MaybeArray<StringWithCommandSuggestions>,
+        command?: MaybeArray<StringWithCommandSuggestions>,
     ): <C extends Context>(ctx: C) => ctx is CommandContext<C>;
     /**
      * Generates a predicate function that can test context objects for
@@ -186,9 +186,8 @@ const checker: StaticHas = {
     },
     command(command) {
         const hasEntities = checker.filterQuery(":entities:bot_command");
-        const atCommands = new Set<string>();
-        const noAtCommands = new Set<string>();
-        toArray(command).forEach((cmd) => {
+        const noAtCommands = new Set<string>(toArray(command ?? []));
+        for (const cmd of noAtCommands) {
             if (cmd.startsWith("/")) {
                 throw new Error(
                     `Do not include '/' when registering command handlers (use '${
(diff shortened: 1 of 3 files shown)


src/context.ts
```diff
@@ -186,43 +186,38 @@ const checker: StaticHas = {
...
-                if (e.type !== "bot_command") return false;
-                if (e.offset !== 0) return false;
-                const cmd = txt.substring(1, e.length);
-                if (noAtCommands.has(cmd) || atCommands.has(cmd)) {
-                    ctx.match = txt.substring(cmd.length + 1).trimStart();
-                    return true;
-                }
-                const index = cmd.indexOf("@");
-                if (index === -1) return false;
+            const e: MessageEntity = msg.entities[0];
```
carafelix, line 203:
```suggestion
            const e : MessageEntity = msg.entities[0] ?? msg.caption_entities?.[0];
```
would also close #686

carafelix (commented):
WDYT?

src/context.ts
KnorpelSenf: No. The order of message entities is unspecified. The bot command at index 0 might be the last entity.

carafelix: So using `entities[0]` is error prone, for example a msg like `/start /end` could match `/end` using this approach, if I'm not mistaken

KnorpelSenf: Exactly

src/context.ts
```diff
@@ -185,39 +185,36 @@ const checker: StaticHas = {
...
+            return entities.some((e) => {
                 if (e.type !== "bot_command") return false;
                 if (e.offset !== 0) return false;
                 const cmd = txt.substring(1, e.length);
-                if (noAtCommands.has(cmd) || atCommands.has(cmd)) {
-                    ctx.match = txt.substring(cmd.length + 1).trimStart();
-                    return true;
+                let index = cmd.indexOf("@");
+                if (index === -1) {
+                    index = Infinity;
```
KnorpelSenf, line 210: This is a big of a hack, isn't it?

test/composer.type.test.ts
```diff
@@ -110,7 +110,7 @@ describe("Composer types", () => {
                 const channelPostCaption = ctx.channelPost?.caption;
                 const channelPostText = ctx.channelPost?.text;
                 const match = ctx.match;
-                assertType<IsExact<typeof msgText, string>>(true);
+                assertType<IsExact<typeof msgText, string | undefined>>(true);
```
KnorpelSenf, line 113: Could you also assert that `ctx.msg.text ?? ctx.msg.caption` is `string`? If it isn't, then I'd like to fix up the filter queries first.

wojpawlik: It's `string | undefined`, no idea why :(

@carafelix #686 required touching 3 files, not 1 line, and uncovered a bug.

src/context.ts
wojpawlik: Not if `.substring()` was replaced with `.slice()`, or
```suggestion
                    index = cmd.length;
```
`undefined` would require adding a type annotation.

test/composer.type.test.ts
KnorpelSenf: It's odd to me that this seems to work correctly for `entities`→`text` but fails for `caption_entities`→`caption`. Smells like
- it's related to bad type distribution or 
- the above assumption is wrong

wojpawlik: `if (ctx.has(":caption_entities")) ctx.msg.caption` actually is `string`, as expected... Only `msg.text ?? msg.caption` breaks.

wojpawlik: Changing https://github.com/grammyjs/grammY/blob/a55168c7532be0761aa0ce54496ad7e0d25d68a5/src/temporary_types/message.ts#L66 to required "fixes" the issue

KnorpelSenf: > `if (ctx.has(":caption_entities")) ctx.msg.caption` actually is `string`, as expected... Only `msg.text ?? msg.caption` breaks.

Very interesting, that fits my intuition, but I'll still have to investigate it properly

KnorpelSenf:
Superseded on the `v2` branch by f12bb4f35bc02695881c30c5dad1e590aab49e31

Issue #737, opened 2025-01-15
Support Azure Functions v4 Model for `webhookCallback`


mildronize:
## Background

First of all, I’d like to thank the team behind grammy for creating such an amazing Telegram framework! I’ve enjoyed working with it, especially on [my open-source bot using Cloudflare Workers](https://github.com/mildronize/grammybot).

Recently, I tried to use the `webhookCallback` feature with Azure Functions v4 ([the latest official version of the Node.js runtime](https://learn.microsoft.com/en-us/azure/azure-functions/functions-node-upgrade-v4?tabs=v4&pivots=programming-language-javascript)).

However, I encountered an issue where Cloudflare Workers did not fully support Node.js APIs ([details here](https://github.com/mildronize/grammybot/issues/2)). As a result, I returned to Azure Functions for deployment. After reviewing the `webhookCallback` interface, I found that it works well with the Azure Functions v3 model (as tested with the following code):

```ts
import { webhookCallback } from 'grammy';
await webhookCallback(bot, "azure")(request, context);
```

## Environment

- **Runtime:** Node.js 20
- **grammy Version:** 1.34.0

## Issue Summary

1. The [grammy documentation](https://grammy.dev/guide/deployment-types#web-framework-adapters) does not specify which versions of Azure Functions are supported. Based on my tests, it seems only Azure Functions v3 is compatible, and Azure Functions v4 is not yet supported.
2. To assist others, I created a [template for grammy with Azure Functions v3](https://github.com/thaitype/grammy-nammatham-v1) using [Nammatham](https://github.com/thaitype/nammatham) v1, to help developers get started more easily.
3. This is not a complaint! The work you’ve done is fantastic, and if I have some free time, I’d be happy to contribute to Azure Functions v4 support. If the maintainers can guide me on where to start, I’d greatly appreciate it.

Thank you for taking the time to read this issue, and keep up the great work!

KnorpelSenf:
Hi! Thanks for the kind words!

Re-reading the code, I am actually surprised that grammY 1.34 works with v3. Isn't v3 supposed to have the request as the second argument?

https://github.com/grammyjs/grammY/blob/b04bfee62e944b0080f0ff5862b1b7ecd0fba6c4/src/convenience/frameworks.ts#L276-L291

It looks like the order of context and request is wrong.

That being said, we clearly need a new adapter to support v4.

I opened # which should fix both issues. You can install it from GitHub via
```sh
[...]

PR #738, opened 2025-01-15
feat: add support for Azure Functions v4

KnorpelSenf:
Adds a new webook adapter following https://learn.microsoft.com/en-us/azure/azure-functions/functions-node-upgrade-v4

Fixes https://github.com/grammyjs/grammY/issues/737

### DIFF
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -87,7 +87,10 @@ export type AzureAdapter = (context: {
     };
 }, request: { body?: unknown }) => ReqResHandler;
 export type AzureAdapterV4 = (
-    request: { headers: Headers; json(): Promise<Update> },
+    request: {
+        headers: { get(name: string): string | null };
+        json(): Promise<unknown>;
+    },
 ) => ReqResHandler<{ status: number; body?: string } | { jsonBody: string }>;
 
 export type BunAdapter = (request: {
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -82,6 +82,7 @@ export type LambdaAsyncAdapter = (
 
 export type AzureAdapter = (context: {
     res?: {
+        // deno-lint-ignore no-explicit-any
         [key: string]: any;
     };
 }, request: { body?: unknown }) => ReqResHandler;
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -86,13 +86,23 @@ export type AzureAdapter = (context: {
         [key: string]: any;
     };
 }, request: { body?: unknown }) => ReqResHandler;
-export type AzureAdapterV4 = (request: Request, context: {
-    res?: {
-        status: number;
-        body: string;
-        headers?: Record<string, string>;
-    };
-}) => ReqResHandler<{ status: number; body?: string } | { jsonBody: string }>;
+type AzureV4HttpRequest = {
+    method: string;
+    url: string;
+    headers: any;
+    query: URLSearchParams;
+    body: any;
+}
+type AzureV4InvocationContext = {
+    invocationId: string;
+    functionName: string;
+}
+export type AzureAdapterV4 = (request: AzureV4HttpRequest, context: AzureV4InvocationContext) => ReqResHandler<{
+    status: number;
+    body?: string;
+} | {
+    jsonBody: string;
+}>;
 
 export type BunAdapter = (request: {
     headers: Headers;
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -82,14 +82,7 @@ export type LambdaAsyncAdapter = (
 
 export type AzureAdapter = (context: {
     res?: {
-        status: number;
-        body: string;
-        headers?: Record<string, string>;
-        set?: (key: string, value: string) => void;
-        send?: {
-            (body: unknown): void;
-            (status: number, body: unknown): void;
-        };
+        [key: string]: any;
     };
 }, request: { body?: unknown }) => ReqResHandler;
 export type AzureAdapterV4 = (request: Request, context: {
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -86,23 +86,9 @@ export type AzureAdapter = (context: {
         [key: string]: any;
     };
 }, request: { body?: unknown }) => ReqResHandler;
-type AzureV4HttpRequest = {
-    method: string;
-    url: string;
-    headers: any;
-    query: URLSearchParams;
-    body: any;
-}
-type AzureV4InvocationContext = {
-    invocationId: string;
-    functionName: string;
-}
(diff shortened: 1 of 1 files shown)


src/convenience/frameworks.ts
```diff
@@ -80,20 +80,18 @@ export type LambdaAsyncAdapter = (
     _context: unknown,
 ) => ReqResHandler;
 
-export type AzureAdapter = (request: {
-    body?: unknown;
-}, context: {
+export type AzureAdapter = (context: {
+    res?: {
+        [key: string]: any;
```
KnorpelSenf, line 85:
```suggestion
        // deno-lint-ignore no-explicit-any
        [key: string]: any;
```

src/convenience/frameworks.ts
```diff
@@ -289,6 +298,23 @@ const azure: AzureAdapter = (request, context) => ({
...
     },
 });
+const azureV4: AzureAdapterV4 = (request, context) => {
+    type Res = NonNullable<
+        Awaited<ReturnType<AzureAdapterV4>["handlerReturn"]>
+    >;
+    let resolveResponse: (response: Res) => void;
+    return {
+        update: Promise.resolve(request.json()) as Promise<Update>,
+        header: context.res?.headers?.[SECRET_HEADER],
```
KnorpelSenf, line 308: Here, we access `context.res`

src/convenience/frameworks.ts
```diff
@@ -80,20 +80,29 @@ export type LambdaAsyncAdapter = (
...
+    method: string;
+    url: string;
+    headers: any;
+    query: URLSearchParams;
+    body: any;
+}
+type AzureV4InvocationContext = {
+    invocationId: string;
+    functionName: string;
+}
```
KnorpelSenf, line 99: but it is not defined here, so the code does not compile

src/convenience/frameworks.ts
```diff
@@ -80,20 +80,29 @@ export type LambdaAsyncAdapter = (
...
+type AzureV4HttpRequest = {
+    method: string;
+    url: string;
+    headers: any;
+    query: URLSearchParams;
+    body: any;
+}
+type AzureV4InvocationContext = {
+    invocationId: string;
+    functionName: string;
```
KnorpelSenf, line 98: In addition, these properties here are defined needlessly. They can be dropped because grammY never accesses them. TypeScript's structural type system will still allow passing in objects that have these properties, so the code will continue to work for library consumers.

src/convenience/frameworks.ts
```diff
@@ -80,20 +80,29 @@ export type LambdaAsyncAdapter = (
...
+        [key: string]: any;
     };
-}) => ReqResHandler;
+}, request: { body?: unknown }) => ReqResHandler;
+type AzureV4HttpRequest = {
+    method: string;
+    url: string;
+    headers: any;
+    query: URLSearchParams;
+    body: any;
```
KnorpelSenf, line 94: Similarly, we should only define the necessary fields here.

mildronize:
@KnorpelSenf I see, actually i don't know which prop is used, or I should be something like this

```ts
export type AzureAdapterV4 = (request: {}, context: {}) => ReqResHandler<{
    status: number;
    body?: string;
} | {
    jsonBody: string;
}>;
```

src/convenience/frameworks.ts
KnorpelSenf: Oh I see, we should use `request.headers.get('content-type')` instead. We can drop the context object entirely, I think.

KnorpelSenf:
No worries, I can fix it up later today.

KnorpelSenf:
@mildronize please check again

src/convenience/frameworks.ts
```diff
@@ -80,20 +80,15 @@ export type LambdaAsyncAdapter = (
...
-            (body: unknown): void;
-            (status: number, body: unknown): void;
-        };
+        // deno-lint-ignore no-explicit-any
+        [key: string]: any;
     };
-}) => ReqResHandler;
+}, request: { body?: unknown }) => ReqResHandler;
+export type AzureAdapterV4 = (
+    request: { headers: Record<string, string>; json(): Promise<Update> },
```
KnorpelSenf, line 90: Especially this

KnorpelSenf: fixed

src/convenience/frameworks.ts
```diff
@@ -289,6 +284,23 @@ const azure: AzureAdapter = (request, context) => ({
         context.res?.send?.(401, WRONG_TOKEN_ERROR);
     },
 });
+const azureV4: AzureAdapterV4 = (request) => {
+    type Res = NonNullable<
+        Awaited<ReturnType<AzureAdapterV4>["handlerReturn"]>
+    >;
```
mildronize, line 290: CleanShot 2568-01-17 at 19 14 36

@KnorpelSenf After I've check out the new code from the latest commit from https://github.com/grammyjs/grammY/pull/738, (Commit id: `8f9e16e5114421dc0dacab51f899f7d4538f0184`)

you checkout the code from here: https://github.com/mildronize/grammy-azure-functions-test/tree/main/func-v4

the request type error:

```
Argument of type 'HttpRequest' is not assignable to parameter of type '{ headers: Headers; json(): Promise<Update>; }'.
  The types returned by 'headers[Symbol.iterator]()' are incompatible between these types.
    Property '[Symbol.iterator]' is missing in type 'SpecIterator<[string, string], any, undefined>' but required in type 'HeadersIterator<[string, string]>
```

src/convenience/frameworks.ts
```diff
@@ -80,20 +80,15 @@ export type LambdaAsyncAdapter = (
...
-            (body: unknown): void;
-            (status: number, body: unknown): void;
-        };
+        // deno-lint-ignore no-explicit-any
+        [key: string]: any;
     };
-}) => ReqResHandler;
+}, request: { body?: unknown }) => ReqResHandler;
+export type AzureAdapterV4 = (
+    request: { headers: Headers; json(): Promise<Update> },
```
mildronize, line 90: Oh, I comment the wrong place, here is what I mentioned in the comment below.

KnorpelSenf:
I'll fix this up soon

KnorpelSenf:
@mildronize sorry for the delay, I fixed both the readers and the JSON type. Can you check again?

mildronize:
@KnorpelSenf Noted, i'll review

mildronize:
@KnorpelSenf Great for minimal type model for Azure v4 interface, sorry for being late review, 

I've checkout the latest commit of this pr and type check correctly, Approved!. Thank you so much. 

You can checkout the type check code here: https://github.com/mildronize/grammy-azure-functions-test/commit/c19c5700c998e4102cd9c7ca80bffdaf5a25bdb9

iScreen Shoter - Google Chrome - 680222115420
iScreen Shoter - 25680222115521408

mildronize (approved):
Both type of Azure Functions v3 and v4 are correct. thanks, approved

KnorpelSenf:
@all-contributors add @mildronize for bug and test and review

KnorpelSenf:
Thank _you_ for the great work!

PR #739, opened 2025-01-15
type: fix Azure Function v3 Adapter Context Type

mildronize:
As mentioned in https://github.com/grammyjs/grammY/issues/737#issuecomment-2593526070


KnorpelSenf (approved):
LGTM

PR #740, opened 2025-01-16
feat:  add possibility to catch errors with `bot.catch` on webhook

shevernitskiy:
This PR adds error handling with `bot.catch` for webhooks.

I have several questions:

1. Should we log to the console if no handler is set in case of webhook (in the case of a default handler)?
2. Is it necessary to extend the `try` scope to cover the cases of `new Api`, `new Context`, and `this.botInfo === undefined`? I remove `try-catch` on a higher level caller `handleUpdates` which is using on polling.


shevernitskiy:
I think i do pr to wrong branch:)

KnorpelSenf:
You could've just edited the target branch :D

shevernitskiy:
> You could've just edited the target branch :D

i just have skill issue with github :)

KnorpelSenf:
Hehe no problem, I'll review #741 in a bit

Issue #702, opened 2024-12-15
feat: make `bot.catch` work on webhooks


KnorpelSenf:
Many developers are confused by this so we should just respect any error handlers if they are set. If no error handlers are set via `bot.catch`, errors will still be passed to the surrounding HTTP server lib.

KnorpelSenf:
Fixed by #741

KnorpelSenf:
@shevernitskiy we might want to do https://github.com/grammyjs/grammY/issues/697#issuecomment-2597393966 now that this is implemented

PR #741, opened 2025-01-16
 feat: add possibility to catch errors with `bot.catch` on webhook

shevernitskiy:
This PR adds error handling with bot.catch for webhooks.

I have several questions:

Should we log to the console if no handler is set in case of webhook (in the case of a default handler)?
Is it necessary to extend the try scope to cover the cases of new Api, new Context, and this.botInfo === undefined? I remove try-catch on a higher level caller handleUpdates which is using on polling.

Closes https://github.com/grammyjs/grammY/issues/702

### DIFF
--- a/test/convenience/webhook.test.ts
+++ b/test/convenience/webhook.test.ts
@@ -11,9 +11,9 @@ import type bodyParser from "npm:@types/koa-bodyparser";
 import type Koa from "npm:@types/koa";
 import type { FastifyInstance } from "npm:fastify";
 import type { NextApiRequest, NextApiResponse } from "npm:next";
-import { Bot, webhookCallback } from "../../src/mod.ts";
+import { Bot, BotError, webhookCallback } from "../../src/mod.ts";
 import type { UserFromGetMe } from "../../src/types.ts";
-import { describe, it } from "../deps.test.ts";
+import { assertIsError, describe, it } from "../deps.test.ts";
 
 describe("webhook", () => {
     const bot = new Bot("dummy", { me: {} as unknown as UserFromGetMe });
@@ -137,5 +137,25 @@ describe("webhook", () => {
         serve((req) => {
             return handler(req);
         });
+    });
+
+    it("bot should catch errors in webhoook handler", async () => {
+        bot.catch((err) => {
+            assertIsError(err, BotError, "Test Error");
+        });
+        bot.on("message", () => {
+            throw new Error("Test Error");
+        });
+
+        const handler = webhookCallback(bot, "std/http");
+        const fake_req = new Request("https://fake-api.com", {
+            method: "POST",
+            body: JSON.stringify({
+                update_id: 9696,
+                message: {},
+            }),
+        });
+
+        await handler(fake_req);
     });
 });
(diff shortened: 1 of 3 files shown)


test/convenience/webhook.test.ts
```diff
@@ -138,4 +138,24 @@ describe("webhook", () => {
             return handler(req);
         });
     });
+
+    it("bot should catch errors in webhoook handler", async () => {
+        bot.catch((err) => {
+            assertIsError(err, BotError, "Test Error");
+        });
```
KnorpelSenf, line 145: Can you assert that this handler is actually executed? Otherwise `assertIsError` might just never be called :D

test/convenience/webhook.test.ts
```diff
@@ -138,4 +138,24 @@ describe("webhook", () => {
...
+        bot.on("message", () => {
+            throw new Error("Test Error");
+        });
+
+        const handler = webhookCallback(bot, "std/http");
+        const fake_req = new Request("https://fake-api.com", {
```
KnorpelSenf, line 151: nit
```suggestion
        const fakeReq = new Request("https://fake-api.com", {
```

test/convenience/webhook.test.ts
```diff
@@ -138,4 +138,24 @@ describe("webhook", () => {
...
+            method: "POST",
+            body: JSON.stringify({
+                update_id: 9696,
+                message: {},
+            }),
+        });
+
+        await handler(fake_req);
```
KnorpelSenf, line 159: and
```suggestion
        await handler(fakeReq);
```

KnorpelSenf (changes requested):
These changes are a breath of fresh air, you made an excellent observation about our ability to drop the whole FATAL error thing. Thank you!

I just have a tiny request about the assertion in the test:

shevernitskiy:
this should work:)

test/convenience/webhook.test.ts
```diff
@@ -11,9 +11,9 @@ import type bodyParser from "npm:@types/koa-bodyparser";
 import type Koa from "npm:@types/koa";
 import type { FastifyInstance } from "npm:fastify";
 import type { NextApiRequest, NextApiResponse } from "npm:next";
-import { Bot, webhookCallback } from "../../src/mod.ts";
+import { Bot, BotError, webhookCallback } from "../../src/mod.ts";
 import type { UserFromGetMe } from "../../src/types.ts";
-import { describe, it } from "../deps.test.ts";
+import { assertEquals, assertIsError, describe, it } from "../deps.test.ts";
```
KnorpelSenf, line 16:
```suggestion
import { assert, assertIsError, describe, it } from "../deps.test.ts";
```

test/convenience/webhook.test.ts
```diff
@@ -138,4 +138,29 @@ describe("webhook", () => {
...
+            method: "POST",
+            body: JSON.stringify({
+                update_id: 9696,
+                message: {},
+            }),
+        });
+
+        await handler(fakeReq);
+
+        assertEquals(called, true);
```
KnorpelSenf, line 164:
```suggestion
        assert(called);
```

KnorpelSenf (changes requested):

KnorpelSenf (approved):
Thank you so much!

PR #742, opened 2025-01-16
type: fix Azure Function v4 Adapter HttpRequest and InvocationContext Type

mildronize:
As mentioned in https://github.com/grammyjs/grammY/issues/737#issuecomment-2595376501

### DIFF
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -85,13 +85,23 @@ export type AzureAdapter = (context: {
         [key: string]: any;
     };
 }, request: { body?: unknown }) => ReqResHandler;
-export type AzureAdapterV4 = (request: Request, context: {
-    res?: {
-        status: number;
-        body: string;
-        headers?: Record<string, string>;
-    };
-}) => ReqResHandler<{ status: number; body?: string } | { jsonBody: string }>;
+export type AzureV4HttpRequest = {
+    method: string;
+    url: string;
+    headers: any;
+    query: URLSearchParams;
+    body: any;
+}
+export type AzureV4InvocationContext = {
+    invocationId: string;
+    functionName: string;
+}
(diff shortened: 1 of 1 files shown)


src/convenience/frameworks.ts
```diff
@@ -85,13 +85,23 @@ export type AzureAdapter = (context: {
...
     };
 }, request: { body?: unknown }) => ReqResHandler;
-export type AzureAdapterV4 = (request: Request, context: {
-    res?: {
-        status: number;
-        body: string;
-        headers?: Record<string, string>;
-    };
-}) => ReqResHandler<{ status: number; body?: string } | { jsonBody: string }>;
+export type AzureV4HttpRequest = {
```
KnorpelSenf, line 88: We also don't need to export this, it isn't used elsewhere :)
```suggestion
type AzureV4HttpRequest = {
```

src/convenience/frameworks.ts
```diff
@@ -85,13 +85,23 @@ export type AzureAdapter = (context: {
...
+    method: string;
+    url: string;
+    headers: any;
+    query: URLSearchParams;
+    body: any;
+}
+export type AzureV4InvocationContext = {
```
KnorpelSenf, line 95: Same here
```suggestion
type AzureV4InvocationContext = {
```

KnorpelSenf (changes requested):
It is not necessary to copy all properties of the type definition to grammY, most of them are not needed. Can you reduce the type definitions to only what's needed here?

src/convenience/frameworks.ts
mildronize: Ok, thanks

src/convenience/frameworks.ts
mildronize: Ok, thanks

mildronize:
@KnorpelSenf Is it look good now?

KnorpelSenf:
I think this actually won't compile, but let's fix the last few things in #738

KnorpelSenf (approved):

Issue #695, opened 2024-12-15
refactor: rename `webhookCallback(bot, "express")` to `webhookAdapter.express(bot)`


KnorpelSenf:
Title says it all. Passing a string was a weird design decision.

Issue #696, opened 2024-12-15
feat: remove `"ignore"` from webhooks adapter config


KnorpelSenf:
All use cases of this option lead to incorrect bot code. There is no reason to ever use this option, so why have it?

The old behaviour can be retained by not awaiting `next` in middleware. Just in case somebody wants to stick with their existing broken code.

KnorpelSenf:
Closed by https://github.com/grammyjs/grammY/pull/755

PR #743, opened 2025-01-16
refactor(webhook): change the way to get webhook handler 

shevernitskiy:
PR aiming to close https://github.com/grammyjs/grammY/issues/695 and https://github.com/grammyjs/grammY/issues/696

I'm not good at typing, so I'm requesting help. There are couple "as" in a commit :)

I created a factory of factories to dynamically generate object with creators of callback handlers. I do not know how to document every object property in this case. If it is not necessary, we can just describe the main object.

Regarding the "return" type, I decided to rename it to "ignore" because it does not make any calls and leads to a safe logic branch. The user can pass an empty callback as an argument, and it will work pretty much the same, but with an unnecessary extra call.

So, if we want to slim down and simplify the API, we need to remove "ignore." If we want to gain a nano performance benefit, we should keep it.

### DIFF
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -138,12 +138,8 @@ export type HonoAdapter = (c: {
         json: <T>() => Promise<T>;
         header: (header: string) => string | undefined;
     };
-    body: (
-        data: string | ArrayBuffer | ReadableStream | null,
-        // deno-lint-ignore no-explicit-any
-        arg?: any,
-        headers?: Record<string, string | string[]>,
-    ) => Response;
+    body(data: string): Response;
+    body(data: null, status: 204): Response;
     // deno-lint-ignore no-explicit-any
     status: (status: any) => void;
     json: (json: string) => Response;
@@ -390,14 +386,14 @@ const hono: HonoAdapter = (c) => {
         update: c.req.json(),
         header: c.req.header(SECRET_HEADER),
         end: () => {
-            resolveResponse(c.body(null));
+            resolveResponse(c.body(""));
         },
         respond: (json) => {
             resolveResponse(c.json(json));
         },
         unauthorized: () => {
             c.status(401);
-            resolveResponse(c.body(null));
+            resolveResponse(c.body(""));
         },
         handlerReturn: new Promise<Response>((resolve) => {
             resolveResponse = resolve;
@@ -545,23 +541,23 @@ const worktop: WorktopAdapter = (req, res) => ({
 
 // Please open a pull request if you want to add another adapter
 export const adapters = {
-    "aws-lambda": awsLambda,
-    "aws-lambda-async": awsLambdaAsync,
+    awsLambda,
+    awsLambdaAsync,
     azure,
     bun,
     cloudflare,
-    "cloudflare-mod": cloudflareModule,
+    cloudflareModule,
     express,
     fastify,
     hono,
     http,
     https: http,
     koa,
-    "next-js": nextJs,
+    nextJs,
     nhttp,
     oak,
     serveHttp,
-    "std/http": stdHttp,
+    stdHttp,
     sveltekit,
     worktop,
 };
--- a/src/convenience/webhook.ts
+++ b/src/convenience/webhook.ts
@@ -96,26 +96,66 @@ function createWebhookAdapter<
  * @param webhookOptions Further options for the webhook setup
  */
 export const webhookAdapters = {
-    awsLambda: createWebhookAdapter(adapters.awsLambda),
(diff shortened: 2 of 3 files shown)


src/convenience/frameworks.ts
```diff
@@ -390,14 +386,14 @@ const hono: HonoAdapter = (c) => {
         update: c.req.json(),
         header: c.req.header(SECRET_HEADER),
         end: () => {
-            resolveResponse(c.body(null));
+            resolveResponse(c.body(""));
```
KnorpelSenf, line 389: What does this change?

shevernitskiy: I just copy hono part from main branch to fix broken test

shevernitskiy:
I realized, that in my implementation we have runtime iteration, so i decided to measure pref.
```sh
benchmark   time/iter (avg)        iter/s      (min … max)           p75      p99     p995    
----------- ----------------------------- --------------------- --------------------------    
callback            30.1 ns    33,220,000 ( 24.6 ns … 197.3 ns)  28.2 ns  64.5 ns  89.0 ns
adapters            36.4 ns    27,470,000 ( 30.7 ns … 263.8 ns)  33.9 ns  94.1 ns  99.3 ns
```

winstxnhdw:
Why not just unroll the loop? It doesn’t seem to require a runtime.

shevernitskiy:
> Why not just unroll the loop? It doesn’t seem to require a runtime.

You mean make static object with all adapters manually? Sure I can do if it your request to change. My initial intention was more generic solution.

winstxnhdw:
Someone who has to add support for a new adapter will have to create the new object anyways. A little more wouldn’t hurt. Also, CPU time is precious on a lambdaz

shevernitskiy:
> Someone who has to add support for a new adapter will have to create the new object anyways. A little more wouldn’t hurt. Also, CPU time is precious on a lambda.

Got it, will commit in couple hours

shevernitskiy:
Ok, now we have ability to make an example for each adapter. I leave as todo for now.

src/convenience/webhook.ts
```diff
@@ -34,8 +34,89 @@ export interface WebhookOptions {
...
+    https: createWebhookAdapter(adapters.http),
+    koa: createWebhookAdapter(adapters.koa),
+    nextJs: createWebhookAdapter(adapters.nextJs),
+    nhttp: createWebhookAdapter(adapters.nhttp),
+    oak: createWebhookAdapter(adapters.oak),
+    serveHttp: createWebhookAdapter(adapters.serveHttp),
+    stdHttp: createWebhookAdapter(adapters.stdHttp),
+    sveltekit: createWebhookAdapter(adapters.sveltekit),
+    worktop: createWebhookAdapter(adapters.worktop),
+    callback: createWebhookAdapter(adapters.callback),
```
wojpawlik, line 118: I'd use getters to avoid unnecessary `createWebhookAdapter` calls at startup.

shevernitskiy: nice catch!

winstxnhdw:
Since we are changing the API entirely, could we have a proper `onTimeout` argument? Currently, we are checking if it is a function or object. We can do away with that if we just have another argument for its own type.

src/convenience/webhook.ts
```diff
@@ -34,8 +34,129 @@ export interface WebhookOptions {
 
 type Adapters = typeof adapters;
 type AdapterNames = keyof Adapters;
-type ResolveName<A extends FrameworkAdapter | AdapterNames> = A extends
-    AdapterNames ? Adapters[A] : A;
+type Adapter<A extends Adapters[AdapterNames]> = (
+    ...args: Parameters<A>
+) => ReturnType<A>["handlerReturn"] extends undefined ? Promise<void>
+    : NonNullable<ReturnType<A>["handlerReturn"]>;
+type WebhookAdapter<
```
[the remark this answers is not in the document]
shevernitskiy: ah, i'm not as fast as you.. okay it would be more simple then:)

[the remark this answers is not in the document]
winstxnhdw: Ahh... All my hard work... Returned to dust... 

image

KnorpelSenf (approved):
Rubber-stamp LGTM. We can fix things later if we discover oddities. Long live the liberty of breaking changes!

KnorpelSenf:
We can make breaking changes until v2 is released. This will only happen when we no longer when to make breaking changes.

winstxnhdw:
Nice! I am still not too happy with this API yet. I'll think of something.

KnorpelSenf:
Looking forward to that!

Issue #675, opened 2024-12-15
Roadmap grammY 2.0


KnorpelSenf:
The following is a tentative list of issues that would  be cool to include in the next major release of grammY.

After 4 years of development without a major version, the grammY core library has accumulated a number of oddities that can only be fixed by doing breaking changes. grammY 2.0 is our opportunity to introduce these changes.

We may add or remove issues from the list below if we think that this makes sense.

There is no ETA. The release will happen when we have reached high quality. It goes faster if you contribute.

KnorpelSenf:
Beta versions are now available on JSR: https://jsr.io/@grammyjs/grammy/versions

KnorpelSenf:
The actual 2.0 release (whenever that happens) will be made available on npm, but the beta releases are going to be JSR-only for a while. You can install it using the npm CLI: `npx jsr add @grammyjs/grammy@2.0.0-beta.3`

roziscoding:
# Plugin Compatibility Checklist

- [ ] [Conversations](https://github.com/grammyjs/conversations/issues/153)
- [ ] [Interactive Menus](https://github.com/grammyjs/menu/issues/63)
- [ ] [Stateless Question](https://github.com/grammyjs/stateless-question/issues/93)
- [ ] [Concurrency](https://github.com/grammyjs/runner/issues/28)
- [ ] [Hydration](https://github.com/grammyjs/hydrate/issues/32)
- [ ] [Retry API Requests](https://github.com/grammyjs/auto-retry/issues/24)
- [ ] [Rate Limit Users](https://github.com/grammyjs/ratelimiter/issues/16) (@Amir-Zouerami)
- [ ] [Files](https://github.com/grammyjs/files/issues/19)
- [ ] [Internationalization (i18n)](https://github.com/grammyjs/i18n/issues/62)
- [ ] [Router](https://github.com/grammyjs/router/issues/7)
- [ ] [Emoji](https://github.com/grammyjs/emoji/issues/24)
- [ ] [Streaming Message Drafts](https://github.com/grammyjs/stream/issues/5)
- [ ] [Chat Members](https://github.com/grammyjs/chat-members/issues/15) (@roziscoding)
- [ ] [Commands](https://github.com/grammyjs/commands/issues/84) (@carafelix)
- [ ] [Format](https://github.com/grammyjs/format/issues/1)
- [x] Parse Mode
- [ ] Sessions

---

The following plugins are being deprecated instead of migrated:

- Internationalization (fluent)
[...]

PR #744, opened 2025-01-16
chore(ci): run CI on all branches and v2 pull requests

KnorpelSenf:
This lets us view CI when working on #675.

### DIFF
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -10,9 +10,7 @@ name: grammY
 
 on:
     push:
-        branches: [main]
-    pull_request:
-        branches: [main]
+        branches: ["*"]
 
 jobs:
     backport:


wojpawlik (approved):

.github/workflows/deno.yml
```diff
@@ -10,9 +10,7 @@ name: grammY
 
 on:
     push:
-        branches: [main]
-    pull_request:
-        branches: [main]
+        branches: ["*"]
```
EdJoPaTo, line 13: Removing `pull_request` means that external pull requests are no longer run against the CI. Its unlikely that this is wanted.

KnorpelSenf: Ohhhh I was not aware, thanks for pointing it out. I'll revert this part of the changes

KnorpelSenf: I'd expect that the rules collectively determine whether or not CI runs, rather than determining the number of times that identical jobs are started. Are you sure about this?

KnorpelSenf: Please see #749

KnorpelSenf: Holy crap, you were right. Wtf

EdJoPaTo: one runs the branch, one merges the branch into main and then runs the CI. When PR aren't on top of the main this might be different → different meaning → two jobs.

KnorpelSenf: @wojpawlik I think I have fixed it now, the trade-off is that jobs will only be duplicated when we merge the v2 branch via a PQ, which only happens once. Can you TAL?

wojpawlik: > one merges the branch into main and then runs the CI

Nope. Pushing to a PQ that wants to merge A into B triggers `push` on A and `pull_request` on B.

EdJoPaTo: I think we are talking past each other. You talk about what triggers them, I talk about what runs and why GitHub doesn’t deduplicate them.

`pull_request` merges with the target branch (in most cases main) before running the CI. (And doesn’t run when a merge can’t be done automatically.) `push` does not do merging. So they have different goals with different meaning.

Issue #697, opened 2024-12-15
feat: make `update` in webhook adapters a function


KnorpelSenf:
Using a function instead of a property for `update` lets use fix #613

shevernitskiy:
As far as `update` is `Promise<Update>`, we can await catch it in `webhookCallback` and gracefully stop Update processing. Like:

```ts
if (header !== secretToken) {
    await unauthorized();
    // TODO: investigate deno bug that happens when this console logging is removed
    console.log(handlerReturn);
    return handlerReturn;
}

const updateData = await update.catch(() => ({}));
if (!updateData.update_id) {
     await bad_request();
     console.log(handlerReturn);
     return handlerReturn;
}

let usedWebhookReply = false;
const webhookReplyEnvelope: WebhookReplyEnvelope = {
    async send(json) {
        usedWebhookReply = true;
        await respond(json);
    },
};
await timeoutIfNecessary(
    bot.handleUpdate(updateData, webhookReplyEnvelope),
    typeof onTimeout === "function"
        ? () => onTimeout(...args)
        : onTimeout,
    timeoutMilliseconds,
);
```

Answer with bad request feels right, but we need to introduce this extra callback in every framework handler then.

KnorpelSenf:
Is this related to the refactoring suggested in the issue description?

shevernitskiy:
> Is this related to the refactoring suggested in the issue description?

Yep, will do a PQ an 5-10mins

KnorpelSenf:
This was already done

Issue #613, opened 2024-07-14
Crashing on calling webhook with empty JSON body using Oak


t3hk0d3:
If somebody occasionally do an GET request to webhook endpoint it will result in entire server crashing.

```
error: Uncaught (in promise) BadRequestError: Unexpected end of JSON input
  return new errors[Status[status] as ErrorStatusKeys](message, options);
         ^
    at createHttpError (https://jsr.io/@oak/commons/0.11.0/http_errors.ts:211:10)
    at Body.json (https://deno.land/x/oak@v16.1.0/body.ts:152:15)
    at eventLoopTick (ext:core/01_core.js:168:7)
```

See also https://github.com/oakserver/oak/issues/661

KnorpelSenf:
The code responsible for oak is here: https://github.com/grammyjs/grammY/blob/fa10509af2d2daf7f3070024785bbc6d334eae17/src/convenience/frameworks.ts#L467-L481
Are you able to fix this?

KnorpelSenf:
Thank you!

rayz1065:
I noticed a crash when sending a POST request with an empty body to the webhook using Hono. I figured out it was due to an uncaught exception from a promise. As a test, changing the code as follows resolves the crash:

```typescript
update: (async () => {
  try {
    return await c.req.json();
  } catch (error) {
    return {} as any;
  }
})(),
```

Such a request would likely be refused due to the `header !== token` check (assuming the secret is set correctly) before reaching await update, but this can't happen since the update promise throws before that.
I believe this issue could be widespread, as we've observed it in two frameworks already.

A potential solution is changing the order in which webhookCallback handles promises:

```
const { update, respond, unauthorized, end, handlerReturn, header } =
  server(...args);
const receivedUpdate = await update;

// ...

await timeoutIfNecessary(
  bot.handleUpdate(receivedUpdate, webhookReplyEnvelope),
  typeof timeout === 'function' ? () => timeout(...args) : timeout,
  ms
);
```

This should fix the error for all frameworks.

KnorpelSenf:
This was fixed in #817 already, which defers the body reading

Issue #350, opened 2023-01-01
Investigate having to access `req.headers` on Deno to prevent requests from being closed


sparfenyuk:
Tried to play around this [example](https://grammy.dev/hosting/supabase.html#preparing-your-code). It basically doesn't work as expected. I am getting the following errors in Supabase dashboard:

```
TypeError: cannot read headers: request closed
    at Object.get headerList (deno:ext/fetch/23_request.js:116:19)
    at Request.request.<computed> (deno:ext/fetch/23_request.js:583:62)
    at Request.get [headers] (deno:ext/fetch/23_request.js:252:48)
    at Request.get headers (deno:ext/fetch/23_request.js:467:18)
    at stdHttp (https://deno.land/x/grammy@v1.13.0/convenience/frameworks.shared.ts:40:21)
    at https://deno.land/x/grammy@v1.13.0/convenience/webhook.ts:29:85
    at async Server.<anonymous> (file:///src/index.ts:14:16)
    at async Server.#respond (https://deno.land/std@0.170.0/http/server.ts:220:24)
```

```
TypeError: Cannot read properties of undefined (reading 'bodyUsed')
    at Server.#respond (https://deno.land/std@0.170.0/http/server.ts:221:26)
```

Note: My edge function works only if grammy is downgraded to 1.8.3.

KnorpelSenf:
Is 1.8.3 the latest version that works? In other words, does this issue already occur with 1.9.0?

KnorpelSenf:
Fascinating. Glad to see you found the root cause of the problem.

Do you think https://grammy.dev/hosting/supabase.html can be improved to include the relevant information?

> Actually, the similar workaround I found in the grammy code

Yes, it's still kinda annoying. I'm sure there must be a way to get rid of this. I'd be happy to hear about it if you find anything useful :)

KnorpelSenf:
This is one of these edge cases that @KnightNiwrem might enjoy spending time on

KnorpelSenf:
@sparfenyuk can you check if this still happens with the latest version of grammY? I assume so, but I'd just like to make sure this is still an issue before putting any time into it

gHashTag:
I have this problem when I run Grammy in Supabase Edge functions

```
[Error] TypeError: cannot read headers: request closed
    at get headerList (ext:deno_fetch/23_request.js:118:17)
    at request.<computed> (ext:deno_fetch/23_request.js:571:60)
[...]

PR #747, opened 2025-01-20
feat: fault tolerant body handling

shevernitskiy:
The issue was originally described here: https://github.com/grammyjs/grammY/issues/613

**Short Summary:**
We assume that the body will be valid JSON. However, it can be empty or contain invalid JSON in some unusual cases. This can cause the server to crash due to an uncaught error triggered by JSON parsing in some of the adapters.

It was suggested to wrap it in a function, but I think that is unnecessary. We already have a promise and just need to safely resolve it to a value that we can then validate.

I suggest catching all promises where a crash can occur and returning an empty object. In `webhookCallback`, we can perform basic validation by checking the `update_id` field. In the case of an invalid body, we should return a 400 status to the client. Here, we need to add an extra callback to the adapter interface.

I added some e2e test cases to check for 400 and 401 errors, but it takes some time (~300-350ms on my machine).

As of `TODO: investigate deno bug that happens when this console logging is removed` - I didn't notice any issues. Maybe it's fixed now?

Closes: https://github.com/grammyjs/grammY/issues/697

### DIFF
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -6,7 +6,7 @@
     },
     "tasks": {
         "check": "deno check --all src/mod.ts",
-        "test": "deno test --seed=123456 test/",
+        "test": "deno test --allow-net --seed=123456 test/",
         "ok": "deno fmt --check && deno lint && deno task test && deno task check",
         "fix": "deno fmt && deno lint --fix",
         // TODO: bring back coverage reports
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -6,7 +6,7 @@
     },
     "tasks": {
         "check": "deno check --all src/mod.ts",
-        "test": "deno test --allow-net --seed=123456 test/",
+        "test": "deno test --seed=123456 test/",
         "ok": "deno fmt --check && deno lint && deno task test && deno task check",
         "fix": "deno fmt && deno lint --fix",
         // TODO: bring back coverage reports
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -2,7 +2,8 @@ import type { Update } from "../types.ts";
 
 const SECRET_HEADER = "X-Telegram-Bot-Api-Secret-Token";
 const SECRET_HEADER_LOWERCASE = SECRET_HEADER.toLowerCase();
-const WRONG_TOKEN_ERROR = "secret token is wrong";
+export const WRONG_TOKEN_ERROR = "secret token is wrong";
+export const BAD_REQUEST_ERROR = "unable to parse request body";
 
 const ok = () => new Response(null, { status: 200 });
 const okJson = (json: string) =>
@@ -10,11 +11,9 @@ const okJson = (json: string) =>
         status: 200,
         headers: { "Content-Type": "application/json" },
     });
-const unauthorized = () =>
-    new Response('"unauthorized"', {
-        status: 401,
-        statusText: WRONG_TOKEN_ERROR,
-    });
+const unauthorized = () => new Response(WRONG_TOKEN_ERROR, { status: 401 });
+const badRequest = () => new Response(BAD_REQUEST_ERROR, { status: 400 });
+const empty = () => ({} as Update);
 
 /**
  * Abstraction over a request-response cycle, providing access to the update, as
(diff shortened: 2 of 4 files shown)


src/convenience/frameworks.ts
```diff
@@ -239,7 +243,7 @@ export type WorktopAdapter = (req: {
 
 /** AWS lambda serverless functions */
 const awsLambda: LambdaAdapter = (event, _context, callback) => ({
-    update: JSON.parse(event.body ?? "{}"),
+    update: Promise.resolve(JSON.parse(event.body ?? "{}")).catch(empty),
```
KnorpelSenf, line 246: What is your intent here? This looks a bit like you're trying to catch JSON parse errors, but this is not what is happening here

src/convenience/webhook.ts
```diff
@@ -183,17 +187,27 @@ function webhookCallback<C extends Context = Context>(
...
             initialized = true;
         }
         if (header !== secretToken) {
             await unauthorized();
-            // TODO: investigate deno bug that happens when this console logging is removed
-            console.log(handlerReturn);
+            return handlerReturn;
+        }
+        const updateData = await update;
+        if (!updateData?.update_id) {
```
KnorpelSenf, line 209: Is this supposed to check for `undefined` or both `undefined` and `0`?
```suggestion
        if (updateData?.update_id === undefined) {
```

src/convenience/frameworks.ts
shevernitskiy: i think you are right, my bad

src/convenience/webhook.ts
```diff
@@ -183,17 +187,27 @@ function webhookCallback<C extends Context = Context>(
...
+            header,
+        } = adapter(...args);
         if (!initialized) {
             // Will dedupe concurrently incoming calls from several updates
             await bot.init();
             initialized = true;
         }
         if (header !== secretToken) {
             await unauthorized();
-            // TODO: investigate deno bug that happens when this console logging is removed
```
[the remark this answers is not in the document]
shevernitskiy: I tested this case and didn't notice any issues. What goes wrong here?

src/convenience/webhook.ts
shevernitskiy: actually we can check against >== 0, update should be positive number

src/convenience/webhook.ts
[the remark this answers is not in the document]
KnorpelSenf: I believe it was something about the request being inaccessible. It might have been fixed in the runtime by now, see #350 for more info

deno.jsonc
```diff
@@ -6,7 +6,7 @@
     },
     "tasks": {
         "check": "deno check --all src/mod.ts",
-        "test": "deno test --seed=123456 test/",
+        "test": "deno test --allow-net --seed=123456 test/",
```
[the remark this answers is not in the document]
shevernitskiy: first time i check it with `hono`, but later i switch back to std to remove hono dep for test..
i can remove this tast cases

src/convenience/webhook.ts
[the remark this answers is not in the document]
shevernitskiy: yes, it seems fixed now

deno.jsonc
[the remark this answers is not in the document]
KnorpelSenf: That's not what I mean. We should definitely keep the tests. What I meant is that they can be written without `Deno.serve` and `fetch`, simply by making one function call the other.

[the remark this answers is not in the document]
shevernitskiy: I understand, but i was checking some `.json()` method provided by framework itself (we still can construct ctx by hand i think)
Sure, in case of `std` we can just test adapter without server, i'll do it:)

KnorpelSenf (approved):

KnorpelSenf:
> Closes: #697

This is not true, it's still a property on the adapters rather than a function

src/convenience/frameworks.ts
```diff
@@ -239,6 +243,7 @@ export type WorktopAdapter = (req: {
 
 /** AWS lambda serverless functions */
 const awsLambda: LambdaAdapter = (event, _context, callback) => ({
+    // TODO: add safe parse workaround
```
KnorpelSenf, line 246: Making it a function will btw fix this

shevernitskiy: i just commited, but again late for 2 secs before you merged:)
```ts
function safeJsonParse(value: string, fallback = {} as any): any {
    try {
        return JSON.parse(value);
    } catch (_err) {
        return fallback;
    }
}
```

KnorpelSenf: Ah

shevernitskiy:
> > Closes: #697
> 
> This is not true, it's still a property on the adapters rather than a function

but function wrapper is not needed now

PR #749, opened 2025-01-23
chore(ci): run on PQs from forks

KnorpelSenf:
Regression introduced in #744

### DIFF
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -12,7 +12,6 @@ on:
     push:
         branches: ["*"]
     pull_request:
-        branches: ["*"]
 
 jobs:
     backport:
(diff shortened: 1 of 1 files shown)


.github/workflows/deno.yml
```diff
@@ -11,6 +11,8 @@ name: grammY
 on:
     push:
         branches: ["*"]
+    pull_request:
+        branches: ["*"]
```
EdJoPaTo, line 15: This line can be omitted as its (as far as I recall) the same

```suggestion
        branches: ["*"]
```

EdJoPaTo (dismissed):

KnorpelSenf:
We now no longer run all jobs twice

wojpawlik (approved):

PR #750, opened 2025-01-27
fix: add getAvailableGifts as zero param method

KnightNiwrem:
https://core.telegram.org/bots/api#getavailablegifts


KnorpelSenf (approved):
LGTM

PR #751, opened 2025-01-28
refactor: introduce new abstraction for webhook adapters

KnorpelSenf:
This just adds a demo. Take a look at the second commit. Does not compile. Do not merge unless this is finished up.


KnorpelSenf:
I think this was already done on the v2 branch.

PR #752, opened 2025-01-30
feat(convenience): elysia webhook adapter

Lutymane:
Hello. Here's an adapter for Elysia web server.

Also to easier type hook arguments I added elysia as optional peer dependency, so that could perhaps be utilized for other adapters to make the code more compact and up-to-date

### DIFF
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -109,6 +109,15 @@ export type CloudflareModuleAdapter = (
     request: Request,
 ) => ReqResHandler<Response>;
 
+export type ElysiaAdapter = (ctx: {
+    body: Update;
+    headers: Record<string, string | undefined>;
+    set: {
+        headers: Record<string, string>;
+        status: number;
+    };
+}) => ReqResHandler<string>;
+
 export type ExpressAdapter = (req: {
     body: Update;
     header: (header: string) => string | undefined;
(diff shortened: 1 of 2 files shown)


KnorpelSenf:
> Hello. Here's an adapter for Elysia web server.

Thanks!

> Also to easier type hook arguments I added elysia as optional peer dependency, so that could perhaps be utilized for other adapters to make the code more compact and up-to-date

We do not want to add dozens of peer dependencies. Many adapters work for serveral libraries, several environements, and several versions of the same library. Using exact types actually make compatibility worse. Also, importing third-party library types in the tests is enough to ensure that our minimal abstractions of web frameworks stay up to date.

src/convenience/frameworks.ts
```diff
@@ -539,6 +539,33 @@ const worktop: WorktopAdapter = (req, res) => ({
     unauthorized: () => res.send(401, WRONG_TOKEN_ERROR),
 });
 
+const elysia = (ctx: import("elysia").Context) => {
```
KnorpelSenf, line 542: Can you refactor this to be consistent with the other adapters? In other words, this should use a minimal abstraction of the elysia types

KnorpelSenf (changes requested):

src/convenience/frameworks.ts
Lutymane: I marked that peer dependency as optional, so it won't be installed by default during grammy development, but still would get proper type checks for the bot's code (which can be beneficial if server types happen to change or next version comes out, so it ensures compatibility, just gotta tweak the semver better). But if you're sure it still should be refactored with minimal abstraction instead, I can do it, sure

KnorpelSenf: I'm aware that the dep usually isn't installed assuming that people don't mess up, but using the types from the dep isn't reliable enough. They can be missing, they can be too narrow, they're always harder to debug. So yes I'm sure. Thanks!

Lutymane:
Should be fine now, please check

Lutymane:
Oh lemme format the code and sign the commit, sorry

KnorpelSenf:
Almost there

Lutymane:
I hope this is finally it xd. Repo should probably have pre-commit linting hook setup

KnorpelSenf:
I don't want/need it myself but you're free to create such a hook on your machine

KnorpelSenf (approved):
LGTM

KnorpelSenf:
There's going to be a release in the coming days which will contain your changes

KnorpelSenf:
@all-contributors add @Lutymane for this code!

PR #755, opened 2025-01-31
refactor: remove `ignore` as onTimeout option

shevernitskiy:
remove `ignore` as onTimeout option and refactor timeout executor a bit

closes https://github.com/grammyjs/grammY/issues/696

### DIFF
--- a/src/convenience/webhook.ts
+++ b/src/convenience/webhook.ts
@@ -5,27 +5,31 @@ import type { Context } from "../context.ts";
 import type { WebhookReplyEnvelope } from "../core/client.ts";
 import type { Update } from "../types.ts";
 import {
-    adapters as nativeAdapters,
-    BAD_REQUEST_ERROR,
+    adapters,
+    // BAD_REQUEST_ERROR,
     type FrameworkAdapter,
-    WRONG_TOKEN_ERROR,
+    // WRONG_TOKEN_ERROR,
 } from "./frameworks.ts";
+
+import { HTTPMethods } from "npm:fastify";
 const debugErr = createDebug("grammy:error");
 
-const callbackAdapter: FrameworkAdapter = (
-    update: Update,
-    callback: (json: string) => unknown,
-    header: string,
-    unauthorized = () => callback(WRONG_TOKEN_ERROR),
-    badRequest = () => callback(BAD_REQUEST_ERROR),
-) => ({
-    update: Promise.resolve(update),
-    respond: callback,
-    header,
-    unauthorized,
-    badRequest,
-});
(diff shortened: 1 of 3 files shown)


src/convenience/webhook.ts
```diff
@@ -252,27 +252,30 @@ function webhookCallback<
 
 function timeoutIfNecessary(
     task: Promise<void>,
-    onTimeout: "throw" | "ignore" | (() => unknown),
+    onTimeout: "throw" | (() => unknown), // need async?
```
KnorpelSenf, line 255: I think we can change this to be an optional function now, it will either call the function or throw if none is specified

src/convenience/webhook.ts
```diff
@@ -237,10 +237,8 @@ function webhookCallback<
         };
         await timeoutIfNecessary(
             bot.handleUpdate(updateData, webhookReplyEnvelope),
-            typeof onTimeout === "function"
-                ? () => onTimeout(...args)
-                : onTimeout,
             timeoutMilliseconds,
+            onTimeout ? () => onTimeout(...args) : onTimeout,
```
KnorpelSenf, line 241:
```suggestion
            onTimeout,
```

src/convenience/webhook.ts
```diff
@@ -252,27 +250,30 @@ function webhookCallback<
...
-        }, timeout);
-        task.then(resolve)
-            .catch(reject)
-            .finally(() => clearTimeout(handle));
-    });
+    const { promise, resolve, reject } = Promise.withResolvers<void>();
+
+    const handle = setTimeout(() => {
+        debugErr(`Request timed out after ${timeout} ms`);
+        if (onTimeout) {
```
KnorpelSenf, line 261:
```suggestion
        if (onTimeout !== undefined) {
```
just to be more explicit

src/convenience/webhook.ts
shevernitskiy: but args?

KnorpelSenf: Right, but then
```suggestion
            onTimeout !== undefined ? () => onTimeout(...args) : undefined,
```

shevernitskiy:
i think we should make `onTimeoute` awaitable
async user callback should not leaks out of the scope of execution

KnorpelSenf:
Yep, sounds good, please go ahead

shevernitskiy:
> Yep, sounds good, please go ahead

i think the all to close original issue:) isn't it?

KnorpelSenf (approved):
Oh, I didn't see that commit. Awesome! LGTM

PR #757, opened 2025-02-12
feat: support Bot API 8.3

KnorpelSenf:
https://core.telegram.org/bots/api#february-12-2025

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -2857,7 +2857,7 @@ export class Context implements RenamedUpdate {
         other?: Other<"sendGift", "user_id" | "chat_id" | "gift_id">,
         signal?: AbortSignal,
     ) {
-        return this.api.sendGift(
+        return this.api.sendGiftToChannel(
             orThrow(this.chat, "sendGift").id,
             gift_id,
             other,
--- a/src/context.ts
+++ b/src/context.ts
@@ -1538,7 +1538,7 @@ export class Context implements RenamedUpdate {
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/context.ts
```diff
@@ -2843,6 +2843,28 @@ export class Context implements RenamedUpdate {
...
+     * @param signal Optional `AbortSignal` to cancel the request
+     *
+     * **Official reference:** https://core.telegram.org/bots/api#sendgift
+     */
+    replyWithGiftToChannel(
+        gift_id: string,
+        other?: Other<"sendGift", "user_id" | "chat_id" | "gift_id">,
+        signal?: AbortSignal,
+    ) {
+        return this.api.sendGift(
```
KnorpelSenf, line 2860: @KnightNiwrem excellent catch
```suggestion
        return this.api.sendGiftToChannel(
```

src/core/api.ts
```diff
@@ -789,7 +789,7 @@ export class Api<R extends RawApi = RawApi> {
     }
 
     /**
-     * Use this method to change the chosen reactions on a message. Service messages can't be reacted to. Automatically forwarded messages from a channel to its discussion group have the same available reactions as messages in the channel. Bots can't use paid reactions. Returns True on success.
+     * Use this method to change the chosen reactions on a message. Service messages of some types can't be reacted to. Automatically forwarded messages from a channel to its discussion group have the same available reactions as messages in the channel. Bots can't use paid reactions. Returns True on success.
```
carafelix, line 792: > Service messages of some types can't be reacted to.

Which ones? Could you elaborate?

KnorpelSenf: No.

(Also, this isn't an explanation, it's simply a copy of the text in the API reference. Even if we knew, we couldn't change this line.)

PR #759, opened 2025-02-13
Distribute chat type in Context.hasChatType

### LINKED ISSUE #619: Improve inference when using ctx.hasChatType with multiple chat types
Suppose we have a set of chat types we want to listen for within a middleware, and a handler we need to call on match:
```typescript
const chatTypes = [
  'channel',
  'group',
  'supergroup',
] as const satisfies Chat['type'][];

const handler: <C extends Context>(
  ctx: ChatTypeContext<C, (typeof chatTypes)[number]>
) => Promise<unknown> = async (ctx) => {
  // do something
};
```
I expect both of the following pieces of code to work properly, but while the first one works fine, the second one shows an error:
```typescript
const composer = new Composer<Context>();

composer.chatType(chatTypes).use(async (ctx, next) => {
  await handler(ctx); // ok
  return next();
});

composer.use(async (ctx, next) => {
  if (ctx.hasChatType(chatTypes)) {
    await handler(ctx); // error
  }
  return next();
});
```

As mentioned, the error only occurs when more than one chat is used for filtering.
```
Argument of type 'Context & Record<"update", ChatTypeUpdate<"channel" | "group" | "supergroup">> & ChatType<"channel" | "group" | "supergroup"> & Record<...> & ChatFrom<...> & Partial<...> & AliasProps<...>' is not assignable to parameter of type 'ChatTypeContext<Context & Record<"update", ChatTypeUpdate<"channel" | "group" | "supergroup">> & ChatType<"channel" | "group" | "supergroup"> & ChatFrom<...> & Partial<...> & AliasProps<...>, "channel" | ... 1 more ... | "supergroup">'.
// ...
      Types of property 'update' are incompatible.
// ...
                    Types of property 'chat' are incompatible.
                      Type '(GroupChat & { type: "channel" | "group" | "supergroup"; }) | (SupergroupChat & { type: "channel" | "group" | "supergroup"; })' is not assignable to type '{ type: "supergroup"; }'.
                        Type 'GroupChat & { type: "channel" | "group" | "supergroup"; }' is not assignable to type '{ type: "supergroup"; }'.
                          Types of property 'type' are incompatible.
                            Type '"group"' is not assignable to type '"supergroup"'.ts(2345)
```
It could be useful to investigate why the two pieces of code are behaving in a different way, as the second version has some use-cases not entirely covered by the first one.

rayz1065:
Closes #619

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -877,7 +877,7 @@ export class Context implements RenamedUpdate {
      */
     hasChatType<T extends Chat["type"]>(
         chatType: MaybeArray<T>,
-    ): this is ChatTypeContextCore<T> {
+    ): this is T extends unknown ? ChatTypeContextCore<T> : never {
         return Context.has.chatType(chatType)(this);
     }
     /**


KnorpelSenf (approved):
LGTM but why doesn't this problem appear with the other types?

rayz1065:
> LGTM but why doesn't this problem appear with the other types?

I'd say it's because this error could only happen for generic `has` filters, which right now are only `ChatTypeContext` and `Filter`.
There isn't any issue with `FilterCore` since it already distributes over the union, so the issue was only with `ChatTypeContextCore` which did not.

KnorpelSenf:
Oh. Right. Sorry, I was a bit slow. In that case, we should not merge this. Instead, we should rather do something like #760, no?

KnorpelSenf:
Are you asking if there is a difference for this particular type in the context of grammY, or if there is a difference between the two type definitions at all?

KnorpelSenf:
> Sure, that's nicer.

Could you leave a review?

rayz1065:
> Are you asking if there is a difference for this particular type in the context of grammY, or if there is a difference between the two type definitions at all?

I was mainly curious about why the type assertion was working, as I'm not sure if the types are equivalent in general.

KnorpelSenf:
I see. The types are not identical. In some situations, they work equivalently, such as with your type tests. This is rather a consequence of your type tests. They do not work correctly for non-function types.

As for your type test, it checks if `<T>() => T` extends your context type helpers. However, those helpers are not functions, so the type will evaluate to false (`2`). This happens for both definitions, so both yield functions that return `2`, so both context types are considered equal.

Check what `{Q,R}{1,2}` are inferred as to see more.

```ts
type P1<X> = (<T>() => T extends X ? 1 : 2) extends (() => 1)
    ? "function that returns 1"
    : "not a function that returns 1";
type P2<X> = (<T>() => T extends X ? 1 : 2) extends (() => 2)
    ? "function that returns 2"
    : "not a function that returns 2";
type Q1 = P1<GroupSupergroup>;
type Q2 = P2<GroupSupergroup>;
type R1 = P1<GroupSupergroup2>;
type R2 = P2<GroupSupergroup2>;
```

PR #760, opened 2025-02-14
fix: distribute chat type context types

### LINKED ISSUE #619: Improve inference when using ctx.hasChatType with multiple chat types

KnorpelSenf:
Closes #619
Closes #759 by superseding it

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -3301,13 +3301,14 @@ export type ShippingQueryContext<C extends Context> = Filter<
     "shipping_query"
 >;
 
-type ChatTypeContextCore<T extends Chat["type"]> =
-    & Record<"update", ChatTypeUpdate<T>> // ctx.update
-    & ChatType<T> // ctx.chat
-    & Record<"chatId", number> // ctx.chatId
-    & ChatFrom<T> // ctx.from
-    & ChatTypeRecord<"msg", T> // ctx.msg
-    & AliasProps<ChatTypeUpdate<T>>; // ctx.message etc
+type ChatTypeContextCore<T extends Chat["type"]> = T extends unknown ?
+        & Record<"update", ChatTypeUpdate<T>> // ctx.update
+        & ChatType<T> // ctx.chat
+        & Record<"chatId", number> // ctx.chatId
+        & ChatFrom<T> // ctx.from
+        & ChatTypeRecord<"msg", T> // ctx.msg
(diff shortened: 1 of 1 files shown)


rayz1065 (approved):
LGTM

PR #763, opened 2025-02-21
feat: add `Composer::boundary()`

KnorpelSenf:
As suggested by @rayz1065


rayz1065:
Suppose I want to write a tree surrounding plugin, with regular middleware plugins I would simply be able to write:
```ts
type Options = { foo: string };
function myPlugin<C extends Context>(options?: Options): Middleware<C> {
    return (ctx, next) => {
        // ...
    };
}
bot.use(myPlugin({ foo: "bar" }));
```
It would be nice to be able to do something similar for a surrounding middleware:
```ts
export type SurroundingMiddlewareFn<C extends Context = Context> = (
    ctx: C,
    tree: () => Promise<{ nextCalled: boolean }>,
    next: NextFunction,
) => MaybePromise<unknown>;

function mySurroundingPlugin<C extends Context>(
    options?: Options,
): SurroundingMiddlewareFn<C> {
    return (ctx, tree, next) => {
        // ...
    };
}
const surrounded = bot.surround(mySurroundingPlugin({ foo: "baz" }));
```

KnightNiwrem (approved):

rayz1065 (approved):

src/composer.ts
```diff
@@ -908,24 +944,65 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
...
+ *
+ * ```ts
+ * // wrap a subtree of middleware with custom logic
+ * const sub = bot.surround(
+ *   // pass a boundary controller, wrapping the subtree
+ *   async (ctx, subtree) => {
+ *     // A
+ *     await subtree() // -- returns false because C didn't call next()
+ *     // E
+ *   },
```
rayz1065, line 983: Should there be a `return` in this doc to conform to the expected type?
```suggestion
 *   async (ctx, subtree) => {
 *     // A
 *     const nextCalled = await subtree() // -- returns false because C didn't call next()
 *     // E
 *     return nextCalled
 *   },
```

rayz1065 (approved):

KnorpelSenf:
This is a mess now. There are several competing approaches on the same branch. The changes did not get strictly better and it's hard to compare them. We also have not found a flawless abstraction, nor have agreed upon one that does the right tradeoffs.

I will close this now. If we want to land this feature, we should reach consensus first and then perform these changes on a new branch, or cherry-pick them from here.

PR #765, opened 2025-02-27
perf: reduce GC churn in `Composer::filter`

wojpawlik:
Each `Composer::lazy` creates a temporary `Composer` for each update.

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -692,7 +692,13 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
         ...middleware: Array<Middleware<C>>
     ) {
         const composer = new Composer(...middleware);
-        this.branch(predicate, composer, pass);
+        this.use(async (ctx, next) => {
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
Can you provide any data that indicates the GC impact of these changes?

wojpawlik (commented):
No. I don't have any bot with enough listeners, nor varied fake updates to supply to it.
I see no way in which it could be worse than the original.

wojpawlik:
Alternatively, the function passed to `.lazy()` could be restricted to return a single  `Middleware<C>`. Users could still explicitly construct and return a temporary `Composer`.

KnorpelSenf:
Attempts to improve the performance without actually providing any evidence are arbitrary and there is no reason to merge them. In particular, it makes no sense to introduce breaking changes that require other plugins to be reworked.

Feel free to re-open if you find a good reason for these changes. Otherwise I'll close this.

PR #767, opened 2025-03-01
feat: add ctx.sessionKey

roziscoding:
- **feat: add ctx.sessionKey**


roziscoding:
I have no idea why type checks are failing but tests pass when ran with `--no-check`

KnorpelSenf:
> I have no idea why type checks are failing

Me neither, but I think they're gonna be fixed after I complete my work on the new types

roziscoding:
I moved this to [grammyjs/sessions](https://github.com/grammyjs/sessions/pull/1) since the plugin will no longer exist on the core lib

PR #768, opened 2025-03-02
fix: reduce `NotUndefined` to `{}`

KnorpelSenf:
The filter queries were still written with TS pre-3.5 in mind. `{}` actually means “anything except null or undefined” nowadays (for a long time already) so we should be able to rely on this in order to remove undefined.

Reported in https://t.me/grammyjs/296604

### DIFF
--- a/src/filter.ts
+++ b/src/filter.ts
@@ -452,7 +452,8 @@ export type FilterQuery = ComputeFilterQueryList;
  * Any kind of value that appears in the Telegram Bot API. When intersected with
  * an optional field, it effectively removes `| undefined`.
  */
-type NotUndefined = string | number | boolean | object;
+// deno-lint-ignore ban-types
+type NotUndefined = {};
 
 /**
  * Given a FilterQuery, returns an object that, when intersected with an Update,


src/filter.ts
```diff
@@ -452,7 +452,8 @@ export type FilterQuery = ComputeFilterQueryList;
  * Any kind of value that appears in the Telegram Bot API. When intersected with
  * an optional field, it effectively removes `| undefined`.
  */
-type NotUndefined = string | number | boolean | object;
+// deno-lint-ignore ban-types
+type NotUndefined = {};
```
rayz1065, line 456:
```suggestion
type NotUndefined = Record<never, never>;
```

rayz1065: Should be the exact same, but doesn't need to ignore ban-types

KnorpelSenf: That means empty object and doesn't permit primitive values, does it?

rayz1065: It does allow them
```
type NotUndefined = Record<never, never>;
const test1: NotUndefined = 1;
const test2: NotUndefined = null; // error
const test3: NotUndefined = undefined; // error
```
That said, I got this info from eslint which used to advise to replace {} with Record<never, never>, it seems to now suggest to simply disable the rule.

KnorpelSenf: Hmmmm semantically we still mean `{}` but now I'm wondering what the difference between `{}` and `Record<never, never>` is

rayz1065: I checked, seems like `Record<never, never>` (or `Record<never, unknown>` for the matter) simply resolves to `{}` (without the linter complaining)

KnorpelSenf: Very strange, this feels like a bug in the linter because anything that reads `Record` is the most in your face version of an object type I can imagine

KnorpelSenf: So if that's not an object then I don't know what else to lint

wojpawlik: Intersections with `Record<never, never>` break several type tests.

wojpawlik (approved):

PR #769, opened 2025-03-03
feat: allow session key to be undefined

### LINKED ISSUE #713: feat: let `ctx.session` be undefined for missing sessions
The fact that an error is thrown upon property access is really annoying. There is no good way of knowing whether session data exists, so we should changes this behaviour.

roziscoding:
Closes #713

### DIFF
--- a/src/convenience/session.ts
+++ b/src/convenience/session.ts
@@ -384,8 +384,7 @@ class PropertySession<O extends {}, P extends keyof O> {
             enumerable: true,
             get: () => {
                 if (key === undefined) {
-                    const msg = undef("access", opts);
-                    throw new Error(msg);
+                    return undefined;
                 }
                 this.read = true;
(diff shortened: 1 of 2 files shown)


src/convenience/session.ts
```diff
@@ -384,8 +384,7 @@ class PropertySession<O extends {}, P extends keyof O> {
             enumerable: true,
             get: () => {
                 if (key === undefined) {
-                    const msg = undef("access", opts);
-                    throw new Error(msg);
+                    return undefined;
```
KnorpelSenf, line 387: This now causes a type mismatch, see https://github.com/grammyjs/grammY/issues/713#issuecomment-2694864398

roziscoding:
I moved this to [grammyjs/sessions](https://github.com/grammyjs/sessions/pull/2) since the plugin will no longer exist on the core lib

PR #783, opened 2025-04-08
feat: add filter query for link_preview_options

SpEcHiDe:
Based on [the message_thread](https://grammyjs.t.me/300334)

Please let me know if this is expected approach.

### DIFF
--- a/src/filter.ts
+++ b/src/filter.ts
@@ -289,7 +289,6 @@ const COMMON_MESSAGE_KEYS = {
     caption: {},
 
     link_preview_options: {
-        is_disabled: {},
         url: {},
         prefer_small_media: {},
         prefer_large_media: {},
--- a/src/filter.ts
+++ b/src/filter.ts
@@ -288,6 +288,7 @@ const COMMON_MESSAGE_KEYS = {
     caption_entities: ENTITY_KEYS,
     caption: {},
 
+    link_preview_options: {},
     effect_id: {},
     has_media_spoiler: {},
 
--- a/src/filter.ts
+++ b/src/filter.ts
@@ -288,7 +288,13 @@ const COMMON_MESSAGE_KEYS = {
     caption_entities: ENTITY_KEYS,
     caption: {},
 
-    link_preview_options: {},
+    link_preview_options: {
+        is_disabled: {},
+        url: {},
+        prefer_small_media: {},
+        prefer_large_media: {},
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
Nice, did you test this locally? Does it work as expected?

SpEcHiDe:
steps i did to test locally:

```
git clone https://github.com/spechide/grammY
git checkout patch-1
npm install
npm run backport
deno task bundle-web
```

screenshot


I was not able to test the `link_preview_options:is_disabled` scenario. Apart from that I was not able to find any issues in the above code.

but by testing below code, I found one ~~issue~~,

```ts
const privateComposer = composer.chatType("private");

privateComposer.on("::url", async ctx => {
    console.log(ctx.update);
    
    const urls = ctx.entities(["url"]);
    console.log(urls);  // 
});
```
by testing with the below message.

screenshot

The `ctx.entities` would not be able to obtain the url yet. Should it be able to do it, or not? 🤔

Can you let me know what are the expectations?

KnorpelSenf:
The filter query `::url` and `ctx.entities` both operate on the message entities. They are not related to the link preview in any way, so it is correct that they do not appear there.

KnorpelSenf (approved):
LGTM.

What needs to be done to move this out of draft state and get this ready to be merged?

SpEcHiDe:
> The filter query `::url` and `ctx.entities` both operate on the message entities. They are not related to the link preview in any way, so it is correct that they do not appear there.

Ok, 

What about the `::is_disabled` thing?

KnorpelSenf:
That's a great catch. I think it makes no sense to filter for `is_disabled` actually. It is already possible to filter for
```ts
bot.on("::url").drop(Context.has.filterQuery(":link_preview_options"))
```
which effectively only gives you messages URL entities but without link preview.

src/filter.ts
```diff
@@ -288,6 +288,13 @@ const COMMON_MESSAGE_KEYS = {
     caption_entities: ENTITY_KEYS,
     caption: {},
 
+    link_preview_options: {
+        is_disabled: {},
```
KnorpelSenf, line 292: I think it is reasonable to remove it.

```suggestion
```

If you want, we can ask the TG devs about it first.

KnorpelSenf:
@all-contributors add @SpEcHiDe for idea and code and test

PR #787, opened 2025-04-13
feat: support Bot Api 9.0

KnorpelSenf:
https://core.telegram.org/bots/api#april-11-2025

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 <!-- deno-fmt-ignore-start -->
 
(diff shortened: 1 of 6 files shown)


rojvv (approved):
LGTM. One doubt tho. Would you add context shortcuts for the newly introduced methods, for example, `readBusinessMessage`? If not, why?

KnorpelSenf:
Nice catch, I'll fix it in a few hours

PR #798, opened 2025-05-04
Update README.md CommonJS -> ESM

pvlvld:
Change CommonJS (require) to ESM (import) syntax in an example.
Newbies don't understand why they don't have types using that example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ npm install grammy
 inside it. Then create a file `bot.js` with this content:
 
 ```ts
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
This isn't possible without adding extra instructions which
- explain how to set up ESM support
- probably justify why this extra step is necessary, which means explaining what CJS and ESM even are

Why doesn't your TS setup support CJS?

KnorpelSenf:
oh?

PR #802, opened 2025-05-25
build: prefer jsr.io/@std over deno.land/std

KnorpelSenf:
The old std no longer compiles. We need to update the version, and thus change the registry.

This only affects Deno-specific code. The Node build output will be byte-identical.


rojvv (commented):
Why not `jsr:@std/path@1.0.9/basename`?

rojvv (approved):

PR #810, opened 2025-07-08
Add "addIf" to keyboards

voidboost:


KnorpelSenf:
Can you explain why you need this? I.e. what does code with and without this feature look like in your use case?

voidboost:
> Can you explain why you need this? I.e. what does code with and without this feature look like in your use case?

To simplify building paginated views in the Telegram interface.

KnorpelSenf:
That's not nearly enough info for me to make sense of this

voidboost:
> That's not nearly enough info for me to make sense of this

KnorpelSenf:
This can already be done imperatively by moving the condition around the button declaration.

```ts
const keyboard = new InlineKeyboard()
if (cond0) keyboard.text(..)
if (cond1) keyboard.text(..)
keyboard.row()
```

However, I see the point that you'd then have to introduce a new variable, which may be undesired. So if that's why you've opened this pull request, we can solve that problem. 

Telegraf has historically had an option `hide` which permits this:

```ts
keyboard.text("top", "page0", { hide: page === 0 })
```

That way, you don't have to call two methods (one to create the button, one to add it). This options object could be added to all methods, both static and non-static ones. Have you considered this?

KnorpelSenf:
What happened? Are you no longer interested in improving the situation?

voidboost:
> What happened? Are you no longer interested in improving the situation?

This pull request was created to be helpful to others — I’ve already implemented a solution for myself.

KnorpelSenf:
I see, I somehow assumed you actually wanted this to get merged. Nevermind then

Issue #352, opened 2023-01-04
feat: integrate the auto-thread plugin into the core lib


EdJoPaTo:
`ctx.reply()` (and similar reply methods) do not reply to the current forum thread. They should as their use is to be a reply to whatever the current context is.

Not sure if this is a breaking change or not as it personally sounds like a bugfix to me.

Workaround until fixed:

```ts
await ctx.reply("something", {
  message_thread_id: 314,
});
```

KnorpelSenf:
This can already be done using https://github.com/grammyjs/auto-thread

EdJoPaTo:
Personally I still think this is something the reply method is supposed to do on its own. The idea of reply is to reply to whatever the update was about. If that's in some thread then reply in some thread. If the reply method is not handling this its a bug to me as the idea of reply is violated.

Integrating some plugin sounds strange to me as its manipulating via a middleware and not natively correct behaviour which would be more efficient. Maybe that is what was meant to be done but the title of this issue should still be something like "reply doesnt work correctly with threads → fix that"

roziscoding:
I agree that this should be the default behavior, but changing this now would lead to a change in behavior of existing bots and would require people who like the current behavior to refractor their code to use ctx.api.sendMessage, or ctx.sendMessage, if we add it.

Also, it'd require us to change the bahavior of all the `replyWith...` too, for consistency, which would require us to have `ctx.send...` methods for people who like the current behavior.

If we were to integrate this behavior into core, I'd do it being a config flag, and then we make it default to true on 2.0.

KnorpelSenf:
The current behaviour can be achieved via `ctx.api` methods. This is preferable over configuration.

Is it correct to leave out the parameter? It used not to be, but perhaps leaving it out will now send messages to the general topic?

I see two possible ways to proceed with this:
1. Wait until 2.0.
2. Label this a bug because it's unexpected behaviour, so we're allowed to fix it immediately without breaking semver.

roziscoding:
> The current behaviour can be achieved via `ctx.api` methods. This is preferable over configuration.

Yes, which means that, if we do point 2, we'll change the behavior of people's code without changing the major, and will require people to refractor they're codebase to stay up to date, which should only really be expected on major releases

I understand that asking people to refactor is preferred over configuration, but that would break SemVer. If we're OK with breaking SemVer because we assume everybody expects `reply` to work this way, then fine. Otherwise, I'd suggest we go another route.

[...]

PR #815, opened 2025-08-12
fix: reply to same topic by default

KnorpelSenf:
When grammY shipped support for forum chats, the default behaviour of `ctx.reply` was to reply to the relevant _chat_ rather than the relevant topic. The rationale behind this was that we should not be overly assertive about which topic is relevant to a reply. Developers would have to specify `{ message_thread_id: ctx.msg?.message_thread_id }` manually in the options if they wanted to reply to the same topic the incoming message was sent to.

This quickly turned out to be counterintuitive for most people (#352).

So far, grammY has not fixed this because it seemed less like a bug and more like a bad default. The idea was to fix this for 2.0.

However, it regularly happens that people find this behaviour so unexpected that they report it as a bug. In addition, we have yet to find a bot that actually wants to do anything else but reply to the same topic. Perhaps we need to admit that this is actually rather a bug (that was introduced intentionally) rather than bad API design. As a consequence, fixing it can be done in a patch release. We do not have to wait for 2.0.

These changes make sure that all context actions send their messages directly to the same topic.

```ts
await ctx.reply("same topic")
await ctx.reply("general topic (restores previous behaviour)", {
  message_thread_id: undefined
})
```


samczsun:
I can think of very few use cases where someone would deliberately call the function called "reply" in response to a message but not want it to display as a reply to the message (at least, not in topic view. maybe it shows as a proper reply in message view). My bots all expect the behavior provided by auto-thread

MKRhere (approved):

KnightNiwrem:
@coderabbitai full review

KnightNiwrem:
> @coderabbitai full review

Jk

PR #816, opened 2025-08-12
fix: support `request.json` throwing errors for bun adapter

KnorpelSenf:
Bun apparently may throw an error rather than rejecting if it can determine early that the JSON cannot be parsed. This catches that.

### DIFF
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -318,7 +318,13 @@ const azureV4: AzureAdapterV4 = (request) => {
 const bun: BunAdapter = (request) => {
     let resolveResponse: (response: Response) => void;
     return {
-        update: Promise.try(() => request.json()),
+        update: new Promise<Update>((resolve, reject) => {
+            try {
+                resolve(request.json());
+            } catch (e) {
+                reject(e);
+            }
+        }),
         header: request.headers.get(SECRET_HEADER) || undefined,
         end: () => {
             resolveResponse(ok());


MKRhere:
Thanks for quick PR -- but I realised this is a more foundational problem in the way grammY handles promises.

If you don't synchronously attach a catch to a Promise, it will throw an unhandled rejection when it rejects.

```TS
import { setTimeout } from "node:timers/promises";

let res = Promise.reject(new Error("test"));

await setTimeout(2000);

res.catch(() => console.log("Error caught"));
```

Node, Deno, and Bun will throw an unhandledRejection for this code and crash the process. This is essentially the simplified flow of what grammY is doing:

```TS
const update = request.json();

await bot.init();

await unauthorized();

await update;
```

Both the awaits are conditional, but the second one is hit when header !== token. When a bad request comes in, it crashes the bot.

For all framework adapters where `request.json()` or similar promise-returning function is a thing, this will throw an unhandled rejection. Making `ReqResHandler::update` a method that returns `request.json()` seems like the easy solution to me. To avoid a breaking change, make it a union: `Promise<update> | () => Promise<update>`, and when you call the function, immediately await it, so error handling will happen correctly.

I recommend changing the PR title to reflect this as well.

MKRhere:
It makes more sense when you think of it like this. When this block is hit:

```TS
if (header !== token) {
	await unauthorized();
	return handlerReturn;
}
```

The update promise has nowhere to go to be handled. So it makes no sense to eagerly create the promise and throw it away.

KnorpelSenf:
Excellent findings. Closing.

PR #817, opened 2025-08-12
fix: unhandled rejections from lazy promise handling

MKRhere:
More explanation here: https://github.com/grammyjs/grammY/pull/816#issuecomment-3180716466

We turned `ReqResHandler::update` into a getter to allow handling update promises lazily without breaking API.

### DIFF
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -52,7 +52,7 @@ export interface ReqResHandler<T = void> {
      * Some frameworks (e.g. Deno's std/http `listenAndServe`) assume that
      * handler returns something
      */
-    handlerReturn?: MaybePromise<T>;
+    handlerReturn?: Promise<T>;
 }
 
 /**
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -280,11 +280,7 @@ const awsLambdaAsync: LambdaAsyncAdapter = (event, _context) => {
                 body: json,
             }),
         unauthorized: () => resolveResponse({ statusCode: 401 }),
-        get handlerReturn() {
-            return new Promise<void>((resolve) => {
-                resolveResponse = resolve;
-            });
-        },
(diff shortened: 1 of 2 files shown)


MKRhere:
Attempted `handlerReturn` change was a fumble, but at least it looks consistent now. Tested this build locally, and can confirm it works as expected, and errors can now be caught.

And now we won't start reading request body if authorization was going to fail anyway.

winstxnhdw:
I have read your [reply](https://github.com/grammyjs/grammY/pull/816#issuecomment-3180716466) but I am still having trouble understanding how this would stop grammY from throwing an exception if `request.json()` does not contain parseable JSON.

MKRhere:
@winstxnhdw It'll still throw a rejection, but it'll be handleable.

In the current case, request.json() is called early, and it becomes a hanging promise until `handleUpdate(await update)`. This means when any other await happens in between, the hanging update stays unhandled, and if it rejects, it gets thrown as an unhandledRejection which you cannot catch, and crashes the bot.

MKRhere:
The flow looks like this:

```TS
// created inside the framework adapter and returned
const update = request.json(); // not awaited

if (header !== token) {
	await unauthorized();
	// stays unawaited
	return handlerReturn;
}

await handleUpdate(await update); // awaited here
```

What happens to the `update` promise in the unauthorized case? It becomes an unhandled promise. And then if it throws, there's nobody to catch the rejection, since the `await update` never happens.

This PR changes it to:

```TS
// framework adapter called
// but the `update` getter is not triggered yet
const handler = server(...args);

if (header !== token) {
	await unauthorized();
	// handler.update is never accessed
	// the promise is never created
	return handler.handlerReturn;
}

// handler.update causes the getter to run
// and create the promise by calling request.json()
// and since it's immediately awaited,
// it becomes a "handled" promise
await handleUpdate(await handler.update);
```

KnorpelSenf:
@all-contributors add @MKRhere for bug and code

PR #819, opened 2025-08-15
feat: support Bot API 9.2

KnorpelSenf:
https://core.telegram.org/bots/api#august-15-2025


src/context.ts
rojvv, line 1: Forgot to add the default value for `direct_messages_topic_id` to `replyWithSticker`?

KnorpelSenf: Yes, will fix it

rojvv (approved):

PR #820, opened 2025-08-16
fix: add is_topic_message check before setting message_thread_id

KnightNiwrem:
Regular message in general topic
```ts
sendMessage
{
  chat_id: 41284431,
  text: "hi",
  business_connection_id: undefined,
  direct_messages_topic_id: undefined,
}
```

Regular message in test topic
```ts
sendMessage
{
  chat_id: -1003070712535,
  text: "hi",
  business_connection_id: undefined,
  message_thread_id: 2,
  direct_messages_topic_id: undefined,
}
```

Reply to (and reply to reply to, and so on) messages in general topic
```ts
sendMessage
{
  chat_id: -1003070712535,
  text: "hi",
  business_connection_id: undefined,
  direct_messages_topic_id: undefined,
}

sendMessage
{
  chat_id: -1003070712535,
  text: "hi",
  business_connection_id: undefined,
  direct_messages_topic_id: undefined,
}

sendMessage
{
  chat_id: -1003070712535,
  text: "hi",
  business_connection_id: undefined,
  direct_messages_topic_id: undefined,
}
```

Reply to (and reply to reply to, and so on) message in test topic
```ts
sendMessage
{
  chat_id: -1003070712535,
  text: "hi",
  business_connection_id: undefined,
  message_thread_id: 2,
  direct_messages_topic_id: undefined,
}

sendMessage
{
  chat_id: -1003070712535,
  text: "hi",
  business_connection_id: undefined,
  message_thread_id: 2,
  direct_messages_topic_id: undefined,
}
```

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -977,16 +977,16 @@ export class Context implements RenamedUpdate {
         other?: Other<"sendMessage", "chat_id" | "text">,
         signal?: AbortSignal,
     ) {
+        const msg = this.msg;
         return this.api.sendMessage(
             orThrow(this.chatId, "sendMessage"),
             text,
             {
                 business_connection_id: this.businessConnectionId,
-                ...(this.msg?.is_topic_message
-                    ? { message_thread_id: this.msg?.message_thread_id }
+                ...(msg?.is_topic_message
+                    ? { message_thread_id: msg?.message_thread_id }
                     : {}),
-                direct_messages_topic_id: this.msg?.direct_messages_topic
-                    ?.topic_id,
+                direct_messages_topic_id: msg?.direct_messages_topic?.topic_id,
                 ...other,
             },
             signal,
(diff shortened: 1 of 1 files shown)


KnorpelSenf (changes requested):
Excellent! 

Can you extract the `ctx.msg` to a variable so the getter isn't computed repeatedly for every API call? 

Also, do you think we should deduplicate the `ctx.msg` and `ctx.msgId` getters so that the `ctx.msg` getter _really_ isn't computed repeatedly?

KnightNiwrem:
> Excellent!
> 
> Can you extract the `ctx.msg` to a variable so the getter isn't computed repeatedly for every API call?

Any chore is possible with LLMs /runs

> Also, do you think we should deduplicate the `ctx.msg` and `ctx.msgId` getters so that the `ctx.msg` getter _really_ isn't computed repeatedly?

I think I have an idea.. for a separate PQ

KnorpelSenf (approved):
Tremendous

PR #821, opened 2025-08-16
feat: Use lazy initialization for msg, chat, and from

KnightNiwrem:
Reduce N number of ternary comparison down to a single boolean comparison, if the property has been accessed before.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -407,6 +407,15 @@ export class Context implements RenamedUpdate {
          */
         public readonly me: UserFromGetMe,
     ) {}
+    // Lazy-init backing fields: these values can legitimately be `undefined`,
+    // so we keep a boolean flag per value to indicate whether we already
+    // computed (inited) them.
+    private _msg?: Message;
+    private _msgInited = false;
+    private _chat?: Chat;
+    private _chatInited = false;
+    private _from?: User;
+    private _fromInited = false;
 
     // UPDATE SHORTCUTS
 
@@ -514,15 +523,17 @@ export class Context implements RenamedUpdate {
      */
     get msg(): Message | undefined {
         // Keep in sync with types in `filter.ts`.
-        return (
-            this.message ??
+        if (!this._msgInited) {
+            this._msg = this.message ??
                 this.editedMessage ??
                 this.channelPost ??
                 this.editedChannelPost ??
                 this.businessMessage ??
                 this.editedBusinessMessage ??
-                this.callbackQuery?.message
-        );
+                this.callbackQuery?.message;
+            this._msgInited = true;
+        }
+        return this._msg;
     }
     /**
      * Get the chat object from wherever possible. Alias for `(this.msg ??
@@ -532,17 +543,21 @@ export class Context implements RenamedUpdate {
      */
     get chat(): Chat | undefined {
         // Keep in sync with types in `filter.ts`.
-        return (
-            this.msg ??
-                this.deletedBusinessMessages ??
-                this.messageReaction ??
-                this.messageReactionCount ??
-                this.myChatMember ??
-                this.chatMember ??
-                this.chatJoinRequest ??
-                this.chatBoost ??
-                this.removedChatBoost
-        )?.chat;
+        if (!this._chatInited) {
+            this._chat = (
+                this.msg ??
+                    this.deletedBusinessMessages ??
+                    this.messageReaction ??
+                    this.messageReactionCount ??
+                    this.myChatMember ??
+                    this.chatMember ??
+                    this.chatJoinRequest ??
+                    this.chatBoost ??
+                    this.removedChatBoost
+            )?.chat;
+            this._chatInited = true;
+        }
+        return this._chat;
     }
     /**
      * Get the sender chat object from wherever possible. Alias for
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
I don't find it obvious that this would make things faster. Sure, we save a few property access operations and nullish checks. On the flipside:
- we increase GC pressure (more references)
- we increase memory consumption
- we optimise for an uncommon case: most bots perform a low, constant number of reads on the context object and they usually cache the values on the stack anyway if they access them several times

I'd at least like to better understand these changes. Perhaps we should bench them for 1M context object creations with `{0..3}` property access calls each to find out how this changes things?

KnightNiwrem:
Some context: This is an alternative way to address, not so much an performance-focused PQ

> Also, do you think we should deduplicate the ctx.msg and ctx.msgId getters so that the ctx.msg getter really isn't computed repeatedly?

The most straightforward way was to simply change all `this.msgId` with `msg?.message_id`, given that we already do `const msg = this.msg`.

However, this creates some degree of cognitive load to remember creating some objects when we intend to access the object more than once in the ctx methods.

The same also applies to plugin writers and developers, if they are interested in such optimisations. It's also not as pretty - we do want to simply do `ctx.msgId` if we can. Especially broadcast bots doing `ctx.copyMessage` or `ctx.forwardMessage`.

This alternative appears accomplish the same thing as the most straightforward way as well. After `this.msgId` is accessed, any further calls to `this.msg?.is_topic_message`, for example, no longer performs duplicate computation.

It covers all methods that don't currently do `const msg = this.msg` for any future updates where it tries to access more properties of msg.

And it incidentally covers plugin writers and developers so they simply need not think about such optimisation, and may write however is most intuitive.

That said, a perf comparison would still be interesting. Just wanted to clarify the thoughts behind this approach.

KnightNiwrem:
Re: GC and memory

Not really expecting much impact in those places. A Context must contain an Update. It cannot not have an Update. So either the Context has 2 references to Message (and other) or it has 0 references to Message (because it is destroyed).

There is no case where Context has a reference to Message via the new field, yet the Message is not referenced by Context indirectly via Update.

I'm expecting the reference to be a fixed byte size value, since it is simply a shallow reference.

The biggest concern for memory probably would be when Context is JSONified and stored. Then the fields would need to be dereferenced and deeply copied. We can probably workaround that for our plugins, but it would still be a concern for JSON.stringify done externally.

~~Or perhaps we can workaround that by using native JS private fields with the `#` prefix.~~

KnightNiwrem:
This is the closest I can get to excluding these fields from enumeration and `JSON.stringify` without resorting to making hard private and mapped types work together.

```ts
class X {
    _a?: number;
    b: boolean = true;

    constructor() {
        Object.defineProperty(this, "_a", {
            enumerable: false,
            writable: true,
        });
    }

    get a() {
        if (!this._a) {
            this._a = 1;
        }
        return this._a;
    }
}

const x = new X();
console.log(x.a);
console.log(JSON.stringify(x));

for (const k of Object.keys(x)) {
    console.log(k);
}
```

Result:
```ts
1
{"b":true}
b
```

KnightNiwrem:
Worsens performance based on benchmarks. Guess we have to just do it the original way

Baseline from v1.38.0
```
| benchmark                               | time/iter (avg) |        iter/s |      (min … max)      |      p75 |      p99 |     p995 |
| --------------------------------------- | --------------- | ------------- | --------------------- | -------- | -------- | -------- |
| Bench calls to ctx.copyMessage never    |          2.6 µs |       391,400 | (  1.6 µs …  12.8 ms) |   1.9 µs |   4.7 µs |   6.5 µs |
| Bench calls to ctx.copyMessage once     |          1.8 µs |       554,400 | (  1.8 µs …   1.9 µs) |   1.8 µs |   1.9 µs |   1.9 µs |
| Bench calls to ctx.copyMessage twice    |          1.8 µs |       551,400 | (  1.8 µs …   2.0 µs) |   1.8 µs |   2.0 µs |   2.0 µs |
| Bench calls to ctx.copyMessage thrice   |          1.8 µs |       559,300 | (  1.7 µs …   1.9 µs) |   1.8 µs |   1.9 µs |   1.9 µs |
```

With lazy init
```
| benchmark                               | time/iter (avg) |        iter/s |      (min … max)      |      p75 |      p99 |     p995 |
| --------------------------------------- | --------------- | ------------- | --------------------- | -------- | -------- | -------- |
| Bench calls to ctx.copyMessage never    |          8.6 µs |       116,100 | (  5.1 µs …  13.0 ms) |   6.4 µs |  13.9 µs |  29.1 µs |
| Bench calls to ctx.copyMessage once     |          5.5 µs |       180,400 | (  5.1 µs …   6.5 µs) |   6.2 µs |   6.5 µs |   6.5 µs |
| Bench calls to ctx.copyMessage twice    |          5.3 µs |       189,100 | (  5.1 µs …   6.3 µs) |   5.2 µs |   6.3 µs |   6.3 µs |
| Bench calls to ctx.copyMessage thrice   |          5.3 µs |       190,500 | (  5.1 µs …   5.4 µs) |   5.3 µs |   5.4 µs |   5.4 µs |
```

KnightNiwrem:
Benchmarking source:

```ts
const BOT_TOKEN = '';
import { Bot, type Context } from 'change this between local build and remote release';

function createUpdate(): Context['update'] {
  return {
    "update_id": 908295308,
    "message": {
      "message_id": 9396,
      "from": {
        "id": 41284431,
        "is_bot": false,
        "first_name": "KnightNiwrem",
        "username": "Knightniwrem",
        "language_code": "en"
      },
      "chat": {
        "id": 41284431,
        "first_name": "KnightNiwrem",
        "username": "Knightniwrem",
        "type": "private"
      },
      "date": 1755357026,
      "text": "hi"
    }
  };
}

const benchMiddlewares = {
  'never': () => {},
  'once': async (ctx: Context) => { await ctx.copyMessage('1') },
  'twice': async (ctx: Context) => { await ctx.copyMessage('2'); await ctx.copyMessage('2'); },
  'thrice': async (ctx: Context) => { await ctx.copyMessage('3'); await ctx.copyMessage('3'); await ctx.copyMessage('3'); },
}
let currentBenchFn: Function;

const bot = new Bot(BOT_TOKEN, {
  botInfo: {
    "id": 1561644845,
    "is_bot": true,
    "first_name": "Heroku Experiment",
    "username": "he_exp_bot",
    "can_join_groups": true,
    "can_read_all_group_messages": false,
    "supports_inline_queries": true,
    "can_connect_to_business": false,
    "has_main_web_app": true
  }
});

// @ts-expect-error
bot.api.config.use(() => {
  return { ok: true, result: true };
});

bot.on('message:text', async ctx => {
  await currentBenchFn(ctx);
});

for (const [benchKey, benchFn] of Object.entries(benchMiddlewares)) {
  currentBenchFn = benchFn;
  Deno.bench(`Bench calls to ctx.copyMessage ${benchKey}`, async () => {
    await bot.handleUpdate(createUpdate());
  });
}
```

KnightNiwrem:
Lazy init without `defineProperty`
```
| benchmark                               | time/iter (avg) |        iter/s |      (min … max)      |      p75 |      p99 |     p995 |
| --------------------------------------- | --------------- | ------------- | --------------------- | -------- | -------- | -------- |
| Bench calls to ctx.copyMessage never    |          3.1 µs |       317,700 | (  1.6 µs …  23.1 ms) |   1.8 µs |   5.1 µs |   7.5 µs |
| Bench calls to ctx.copyMessage once     |          1.8 µs |       551,900 | (  1.7 µs …   2.5 µs) |   1.8 µs |   2.5 µs |   2.5 µs |
| Bench calls to ctx.copyMessage twice    |          1.8 µs |       543,600 | (  1.7 µs …   2.0 µs) |   1.9 µs |   2.0 µs |   2.0 µs |
| Bench calls to ctx.copyMessage thrice   |          1.8 µs |       563,700 | (  1.7 µs …   1.9 µs) |   1.8 µs |   1.9 µs |   1.9 µs |
```

KnorpelSenf:
https://t.me/grammyjs/317684

PR #822, opened 2025-09-01
fix: do not pass message_thread_id to unrelated chats

KnorpelSenf:
When sending messages to other chats, we must not transfer the thread identifier of the origin chat to the target thread.


KnorpelSenf:
@all-contributors add @dotvhs for bug and review btw

PR #825, opened 2025-09-08
fix: properly end fastify requests

### LINKED ISSUE #791: "TypeError: reply.status is not a function" while using fastify adapter
The `end` method inside fastify adapter doesn't work correctly for now:
 `TypeError: reply.status is not a function` appears.

 It seems related to the problem mentioned in #540 . `Reply` was fixed, but `end` wan't

KnorpelSenf:
Fixes #791


KnorpelSenf:
@shevernitskiy could you maybe review this?

PR #828, opened 2025-10-03
fix: request types of Cloudflare Workers adapter

### LINKED ISSUE #762: Cloudflare Request type error
[Cloudflare Request](https://developers.cloudflare.com/workers/runtime-apis/request/#the-cf-property-requestinitcfproperties) is different from normal request. It should use this [Request](https://www.npmjs.com/package/@cloudflare/workers-types] by Cloudflare.

https://github.com/grammyjs/grammY/blob/3299200f6fe56075aa7d4c4679e54ecef051ef29/src/convenience/frameworks.ts#L108-L110

SpEcHiDe:
Fixes #762 

Copied most of types from: https://github.com/cloudflare/workerd/blob/07db06bfe7320b1c5393a3702a6902eea41eee9e/types/generated-snapshot/latest/index.d.ts#L1697

### DIFF
--- a/src/convenience/frameworks.ts
+++ b/src/convenience/frameworks.ts
@@ -100,19 +100,21 @@ export type BunAdapter = (request: {
     json: () => Promise<Update>;
 }) => ReqResHandler<Response>;
 
-interface CfPropertiesRequest extends Body {
-    method: string;
-    url: string;
-    headers: Headers;
-}
-
 export type CloudflareAdapter = (event: {
-    request: CfPropertiesRequest;
+    request: Body & {
+        method: string;
+        url: string;
+        headers: Headers;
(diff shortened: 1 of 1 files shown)


KnorpelSenf (changes requested):
Can you

1. delete all properties that aren't relevant to grammY? This does not break assignability, as excess properties are permitted.
2. delete all comments?

KnorpelSenf (approved):
Very nice, thank you!

KnorpelSenf:
@SpEcHiDe consider signing your commits in the future. We usually require this from all contributors. I will ignore that for now and merge anyway.

FYI his is easily done using your existing SSH keys, it's a straightforward setup.

PR #833, opened 2025-10-05
Fix a few comments in types containing strings interpreted as JSDoc

rayz1065:
Really minor change, a few strings are interpreted as JSDoc currently, for example see how the code below is rendered on vscode:
```ts
    /**
     * Use this parameter if you want to remove the keyboard for specific users only. Targets: 1) users that are @mentioned in the _text_ of the {@link Message | Message} object; 2) if the bot's message is a reply to a message in the same chat and forum topic, sender of the original message.
     *
     * _Example:_ A user votes in a poll, bot returns confirmation message in reply to the vote and removes the keyboard for that user, while still showing the keyboard with poll options to users who haven't voted yet.
     */
    selective?: boolean;
```
image

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -4514,7 +4514,7 @@ export interface ReplyKeyboardMarkup {
      */
     input_field_placeholder?: string;
     /**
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
*a few more

Can you review #835?

PR #835, opened 2025-10-07
docs: fix bad usages of @

KnorpelSenf:
Closes #833


KnorpelSenf:
Now that I think about it, I actually believe it doesn't matter much. I'm going to rewrite all TSDoc comments after the implementation is complete. So I could've skipped these files actually.

My style has improved over the last 4 years, and I would like to do it properly before releasing. Probably after the website docs are rewritten and while they're being translated.

PR #853, opened 2025-12-06
fix: implement constant-time comparison for webhook secret token

### LINKED ISSUE #641: Concern regarding secret token validation and timing attacks
https://github.com/grammyjs/grammY/blob/48153f97840df5af0ce3706313d5052c370e0fce/src/convenience/webhook.ts#L112-L117

A timing attack is when an attacker can infer information about a secret by measuring how long it takes to compare values. If the comparison time varies based on the content, it could be exploited.

I found a relevant discussion on Stack Overflow that touches on this topic: https://stackoverflow.com/questions/31095905/whats-the-difference-between-a-secure-compare-and-a-simple

To address this risk, I was thinking we could implement a constant-time comparison. Here’s a rough idea (still untested):

```ts
const te = new TextEncoder();
const headerHash = await crypto.subtle.digest(
    "SHA-256",
    te.encode(header),
);
const tokenHash = await crypto.subtle.digest(
    "SHA-256",
    te.encode(token),
);
if (timingSafeEqual(headerHash, tokenHash)) {
    await unauthorized();
    // TODO: investigate deno bug that happens when this console logging is removed
    console.log(handlerReturn);
    return handlerReturn;
}
```

Does it make sense to be concerned about timing attacks in this context?

arunr-inji:
## Fix: Constant-time comparison for webhook secret token

Replaces vulnerable string comparison with constant-time byte-by-byte comparison to prevent timing attacks that could reveal secret token information.

- Always iterates `tokenBytes.length` times (constant-time)
- Prevents timing-based information leakage
- Handles undefined cases correctly

Fixes #641

### DIFF
--- a/src/convenience/webhook.ts
+++ b/src/convenience/webhook.ts
@@ -52,7 +52,7 @@ function compareSecretToken(
 
     // If lengths differ, reject
     if (headerBytes.length !== tokenBytes.length) {
-      return false;
+        return false;
     }
 
     let hasDifference = 0;


arunr-inji:
@KnorpelSenf let me know if you have any feedback for this - thank you!

src/convenience/webhook.ts
```diff
@@ -23,6 +23,53 @@ const callbackAdapter: FrameworkAdapter = (
...
+    // that could reveal the secret token's length. The loop time is constant
+    // relative to the secret token length, not the attacker's input length.
+    for (let i = 0; i < tokenBytes.length; i++) {
+        // If header is shorter than token, pad with 0 for comparison
+        const headerByte = i < headerBytes.length ? headerBytes[i] : 0;
+        const tokenByte = tokenBytes[i];
+
+        // If bytes differ, mark that we found a difference
+        // Using bitwise OR to maintain constant-time (no short-circuit evaluation)
+        hasDifference = hasDifference | (headerByte !== tokenByte ? 1 : 0);
```
KnorpelSenf: Let's avoid the needless branching:
```suggestion
        hasDifference |= headerByte ^ tokenByte;
```

src/convenience/webhook.ts
```diff
@@ -23,6 +23,53 @@ const callbackAdapter: FrameworkAdapter = (
...
+    }
+
+    // Convert strings to Uint8Array for byte-by-byte comparison
+    const encoder = new TextEncoder();
+    const headerBytes = encoder.encode(header);
+    const tokenBytes = encoder.encode(token);
+
+    // Check if lengths differ (1 if different, 0 if same)
+    let hasDifference = headerBytes.length !== tokenBytes.length ? 1 : 0;
+
```
KnorpelSenf: There is nothing to be gained from keeping the token length a secret. We can reject early.
```suggestion
    // If lengths differ, reject
    if (headerBytes.length !== tokenBytes.length) {
      return false;
    }

    let hasDifference = 0;
```

KnorpelSenf (changes requested):
Almost there. Sorry for the delay, I had way too many other things to do. Perhaps we can even get it done this year?

arunr-inji:
> Almost there. Sorry for the delay, I had way too many other things to do. Perhaps we can even get it done this year?

I just pushed an update and also setup verification for my account! Let me know if it looks ok to you

KnorpelSenf (approved):
Excellent changes! 

Thanks for signing your commits now :)

It's 23:50 in my timezone, so I'll count this as _merged in 2025_ 🎇

KnorpelSenf:
@all-contributors add @arunr-inji for ideas and code and test!

arunr-inji:
thanks for all the feedback and for adding @KnorpelSenf !

KnorpelSenf:
You're very welcome! Thank you for everything you've been doing lately, and for the patience when I'm down some rabbit hole and not replying as fast as I should. Happy New Year! 🎉

PR #855, opened 2025-12-22
perf: do not perform useless checks

KnorpelSenf:
Those were added needlessly, so we can remove them


KnightNiwrem:
For future readers: if msg?.xyz is truthy, then msg cannot be falsy. Hence msg is not undefined in the true branch of the ternary expression.

PR #858, opened 2025-12-31
feat: support Bot API 9.3

KnorpelSenf:
https://core.telegram.org/bots/api#december-31-2025

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 <!-- deno-fmt-ignore-start -->
 
-[![Bot API](https://img.shields.io/badge/Bot%20API-9.2-blue?logo=telegram&style=flat&labelColor=000&color=3b82f6)](https://core.telegram.org/bots/api)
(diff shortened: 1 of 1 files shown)


src/context.ts
```diff
@@ -993,6 +993,35 @@ export class Context implements RenamedUpdate {
...
+        other?: Other<"sendMessageDraft", "chat_id" | "text">,
+        signal?: AbortSignal,
+    ) {
+        const msg = this.msg;
+        return this.api.sendMessageDraft(
+            orThrow(this.chatId, "sendMessageDraft"),
+            this.update.update_id,
+            text,
+            {
+                ...(msg?.is_topic_message
```
roziscoding: question (non-blocking): will this ever be false? and, if it is, will the API call fail? should we error preemptively to avoid an unnecessary API call and provide a clear error?

roziscoding (approved):

src/context.ts
KnorpelSenf: Yes, it is false if the bot does not have topics enabled. grammY should then call the method without the thead identifier as usual.

The API call will fail with the error message that the Bot API server generates. For example, if it ever decides to support this message without a thread identifier, then this code will keep on performing the right thing.

Issue #851, opened 2025-12-06
Improve test coverage for bot.ts


arunr-inji:
## Current State

`bot.ts` has 0% branch coverage and 41.5% line coverage. The existing tests only check basic token validation but don't test the rest of the functionality. 

<img width="319" height="273" alt="Image" src="https://github.com/user-attachments/assets/ac5171dc-9e54-4917-a209-23c60b4852ea" />

This issue is to improve test coverage for bot.ts. This is coming off of discussion [here](https://github.com/grammyjs/grammY/pull/842#discussion_r2591561078)

arunr-inji:
@KnorpelSenf let me know if there's anything specific you'd like to see here. I can pick this issue up!

KnorpelSenf:
I don't think there should be any surprises here.

The hard part here is probably to have good coverage of all the different ways in which things can happen. There is a bunch of promise handling, and the order of operations (when does which promise resolve) leads to vastly different code paths. This is probably best implemented using `Promise.withResolvers` which lets you resolve promises explicitly.

Otherwise, things are straightforward.

arunr-inji:
@KnorpelSenf here's a first attempt at improving coverage for bot.ts: https://github.com/grammyjs/grammY/pull/860 

Let me know what you think!

KnorpelSenf:
@arunr-inji thank you! Review is there

PR #860, opened 2026-01-10
test: cover bot.ts 

arunr-inji:
## Description
This PR adds comprehensive test coverage for `bot.ts` 

Closes https://github.com/grammyjs/grammY/issues/851

## Test Coverage Improvements
- **Before**: 15.4% line coverage, 50.0% branch coverage
- **After**:
image

### DIFF
--- a/test/bot.test.ts
+++ b/test/bot.test.ts
@@ -144,7 +144,7 @@ describe("Bot initialization", () => {
     it("should retry on HttpError during initialization", async () => {
         const bot = new Bot(token);
         let callCount = 0;
-        using getMeStub = stub(bot.api, "getMe", () => {
+        using _ = stub(bot.api, "getMe", () => {
             callCount++;
             if (callCount === 1) {
                 throw new HttpError("Network error", { error: "ECONNREFUSED" });
@@ -161,7 +161,7 @@ describe("Bot initialization", () => {
     it("should retry on 5xx errors during initialization", async () => {
         const bot = new Bot(token);
         let callCount = 0;
-        using getMeStub = stub(bot.api, "getMe", () => {
+        using _ = stub(bot.api, "getMe", () => {
             callCount++;
             if (callCount === 1) {
                 throw new GrammyError(
@@ -187,7 +187,7 @@ describe("Bot initialization", () => {
     it("should handle 429 without retry_after parameter", async () => {
         const bot = new Bot(token);
         let callCount = 0;
-        using getMeStub = stub(bot.api, "getMe", () => {
+        using _ = stub(bot.api, "getMe", () => {
             callCount++;
             if (callCount === 1) {
                 throw new GrammyError(
@@ -481,29 +481,33 @@ describe("Bot error handling", () => {
             const { promise: secondCall, resolve: resolveSecondCall } = Promise
                 .withResolvers<Update[]>();
 
-            using _ = stub(bot.api, "getUpdates", (_params, signal?: AbortSignal) => {
-                callCount++;
-                switch (callCount) {
-                    case 1:
-                        // First call: return update that triggers middleware error
-                        signal!.addEventListener(
-                            "abort",
-                            () => resolveGetUpdates([]),
-                        );
-                        return getUpdatesPromise;
-                    case 2:
-                        // Second call: next polling iteration
-                        notifySecondCall();
-                        signal!.addEventListener(
-                            "abort",
-                            () => resolveSecondCall([]),
-                        );
-                        return secondCall;
-                    default:
-                        // Confirmation call
-                        return Promise.resolve([]);
-                }
-            });
+            using _ = stub(
+                bot.api,
+                "getUpdates",
+                (_params, signal?: AbortSignal) => {
+                    callCount++;
+                    switch (callCount) {
+                        case 1:
+                            // First call: return update that triggers middleware error
+                            signal!.addEventListener(
+                                "abort",
+                                () => resolveGetUpdates([]),
+                            );
+                            return getUpdatesPromise;
+                        case 2:
+                            // Second call: next polling iteration
+                            notifySecondCall();
+                            signal!.addEventListener(
+                                "abort",
+                                () => resolveSecondCall([]),
+                            );
+                            return secondCall;
+                        default:
+                            // Confirmation call
+                            return Promise.resolve([]);
+                    }
+                },
+            );
 
             // Start polling
             const startPromise = bot.start();
@@ -551,7 +555,7 @@ describe("Bot polling lifecycle", () => {
         const { promise: firstCallStarted, resolve: notifyFirstCall } = Promise
             .withResolvers<void>();
 
-        using getUpdatesStub = stub(
+        using _ = stub(
             bot.api,
             "getUpdates",
             (_params, signal?: AbortSignal) => {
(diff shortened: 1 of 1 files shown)


test/bot.test.ts
```diff
@@ -1,22 +1,704 @@
...
+                .withResolvers<Update[]>();
+
+            const getUpdatesStub = stub(
+                bot.api,
+                "getUpdates",
+                (_params, signal?: AbortSignal) => {
+                    // Check for confirmation call first (no signal)
+                    if (!signal) {
+                        return Promise.resolve([]);
+                    }
```
KnorpelSenf: This test is good, but it would be slightly easier to understand if the code path that is taken last is also placed last. Pseudo:
```
callCount++
switch(callCount) {
  case 1:
    // first case
    return Promise.reject(...)
  case 2:
    // second case
    return retryPromise
  // etc
  default:
    return Promise.resolve([])
}
```

KnorpelSenf: Just an idea, I'll leave it up to you if you want to do this

KnorpelSenf:
Still some unaddressed comments from the previous review about needles fake time usage (unless it's needed and I don't understand why), feel free to ping me once this is ready

arunr-inji:
> Still some unaddressed comments from the previous review about needles fake time usage (unless it's needed and I don't understand why), feel free to ping me once this is ready

Yeah I meant to convert this into a draft again. I'll do that and re open for review!

KnorpelSenf:
Top notch

test/bot.test.ts
```diff
@@ -1,22 +1,821 @@
...
+            await assertRejects(() => init2);
+            assertEquals(bot.isInited(), false);
+            // Should only call getMe once (deduplication even on error)
+            assertEquals(getMeStub.calls.length, 1);
+        } finally {
+            getMeStub.restore();
+        }
+    });
+
+    it("should retry on HttpError during initialization", async () => {
```
arunr-inji: @KnorpelSenf added this and two more tests below (5xx, 429 handling) to this describe block since you last reviewed

arunr-inji:
> Top notch

I think I addressed your feedback + added 3 more tests (mentioned in my other comment)

test/bot.test.ts
```diff
@@ -1,22 +1,821 @@
...
+};
 
-Deno.test("should take a token", () => {
-    const bot = createBot("fake-token");
-    assertEquals(bot.token, "fake-token");
+const testUpdate: Update = {
+    update_id: 1,
+    message: {
+        message_id: 1,
+        date: Date.now(),
```
KnorpelSenf: I know it does not matter in any way, shape, or form, but for some reason it is unsettling to me
```suggestion
        date: Date.now() / 1000,
```

test/bot.test.ts
```diff
@@ -1,22 +1,821 @@
...
+            // Resolve the single getMe call
+            resolvers[0](botInfo);
+
+            await Promise.all([init1, init2, init3]);
+            assertEquals(bot.isInited(), true);
+            // Verify stub was only called once
+            assertEquals(getMeStub.calls.length, 1);
+        } finally {
+            getMeStub.restore();
+        }
```
KnorpelSenf: Ditto:
```suggestion
        using getMeStub = stub(bot.api, "getMe", () => {
            const { promise, resolve } = Promise.withResolvers<UserFromGetMe>();
            resolvers.push(resolve);
            return promise;
        });

        // Start three concurrent inits
        const init1 = bot.init();
        const init2 = bot.init();
        const init3 = bot.init();

        // Verify deduplication: only one getMe call made
        assertEquals(resolvers.length, 1);

        // Resolve the single getMe call
        resolvers[0](botInfo);

        await Promise.all([init1, init2, init3]);
        assertEquals(bot.isInited(), true);
        // Verify stub was only called once
        assertEquals(getMeStub.calls.length, 1);
```

test/bot.test.ts
```diff
@@ -1,22 +1,821 @@
...
+            reject(new Error("Network error"));
+
+            await assertRejects(() => init1);
+            await assertRejects(() => init2);
+            assertEquals(bot.isInited(), false);
+            // Should only call getMe once (deduplication even on error)
+            assertEquals(getMeStub.calls.length, 1);
+        } finally {
+            getMeStub.restore();
+        }
```
KnorpelSenf: Ditto:
```suggestion
        using getMeStub = stub(bot.api, "getMe", () => promise);

        const init1 = bot.init();
        const init2 = bot.init();

        // Reject with plain Error (won't trigger retries, only HttpError/GrammyError do)
        reject(new Error("Network error"));

        await assertRejects(() => init1);
        await assertRejects(() => init2);
        assertEquals(bot.isInited(), false);
        // Should only call getMe once (deduplication even on error)
        assertEquals(getMeStub.calls.length, 1);
```

test/bot.test.ts
```diff
@@ -1,22 +1,821 @@
...
+
+            // Verify no additional calls were made
+            assertEquals(deleteWebhookStub.calls.length, 1);
+            assertEquals(getUpdatesCount, 1);
+
+            await bot.stop();
+            await startPromise1;
+        } finally {
+            getUpdatesStub.restore();
+        }
```
KnorpelSenf: Another case of `using`:

```suggestion
        using getUpdatesStub = stub(
            bot.api,
            "getUpdates",
            (_params, signal?: AbortSignal) => {
                getUpdatesCount++;
                switch (getUpdatesCount) {
                    case 1:
                        notifyFirstCall();
                        return new Promise((_, reject) => {
                            signal!.addEventListener(
                                "abort",
                                () => reject(new Error("Aborted")),
                            );
                        });
                    default:
                        // Subsequent calls
                        return Promise.resolve([]);
                }
            },
        );

        const startPromise1 = bot.start();
        // Wait for polling to actually start
        await firstCallStarted;

        // Verify first start called setup methods
        assertEquals(deleteWebhookStub.calls.length, 1);
        assertEquals(getUpdatesCount, 1);

        // Second start should be a no-op
        await bot.start();

        // Verify no additional calls were made
        assertEquals(deleteWebhookStub.calls.length, 1);
        assertEquals(getUpdatesCount, 1);

        await bot.stop();
        await startPromise1;
```

test/bot.test.ts
```diff
@@ -1,22 +1,821 @@
...
+    });
+
+    it("should delete webhook on start", async () => {
+        let callCount = 0;
+        const { promise: firstCallStarted, resolve: notifyFirstCall } = Promise
+            .withResolvers<void>();
+
+        stub(bot.api, "getUpdates", (_params, signal?: AbortSignal) => {
```
KnorpelSenf:
```suggestion
        using _ = stub(bot.api, "getUpdates", (_params, signal?: AbortSignal) => {
```

KnorpelSenf (changes requested):
A bunch of stubs are not cleaned up properly, and a few usages can be simplified. All of my comments are nitpicks, I assume that you can just commit my suggestions (perhaps take a look if I messed up) and then we can merge. Almost there!

arunr-inji:
> A bunch of stubs are not cleaned up properly, and a few usages can be simplified. All of my comments are nitpicks, I assume that you can just commit my suggestions (perhaps take a look if I messed up) and then we can merge. Almost there!

I committed your changes but had to fix some lint failures so pushed a commit. 

Thanks for the thorough reviews!

test/bot.test.ts
KnorpelSenf: Ah great point, the first error is retried immediately. This is reasonable because of a quirk in the Bot API server.

The tests are fine the way they are, let's merge them now. In a subsequent PQ, it is worth improving the tests, for the retry mechanism in particular.

PR #861, opened 2026-01-11
chore: remove duplicate contributor

arunr-inji:
https://github.com/grammyjs/grammY/pull/859 added a duplicate entry due to a github username change (but the account remains the same) 

Is it possible to merge the two?


KnorpelSenf:
That's absolutely possible. I'll merge right after the conflicts are fixed.

In the long run, we should maybe find a way to detect username changes and update our file.

arunr-inji:
@KnorpelSenf re: 
> In the long run, we should maybe find a way to detect username changes and update our file.

One approach that comes to mind is via a weekly github action workflow; this can detect if a user's github login (account name) changed and updating it in the contributors file. This means that we'd need to use Github's APIs to detect changes. lmk what you think!

KnorpelSenf:
I like it! It's essentially like renovate but for contributors not dependencies :D

Do you know how to build this? Wanna open a pull request?

arunr-inji:
> I like it! It's essentially like renovate but for contributors not dependencies :D
> 
> Do you know how to build this? Wanna open a pull request?

yeah i can take a stab at it!

KnorpelSenf:
Amazing!

Issue #864, opened 2026-01-16
perf: move attachment identifier to `InputFile`


KnorpelSenf:
Instead of creating a function object per `InputFile` instance, place the `toJSON` function on the prototype. This reduces complexity and slightly improves CPU load and GC pressure.

KnorpelSenf:
Done on the `v2` branch

PR #865, opened 2026-01-16
fix: upload retries

### LINKED ISSUE #862: extractFiles and auto-retry plugin conflict
When sending a lot of medias using sendMediaGroup/sendPhoto/sendVideo/... I can hit Telegram rate limit. I use https://grammy.dev/plugins/auto-retry to auto-retry sending requests in this case. It works well for any text messages and already uploaded medias (with file_id reference). But when I try to send many files from file system I often face `Call to 'sendMediaGroup' failed! (400: Bad Request: can't parse InputMedia: media not found)`.

This error is returned from Telegram when media field is `attach://...` but no file data is included

The problem is that grammy modifies payload in `extractFiles` function and it is fine unless `payloadToMultipartItr` is called once. But with auto-retry plugin `payloadToMultipartItr` can be triggered multiple times causing `extractFiles` being called 1+ times.
On first iteration it takes `InputFile`s and replaces them with `attach://...` references, but on a next iteration medias are already kept as references but data was already sent during first iteration and on second request Telegram only receives references without data.

Reproduction:
Download small png from https://evanhahn.com/worlds-smallest-png/ (we have to send files but size does not matter, only amount per second)
```javascript
import fs from 'node:fs'
import { Bot, InputFile, GrammyError } from 'grammy'
import { autoRetry } from '@grammyjs/auto-retry'

const bot = new Bot('BOTTOKEN')
bot.api.config.use(
  autoRetry({
    maxRetryAttempts: 2,
    maxDelaySeconds: 60
  })
)

const arrayToChunks = (arr, amount) => {
  let resp = []
  for (let i = 0; i < arr.length; i += amount) {
    const chunk = arr.slice(i, i + amount)
    resp.push(chunk)
  }
  return resp
}

;(async () => {
  const medias = Array.from({ length: 60 }).map((_, i) => ({
    type: 'photo',
    media: ['./small.png', `file${i}.png`],
    caption: `c ${i}`
  }))
  await Promise.all(
    arrayToChunks(medias, 10).map(async (imagesChunk) => {
      await bot.api.raw.sendMediaGroup({
        chat_id: CHATID,
        media: imagesChunk.map((e) => ({
          ...e,
          media: new InputFile(e.media[0], e.media[1])
        }))
      })
    })
  )
})()
```

I think `extractFiles` should not mutate payload using Object.assign as it is only suitable for single request, but not for auto-retried ones.
My very silly workaround for sendMediaGroup call is to add deep copy of input payload to avoid its mutation
```javascript
const payload = {..._payload, media: _payload.media.map(m=>({...m, media: new types_js_1.InputFile(m.media.fileData) })) };
```
in `payloadToMultipartItr` before `extractFiles` call

KnorpelSenf:
Fixes #862 by dynamically installing a `toJSON` implementation on each `InputFile`.

Note that this does not allow the same `InputFile` instance to be used in concurrent requests. That seems like an acceptable limitation for now.

For 2.0, performance can be slightly improved. This is tracked in #864.

### DIFF
--- a/src/core/payload.ts
+++ b/src/core/payload.ts
@@ -107,7 +107,7 @@ async function* payloadToMultipartItr(
     payload: Record<string, unknown>,
     boundary: string,
 ): AsyncIterableIterator<Uint8Array> {
-    const files = extractFiles(payload);
+    const files = collectFiles(payload);
     // Start multipart/form-data protocol
     yield enc.encode(`--${boundary}\r\n`);
     // Send all payload fields
@@ -116,7 +116,14 @@ async function* payloadToMultipartItr(
     for (const [key, value] of Object.entries(payload)) {
         if (value == null) continue;
         if (!first) yield separator;
-        yield valuePart(key, typeof value === "object" ? str(value) : value);
+        yield valuePart(
+            key,
+            value instanceof InputFile
+                ? value.toJSON()
+                : typeof value === "object"
+                ? str(value)
+                : value,
+        );
         first = false;
     }
     // Send all files
@@ -130,7 +137,7 @@ async function* payloadToMultipartItr(
 }
 
 /** Information about a file extracted from a payload */
-type ExtractedFile = {
+type CollectedFile = {
     /** To be used in the attach:// string */
     id: string;
     /** Hints about where the file came from, useful for filename guessing */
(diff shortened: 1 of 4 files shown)


wojpawlik (approved):

KnorpelSenf:
@all-contributors add @crwnd for review

PR #868, opened 2026-01-21
Monomorphic functions

### LINKED ISSUE #843: perf: evaluate automatic update monomorphization
Currently, the JSON received by the Bot API is highly irregular. We suspect that this causes the inline cache to be polymorphic as soon as we access data in the update object. This happens for basically all non-ignored objects, so the hypothesis here is that all context actions and getters are slower than they need to be.

We can track the shapes of objects, and install seen excess properties to make the update monomorphic. That would allow the context object to have monomorphic IC entries.

We should
1. find out whether V8 does in fact have the struggles outlined above, likely by implementing performance benchmarks and observing V8's behaviour
2. implement a step in grammY that recursively tracks all seen properties on the update objects, and installs undefined or null values on new objects
3. run the benchmarks
4. merge step 2 if our data indicates this

Some bots handle dozens of different update types, while some bots essentially only handle text messages. We should account for both extremes in our benchmarks. Most likely, bots with high load process many different types of updates.

As the diversity of update objects increases, we would collect more and more excess keys that need to be set to undefined or null. Traversing the entire update and installing dozens of empty values just to speed up 1-2 points of property access might in fact slow down the library. This is a known unknown so the above benchmarks must be good, and be taken seriously.

KnorpelSenf:
Results: https://t.me/grammyjs/344723

Fixed #843

### DIFF
--- a/bench.js
+++ b/bench.js
@@ -1,0 +1,239 @@
+import { Bot as MegaBot } from "./src_old/bot.ts";
+import { Bot as MonoBot } from "./src/bot.ts";
+
+function createBot(bot, complexity) {
+    // skip all API calls
+    bot.api.config.use(() =>
+        Promise.resolve({ ok: true, result: true })
+    );
+    if (complexity === "high") {
+        // fake plugins
+        bot.use(async (ctx, next) => {
+            Object.assign(ctx, { foo: ctx.update.update_id++ });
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
Does not improve performance very much. Abandoning.

PR #895, opened 2026-04-17
fix: resolve `ctx.from` for `managed_bot` updates

Dramex:
## Summary

Bot API 9.6 introduced the `managed_bot` update type. `ManagedBotUpdated.user` is a required `User` field per the [spec](https://core.telegram.org/bots/api#managedbotupdated). The `ctx.managedBot` alias, the `managed_bot` filter key, and the API/keyboard helpers were wired up in #892, but `ctx.from` was not updated:

- Runtime (`src/context.ts`): the `from` getter's `.user` resolution group only covers `businessConnection`, `messageReaction`, `chatBoost.boost.source`, and `removedChatBoost.source` — so inside `bot.on("managed_bot", ctx => ctx.from)` the getter returns `undefined`.
- Type-level (`src/filter.ts`): `Shortcuts.from` has no branch for `U["managed_bot"]`, so the filtered context types `ctx.from` as `undefined` too.

## Changes

- Add `this.managedBot` to the `.user` fallback chain in the `from` getter, and update its JSDoc.
- Add a matching conditional for `U["managed_bot"]["user"]` in `Shortcuts.from`.
- Add an assertion in `test/context.test.ts` that `ctx.from === update.managed_bot.user` (the `managed_bot` fixture was already there).

## Test plan

- [x] `deno task test` — 38 suites / 443 steps pass
- [x] `deno fmt --check`
- [x] `deno lint`
- [x] `deno task check`

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -559,7 +559,7 @@ export class Context implements RenamedUpdate {
     }
     /**
      * Get the user object from wherever possible. Alias for
-     * `(this.businessConnection ?? this.messageReaction ??
+     * `(this.businessConnection ?? this.messageReaction ?? this.managedBot ??
      * (this.chatBoost?.boost ?? this.removedChatBoost)?.source)?.user ??
      * (this.callbackQuery ?? this.msg ?? this.inlineQuery ??
      * this.chosenInlineResult ?? this.shippingQuery ?? this.preCheckoutQuery ??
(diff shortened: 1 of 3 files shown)


KnorpelSenf (approved):
I decided against implementing this because there's no shortcut for `managed_user.bot` so this felt inconsistent and confusing. But I now disagree with my past self, it's even more confusing to claim that `ctx.from` gets the user object from “wherever possible” and then not get it from `managed_bot` updates. So these changes are good.

KnorpelSenf:
@Dramex we usually require all contributors to sign their commits. Since this is your first time contributing to this repository, I'll make an exception, but please consider setting this up for future contributions.

KnorpelSenf:
@all-contributors add @Dramex for bug and code

Dramex:
Thanks, noted. Got SSH signing set up on my side just now, next commits in the org will go in signed.

PR #906, opened 2026-05-23
fix: remove dead code in constant-time token comparison

arunr-inji:
While reading this, I noticed an issue in `compareSecretToken` (that I introduced 😅 ):

The ternary `i < headerBytes.length ? headerBytes[i] : 0` is dead code. The length check above this already returns early if lengths differ, so `headerBytes[i]` is always safe to access directly.

### DIFF
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -2,7 +2,7 @@
     "lock": false,
     "nodeModulesDir": "none",
     "tasks": {
-        "check": "deno cache --allow-import src/mod.ts",
+        "check": "deno check --allow-import src/mod.ts",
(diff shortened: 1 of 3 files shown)


arunr-inji:
Also fixed some typecheck errors reported [here](https://github.com/grammyjs/grammY/actions/runs/26324432505/job/77499294318?pr=906) and [here](https://github.com/grammyjs/grammY/actions/runs/26324486419/job/77499443826?pr=906). 

Not sure how they were introduced 🤔

KnorpelSenf (approved):
I fixed the merge conflicts, LGTM now

PR #907, opened 2026-05-31
docs: fix broken All Contributors emoji key link

LWJerri:
The old link returns a 404 error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ You can simply do `import { Bot } from "grammy/web"`.
 
 ## Contributors ✨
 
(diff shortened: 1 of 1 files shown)


KnorpelSenf (approved):
Nice catch

KnorpelSenf:
CI just fails because Deno did their breaking change that “typically doesn't affect any projects” but we are am untypical project apparently. We need a separate PQ before this one can be merged.

KnorpelSenf:
ref https://deno.com/blog/v2.8#settimeout-and-setinterval

KnorpelSenf (approved):

PR #913, opened 2026-06-16
fix(platform): set default fetch duplex to "half" for all API requests

### LINKED ISSUE #910: Cannot upload files in browser environment
When uploading files in a browser environment, the following error occurs:

**TypeError: Failed to execute 'fetch' on 'Window': The `duplex` member must be specified for a request with a streaming body**

This is because the browser's Fetch API requires that for requests with a streaming body, the `duplex='half'` option must be explicitly specified.

## Suggestion

In the `createFormDataPayload` function in `src/core/payload.ts`, detect whether the current environment is a browser and add the `duplex: 'half'` option accordingly:

```typescript
export function createFormDataPayload(
    payload: Record<string, unknown>,
    onError: (err: unknown) => void,
) {
    const boundary = createBoundary();
    const itr = payloadToMultipartItr(payload, boundary);
    const safeItr = protectItr(itr, onError);
    const stream = itrToStream(safeItr);
    return {
        method: POST,
        headers: {
            content-type: `multipart/form-data; boundary=${boundary}`,
            connection: keep-alive,
        },
        body: stream,
        // Browsers require duplex: 'half' for streaming uploads (e.g., multipart/form-data)
        ...((typeof window !== undefined) ? { duplex: half } : {})
    };
}
```

> This has been tested in a browser environment.

## Workaround

As a workaround, a custom fetch function can be used:

```typescript
async function myFetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
    if (init?.body instanceof ReadableStream) {
        init = { ...init, ...{ duplex: half } };
    }
    return await fetch(input, init);
}
```

KnightNiwrem:
## Summary
Set `duplex: "half"` in `baseFetchConfig` defaults across all platform implementations to prevent streaming-body fetch errors in browsers.

Closes #910.

## Why
Issue [#910](https://github.com/grammyjs/grammY/issues/910) reports browser upload failures with:

- `TypeError: Failed to execute 'fetch' on 'Window': The 'duplex' member must be specified for a request with a streaming body`

Setting `duplex: "half"` is a strict WHATWG requirement if you try to pass a ReadableStream to the body of a fetch request.

## Changes
- Updated `src/platform.node.ts`:
  - Added `duplex: "half"` to all return paths of `baseFetchConfig`.
- Updated `src/platform.web.ts`:
  - `baseFetchConfig` now returns `{ duplex: "half" }`.
- Updated `src/platform.deno.ts`:
  - `baseFetchConfig` now returns `{ duplex: "half" }`.
- Updated API docs in `src/core/client.ts`:
  - `baseFetchConfig` default comment now reflects the new defaults for Node, Deno, and Web.

## Compatibility note
In browsers, `duplex` is required for streaming `fetch` bodies.  
In Deno/Node(Undici), the docs state `request.duplex` is set for streaming bodies and is treated as effectively full duplex, so adding this default has no effect on those runtimes, but would cover the case when these runtime are transpiled for the web.

- Undici docs (request.duplex behavior): [https://github.com/nodejs/undici#requestduplex](https://github.com/nodejs/undici#requestduplex)

That said, grammY currently uses an old `node-fetch` rather than Node's Undici. However, `baseFetchConfig` may still be reused in the case where user replaces `fetch` function.

---

## Disclosure
Code writing by GPT-5.3-Codex-Spark, directly commandeered by human

PQ reopened with commits signed

### DIFF
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -199,7 +199,7 @@ export interface ApiClientOptions {
     /**
      * Base configuration for `fetch` calls. Specify any additional parameters
      * to use when fetching a method of the Telegram Bot API. Default: `{
-     * compress: true }` (Node), `{}` (Deno)
+     * compress: true, duplex: "half" }` (Node), `{ duplex: "half" }` (Deno, Web)
      */
     baseFetchConfig?: Omit<
         NonNullable<Parameters<typeof fetch>[1]>,
--- a/src/platform.deno.ts
+++ b/src/platform.deno.ts
@@ -23,7 +23,7 @@ export const itrToStream = (itr: AsyncIterable<Uint8Array>) =>
     ReadableStream.from(itr);
 
 // === Base configuration for `fetch` calls
-export const baseFetchConfig = (_apiRoot: string) => ({});
+export const baseFetchConfig = (_apiRoot: string) => ({ duplex: "half" });
 
 // === Default webhook adapter
 export const defaultAdapter = "oak";
--- a/src/platform.node.ts
+++ b/src/platform.node.ts
@@ -31,6 +31,7 @@ export function baseFetchConfig(apiRoot: string) {
                 apiRoot,
                 () => new HttpsAgent({ keepAlive: true }),
             ),
+            duplex: "half",
         };
     } else if (apiRoot.startsWith("http:")) {
         return {
@@ -39,8 +40,9 @@ export function baseFetchConfig(apiRoot: string) {
                 apiRoot,
                 () => new HttpAgent({ keepAlive: true }),
             ),
+            duplex: "half",
         };
-    } else return {};
+    } else return { duplex: "half" };
 }
 
 // === Default webhook adapter
--- a/src/platform.web.ts
+++ b/src/platform.web.ts
@@ -18,6 +18,6 @@ export const itrToStream = (itr: AsyncIterable<Uint8Array>) => {
 };
 
 // === Base configuration for `fetch` calls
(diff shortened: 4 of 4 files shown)


KnorpelSenf:
@KnightNiwrem did you test this? I remember a case where this would type-check but fail at runtime. It might have been a very early version of Deno, though, so perhaps this was fixed years ago and we're fine now?

KnightNiwrem:
> @KnightNiwrem did you test this? I remember a case where this would type-check but fail at runtime. It might have been a very early version of Deno, though, so perhaps this was fixed years ago and we're fine now?

I have tested it with Node 12.20.0, 14.13.1, and Deno 1.46.3 and 2.0.0.

What was the case where it fails at runtime though? I would be surprised to hear that an unknown field would cause a failure. Typically, an unknown config field should simply be ignored as it would not be referenced in any of the call path.

KnorpelSenf:
Good to hear. I just have this vague memory that the `duplex` property cost me half a day of debugging and then I ended up not implementing it because I couldn't make it work across all runtimes. But I trust your testing more than my memory, so let's merge.

KnorpelSenf (approved):

KnightNiwrem:
> Good to hear. I just have this vague memory that the `duplex` property cost me half a day of debugging and then I ended up not implementing it because I couldn't make it work across all runtimes. But I trust your testing more than my memory, so let's merge.

Oof. I guess we will know if anyone complains. Heh.

PR #915, opened 2026-07-03
fix: migrate Skypack CDN to esm.sh

ghassan-gaidi:
Skypack has been in maintenance mode since 2024. Migrate to esm.sh which is actively maintained and more reliable.

Changes:
- platform.deno.ts: skypack.dev/obug → esm.sh/obug
- platform.web.ts: skypack.dev/obug → esm.sh/obug


KnorpelSenf:
Not even gonna review anymore

PR #916, opened 2026-07-03
fix: add LRU-eviction limit to filter query cache

ghassan-gaidi:
The filterQueryCache Map grows unbounded without eviction. While unique filter queries per bot are few, long-running bots or dynamic filter registration could accumulate entries. Added a max-size LRU eviction strategy capped at 500 entries.


KnorpelSenf:
Alright, this is a pattern now. It seems like you really have no idea what you're doing

PR #917, opened 2026-07-03
fix: add max-depth guard to requiresFormDataUpload

ghassan-gaidi:
The recursive requiresFormDataUpload function had no depth limit. Deeply nested payloads could cause a stack overflow. Added a depth parameter with a max of 10 levels as a safety guard.


KnorpelSenf:
This is the third pull request from you that is just nonsense. I'm starting to get a feeling that I'm wasting my time

PR #921, opened 2026-07-15
feat: support Bot API 10.2

KnorpelSenf:
https://core.telegram.org/bots/api#july-14-2026

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -20,10 +20,12 @@ import {
     type InputMediaLivePhoto,
     type InputMediaPhoto,
     type InputMediaVideo,
+    type InputMediaWithoutUpload,
     type InputPaidMedia,
     type InputPollOption,
     type InputProfilePhoto,
     type InputRichMessage,
+    type InputRichMessageWithoutUpload,
     type InputStoryContent,
     type KeyboardButton,
     type LabeledPrice,
@@ -514,6 +516,10 @@ export class Context implements RenamedUpdate {
     get purchasedPaidMedia() {
         return this.update.purchased_paid_media;
     }
+    /** Alias for `ctx.update.subscription` */
+    get subscription() {
+        return this.update.subscription;
+    }
 
     // AGGREGATION SHORTCUTS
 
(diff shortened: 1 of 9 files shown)


KnorpelSenf:
11 parameterised types are still missing the `InputFile` bindings, this concerns `InputMessageContent` as well as most of the `*Cached` types

KnorpelSenf:
> 11 parameterised types are still missing the `InputFile` bindings, this concerns `InputMessageContent` as well as most of the `*Cached` types

I did a bad `grep`, it was just 2 types

src/core/api.ts
```diff
@@ -2632,6 +2635,130 @@ export class Api<R extends RawApi = RawApi> {
...
+     * @param other Optional remaining parameters, confer the official reference below
+     * @param signal Optional `AbortSignal` to cancel the request
+     *
+     * **Official reference:** https://core.telegram.org/bots/api#editephemeralmessagecaption
+     */
+    editEphemeralMessageCaption(
+        chat_id: number | string,
+        receiver_user_id: number,
+        ephemeral_message_id: number,
+        caption: string,
```
KnightNiwrem, line 2717: Caption is actually optional for this method. With this function signature, users need to pass an empty string to remove caption. Is that intended?

src/context.ts
```diff
@@ -3401,6 +3409,121 @@ export class Context implements RenamedUpdate {
...
+     * Context-aware alias for `api.editEphemeralMessageCaption`. Use this method to edit the caption of an ephemeral message. Note that it is not guaranteed that the user will receive the message edit event, especially if they are offline. On success, True is returned.
+     *
+     * @param caption New caption of the message, 0-1024 characters after entities parsing
+     * @param other Optional remaining parameters, confer the official reference below
+     * @param signal Optional `AbortSignal` to cancel the request
+     *
+     * **Official reference:** https://core.telegram.org/bots/api#editephemeralmessagecaption
+     */
+    editEphemeralMessageCaption(
+        caption: string,
```
KnightNiwrem, line 3478: Same question as the api counterpart

src/core/api.ts
KnorpelSenf: Yes. It's more explicit and leads to more readable bot code. You'll actually find more cases of this throughout the API. 

Only counterexample being reply markup edits, I think.

src/context.ts
KnorpelSenf: Same answer as above

HeeroML:
@KnorpelSenf yes, can also validate that it were only two. Initially I only found InputMessageContent, but this looks solved now.

ExposedCat (approved):

HeeroML (approved):
The caption comment is resolved already and types are changed

PR #928, opened 2026-08-05
feat: basic built-in formatting utilities

### LINKED ISSUE #669: Native support for templating messages
We have a custom message distribution service, which should send messages to a segment of our users. I'm planning to add an option for templating, for example:

```
{{username}}, you won {{amount}}{{currency}} for being awesome.
```

This is easy task to do using some templating libraries or i18n libraries, if the text don't contain formatting. But with formatting it will break. The reason for that, is that depending on the length of the variables inside the template, the offset of entities may become invalid/wrong.

Note: the message "templates" are stored as json, serialized message object which bot will get as a message.

So I have multiple questions here:

1. Are there some simple solutions I don't see? For example convert to html, replace variables. (I prefer to keep entities to debug them later easier).
2. Are there any plans to add an utility function to grammy which will fix the entities. (check the diffs and make the shifts in entities)?
3. Is this a good fit for [plugin ideas](https://github.com/grammyjs/grammY/issues/110)?
3. Do you see this as a different library?

_Thank you very much for your great work on this project. This library is one of my most beloved libraries in javascript ecosystem._

### LINKED ISSUE #722: feat: built-in support for formatted strings
We should add support for formatted strings by effectively inlining the parse-mode plugin. We should improve on the plugin, too. Details to be discussed. Possible related: https://t.me/grammyjs/229443

KnorpelSenf:
As outlined in https://t.me/grammyjs/372142.

Closes #722.

### DIFF
--- a/src/format.ts
+++ b/src/format.ts
@@ -3,6 +3,7 @@ import type { MessageEntity, User } from "./types.ts";
 export class EntityString {
     private rawText: string;
     private rawEntities: MessageEntity[];
+
     constructor(text?: string, entities?: MessageEntity[]) {
         this.rawText = text ?? "";
         this.rawEntities = entities ?? [];
@@ -12,6 +13,13 @@ export class EntityString {
         this.rawText += text;
         this.rawEntities.push(...entities);
         return this;
+    }
+    concat(other: EntityString) {
+        const off = this.rawText.length;
+        this.rawText += other.rawText;
+        const shifted = other.rawEntities
+            .map((e) => ({ ...e, offset: off + e.offset }));
+        this.rawEntities.push(...shifted);
     }
 
     plain(text: string): this {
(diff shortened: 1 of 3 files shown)


src/format.ts
```diff
@@ -0,0 +1,162 @@
+import type { MessageEntity, User } from "./types.ts";
+
+export class EntityString {
+    private rawText: string;
+    private rawEntities: MessageEntity[];
+
+    constructor(text?: string, entities?: MessageEntity[]) {
+        this.rawText = text ?? "";
+        this.rawEntities = entities ?? [];
```
wojpawlik:
```suggestion
    constructor(text: string = "", entities: MessageEntity[] = []) {
        this.rawText = text;
        this.rawEntities = entities;
```

wojpawlik (approved):

src/format.ts
KnorpelSenf: No need to annotate string. 
```suggestion
    constructor(text = "", entities: MessageEntity[] = []) {
        this.rawText = text;
        this.rawEntities = entities;
```

PR #933, opened 2026-08-07
test: migrate seed-dependent payload fixtures to snapshot testing

### LINKED ISSUE #930: Migrate tests to snapshot testing
We should stop storing the expected strings in the test source code and move towards https://docs.deno.com/runtime/test/snapshots/ again. This is because very few months, Deno updates V8 which causes the seed to change (see https://github.com/denoland/deno/issues/32093).

Snapshot testing does not solve this problem in any way, but it allows us to run the tests with `--update-snapshots` rather than going through the strings manually one-by-one, re-running the tests at every step and copying the expected strings into the actual assertions.

KnightNiwrem:
> [!NOTE]
> This PR was created autonomously by Claude Fable 5 (an AI model).

Closes #930

Every Deno V8 upgrade changes the seeded RNG output, which until now meant manually re-deriving the multipart boundaries and attach IDs inlined in `test/core/payload.test.ts` (most recently in #931, and before that in ca31785). This migrates those tests to [Deno snapshot testing](https://docs.deno.com/runtime/test/snapshots/) so the next V8 bump only needs a single `deno task test:update` run.

## Changes

- `test/core/payload.test.ts`: the two multipart/form-data tests now use Deno's native `t.assertSnapshot({ headers, body })` instead of asserting against inlined boundary/attach-ID strings. The three copy-pasted runs of the "repeatedly" test collapsed into a loop. All other assertions (`method`, the `requiresFormDataUpload` checks) are unchanged.
- `test/core/__snapshots__/payload.test.ts.snap`: generated snapshot file. It contains the **exact same boundaries and attach IDs** as the previously inlined fixtures, confirming the migration is behavior-equivalent.
- `deno.jsonc`: new `test:update` task that runs the suite with `--update-snapshots` to regenerate snapshots. The `test` task is unchanged — the native snapshot API needs no extra filesystem permissions.

An earlier revision used the `@std/testing/snapshot` module; after external review feedback this was switched to the built-in `Deno.TestContext.assertSnapshot()`, which reads the same snapshot format from the same location and is updated via the `--update-snapshots` flag that #930 explicitly asks for. The snapshot file was generated once and has been byte-identical across both APIs.

## Verification

- `deno task test`: 39 tests / 448 steps pass in assert mode against the committed snapshot file
- `deno task test:update`: runs cleanly and leaves the snapshot file unchanged
- `deno fmt --check` and `deno lint`: clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)


KnightNiwrem:
A bit comment-y, but "I guess it's fine".

KnightNiwrem:
Removed note

> [!NOTE]
> and has **not yet been reviewed by a human**.

KnorpelSenf:
Commit signing is so pre-loop era

deno.jsonc
```diff
@@ -6,6 +6,7 @@
         "check": "deno check --allow-import src/mod.ts",
         "backport": "deno --no-prompt --allow-read=. --allow-write=. https://deno.land/x/deno2node@v1.16.0/src/cli.ts tsconfig.json",
         "test": "deno test --seed=123456 --parallel --allow-import ./test/",
+        "test:update": "deno test --seed=123456 --parallel --allow-import --update-snapshots ./test/",
```
KnorpelSenf: Would it be better to run `deno task test -u` here?

KnightNiwrem: Will update as-is first.

Personally, I prefer long flags in aliases and package scripts, because it has the benefit of communicating flag intent directly (e.g. contributer X doesn't have to go check what `-u` means with `deno test -h`), while avoiding cost of having to actually type out the long flag by hand.

So perhaps `deno task test --update-snapshot` instead?

KnorpelSenf: Yep! I was rather trying to say that we could re-use the other task. Regarding short and long flags, I prefer typing short ones but writing long ones to files; which is a long way of saying I agree with you.

KnorpelSenf (approved):
Good stuff

KnorpelSenf:
I'll ignore missing signatures and merge anyway

PR #939, opened 2026-08-14
test: reproduce TransformerComposer crash on unmatched methods

KnightNiwrem:
> [!NOTE]

## Summary

This test-only draft PR adds two regression tests for `TransformerComposer.on`:

- a direct test of the composed transformer function
- an integration-level test that installs the composer on a public `Api` instance and calls `api.getMe()` with `fetch` stubbed

Both tests scope a transformer to `sendMessage`, then invoke the nonmatching `getMe` method. A nonmatching call should bypass the scoped transformer and continue to the previous API call function. Through `Api`, it should proceed to the stubbed fetch and return the Bot API result.

## Current result

Both paths fail before the previous function or stubbed fetch is reached:

```text
```

The public-API stack passes through `Api.getMe` and `ApiClient.callApi` before failing in `src/transform.ts`.

## Purpose

This PR intentionally does not contain a production fix. It is a minimal reproducer so CI can record the failure and the behavior can be discussed separately. It should not be merged in its current failing state.

### DIFF
--- a/test/transform.test.ts
+++ b/test/transform.test.ts
@@ -1,11 +1,57 @@
+import { Api } from "../src/api.ts";
 import type { ApiCallFn } from "../src/client.ts";
 import { TransformerComposer } from "../src/transform.ts";
-import { assertEquals, describe, it } from "./deps.test.ts";
+import type { UserFromGetMe } from "../src/types.ts";
+import {
+    assertEquals,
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
Amazing contributions, both of them good bug reports

PR #943, opened 2026-08-14
fix: allow videoHtml options to be omitted

### LINKED ISSUE #942: InlineQueryResultBuilder.videoHtml incorrectly requires an options argument
> [!NOTE]

## Summary

In `@grammyjs/grammy@2.0.0-beta.8` and the current `v2` branch, the public signature of `InlineQueryResultBuilder.videoHtml` requires an `options_` argument even though the implementation defaults that argument to `{}`.

Consequently, the natural four-argument call is valid at runtime but rejected by TypeScript.

## Minimal reproduction

```ts
import { InlineQueryResultBuilder } from "@grammyjs/grammy";

const result = InlineQueryResultBuilder.videoHtml(
    "id",
    "Embedded video",
    "https://example.com/video",
    "https://example.com/thumbnail.jpg",
);
```

Type checking fails with:

```text
TS2554: Expected 5 arguments, but got 4.

An argument for 'options_' was not provided.
```

No Telegram API request is needed to reproduce this.

## Expected behavior

The four-argument call should type-check. The options object should be optional, consistent with the runtime default and the other inline-query builder methods that default their options to `{}`.

## Actual behavior

TypeScript requires a fifth argument. Supplying `{}` manually avoids the diagnostic:

```ts
InlineQueryResultBuilder.videoHtml(
    "id",
    "Embedded video",
    "https://example.com/video",
    "https://example.com/thumbnail.jpg",
    {},
);
```

## Suspected cause

The exported interface declares `options_` without `?`:

```ts
videoHtml(
    id: string,
    title: string,
    video_url: string | URL,
    thumbnail_url: string | URL,
    options_: InlineQueryResultOptions<
        InlineQueryResultVideo,
        "mime_type" | "title" | "video_url" | "thumbnail_url"
    >,
): WithInputMessageMethods<InlineQueryResultVideo>;
```

Source: https://github.com/grammyjs/grammY/blob/v2/src/convenience/inline_query.ts#L476-L485

The implementation, however, supplies a default:

```ts
videoHtml(id, title, video_url, thumbnail_url, options = {}) {
    // ...
}
```

Source: https://github.com/grammyjs/grammY/blob/v2/src/convenience/inline_query.ts#L688-L702

The interface parameter likely only needs to be marked optional, matching the implementation.

This was found during follow-up analysis of #940.

roshnicdave:
Fixes #942.

Make the `videoHtml` type match its existing default for `options`, with coverage for the four-argument call.

### DIFF
--- a/src/convenience/inline_query.ts
+++ b/src/convenience/inline_query.ts
@@ -478,7 +478,7 @@ export interface InlineQueryResultBuilder {
         title: string,
         video_url: string | URL,
         thumbnail_url: string | URL,
-        options_: InlineQueryResultOptions<
+        options_?: InlineQueryResultOptions<
             InlineQueryResultVideo,
             "mime_type" | "title" | "video_url" | "thumbnail_url"
         >,
--- a/test/convenience/inline_query.test.ts
+++ b/test/convenience/inline_query.test.ts
@@ -1767,6 +1767,27 @@ describe("InlineQueryResultBuilder", () => {
                     },
                 });
             });
+            it("should build an HTML InlineQueryResultVideo without options", () => {
+                const video = InlineQueryResultBuilder.videoHtml(
+                    "id",
+                    "title",
+                    "https://grammy.dev/",
+                    "https://grammy.dev/thumb",
+                ).text("#Text", { parse_mode: "Markdown" });
+
+                assertObjectMatch(video, {
+                    type: "video",
+                    mime_type: "text/html",
+                    id: "id",
+                    title: "title",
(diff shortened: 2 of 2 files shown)


KnorpelSenf (approved):

KnorpelSenf:
@all-contributors add @roshda for bug and test!

KnorpelSenf:
Oops

KnorpelSenf:
@all-contributors add @roshda for code and test

KnorpelSenf:
Sorry for the mess. Thanks for the fix!

PR #948, opened 2026-08-15
fix: read Azure secret header from request on v2

KnightNiwrem:
> [!NOTE]
> This pull request was primarily written by GPT-5.6 Sol xhigh in ChatGPT Work.

## Summary

Ports the Azure Functions v3 secret-header fix from main-branch PR #896 to the `v2` branch.

The v2 Azure adapter reads `X-Telegram-Bot-Api-Secret-Token` from `context.res.headers`, which represents response state, instead of reading the incoming request. As a result, a correctly authenticated webhook request is rejected with HTTP 401.

## Changes

- Add the optional incoming `headers` map to `WebhookAdapterAzure`'s request type.
- Read the lowercased secret-token header from `request.headers`.
- Keep the port source-only, matching the scope of the main-branch fix.

## Main-branch fix

This is the v2 adaptation of:

- #896

The behavior is intentionally the same as the merged main-branch fix, adapted to v2's combined webhook implementation in `src/convenience/webhook.ts`.

## Verification

- Formatting passes for the changed source file.
- Linting passes for the changed source file.
- All v2 source files type-check with the local debug import shim.

### DIFF
--- a/src/convenience/webhook.ts
+++ b/src/convenience/webhook.ts
@@ -318,6 +318,7 @@ export type WebhookAdapterAzure = (context: {
     };
 }, request: {
     body?: unknown;
+    headers?: Record<string, string | undefined>;
     method: string;
     url: string;
 }) => ReqResHandler;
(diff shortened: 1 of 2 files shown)


KnorpelSenf:
Thanks for the port.

This test adds no value, though. The issue is that this repo wasn't in sync with AWS. This library is only correct if we happen to be in sync. That cannot be fixed by duplicating the places we have to keep in sync. 

There's a reason why the original changes in #896 contain no test. Please remove it again.

KnorpelSenf (approved):
Nice, thanks

PR #956, opened 2026-08-17
fix: handle undefined falseTransfomer on TransformerComposer

### LINKED ISSUE #704: feat: support Response objecs on Node in InputFile
We should bump the Node version and add native support for web standards as far as they are implemented in Node.js.

### LINKED ISSUE #705: feat: support base64 strings in InputFile
Title says it all. Should just be `new InputFile({ base64 })`.

### LINKED ISSUE #940: TransformerComposer.on crashes for nonmatching API methods
> [!NOTE]

## Summary

In `@grammyjs/grammy@2.0.0-beta.8` and the current `v2` branch, a transformer scoped with `TransformerComposer.on` crashes when it receives a nonmatching Bot API method.

A matching method runs successfully. A nonmatching method should skip the scoped transformer and continue to the previous API call function, but instead throws:

```text
```

## CI reproducer

Draft test-only PR: https://github.com/grammyjs/grammY/pull/939

The PR deliberately contains only regression tests so CI can display the current failure. It reproduces the bug both by invoking the composed transformer directly and by attaching it to a public `Api` instance and calling the unscoped `api.getMe()` method with `fetch` stubbed.

## Minimal reproduction

```ts
const scoped = composer.on("sendMessage");
scoped.use((prev, data, signal) => prev(data, signal));

const transform = composer.transformer();

// The matching method succeeds.
await transform(previous, {
    method: "sendMessage",
    payload: { chat_id: 1, text: "test" },
});

// The nonmatching method throws instead of calling `previous`.
await transform(previous, {
    method: "getMe",
    payload: {},
});
```

No Telegram API request is needed to reproduce the exception.

## Suspected cause

`.on` delegates to `.filter`, which calls `branch` without a false transformer:

```ts
this.branch(predicate, composer);
```

In `branch`, the omitted argument is handled as follows:

```ts
const otherwise = Array.isArray(falseTransformer)
    ? new TransformerComposer(...falseTransformer)
    : new TransformerComposer(falseTransformer);
```

Because `falseTransformer` is `undefined`, this constructs:

```ts
new TransformerComposer(undefined)
```

That is different from an empty, pass-through composer. The constructor receives one element, so it applies `flatten` to `undefined`. The resulting closure does not throw immediately, but when a predicate is false and the false branch runs, it eventually evaluates:

```ts
tf.transformer()
```

where `tf` is `undefined`.

This explains why matching methods work while nonmatching methods crash. The omitted false branch appears to need pass-through behavior, such as an empty `TransformerComposer`, rather than a composer containing `undefined`.

Because `.filter` uses the same omitted false branch, the underlying problem may affect more than the `.on` convenience method.

roziscoding:
Fixes #940

Includes work from #939

### DIFF
--- a/src/transform.ts
+++ b/src/transform.ts
@@ -195,7 +195,7 @@ export class TransformerComposer<
             boolean | ((data: D, signal?: AbortSignal) => boolean)
         >,
         trueTransformer: MaybeArray<Transformer<R, D>>,
-        falseTransformer: MaybeArray<Transformer<R, D>>,
+        falseTransformer: MaybeArray<Transformer<R, D>> = [],
     ) {
         const then = Array.isArray(trueTransformer)
             ? new TransformerComposer(...trueTransformer)
--- a/test/transform.test.ts
+++ b/test/transform.test.ts
@@ -1,0 +1,218 @@
+import { Api } from "../src/api.ts";
+import type { ApiCallFn } from "../src/client.ts";
+import { TransformerComposer } from "../src/transform.ts";
+import type { TransformerObj } from "../src/transform.ts";
+import type { UserFromGetMe } from "../src/types.ts";
+import {
+    assertEquals,
+    assertSpyCalls,
+    describe,
+    it,
+    stub,
+} from "./deps.test.ts";
+
+const me: UserFromGetMe = {
+    id: 123456789,
+    is_bot: true,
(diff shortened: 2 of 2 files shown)


KnightNiwrem (approved):
Approved the code. Test not checked.

PR #957, opened 2026-08-17
fix: accept an array of methods on TransformerComposer#on

### LINKED ISSUE #941: TransformerComposer.on supports method arrays at runtime but rejects them in its type
> [!NOTE]

## Summary

In `@grammyjs/grammy@2.0.0-beta.8` and the current `v2` branch, `TransformerComposer.on` handles an array of Bot API methods at runtime, but its public TypeScript signature accepts only a single method.

This means the array branch that is explicitly implemented cannot be used without a type assertion.

## Minimal reproduction

```ts
import { TransformerComposer } from "@grammyjs/grammy";

composer.on(["sendMessage", "getMe"]);
```

Type checking fails with an error similar to:

```text
TS2345: Argument of type 'string[]' is not assignable to parameter of type
'"getUpdates" | "setWebhook" | ... | "getMe" | "sendMessage" | ...'.
```

No Telegram API request is needed to reproduce this.

## Expected behavior

The array form should type-check because the implementation explicitly supports it, allowing one scoped composer to match any method in the array.

## Actual behavior

The implementation reaches its `Array.isArray(method)` branch at runtime, but TypeScript rejects the call before it can be used normally.

## Suspected cause

The parameter is declared as a scalar `M`:

```ts
on<M extends D["method"]>(
    method: M,
    ...transformers: Array<Transformer<R, Extract<D, { method: M }>>>
)
```

However, the implementation immediately handles an array:

```ts
return Array.isArray(method)
    ? this.filter(
        (data): data is Extract<D, { method: M }> =>
            method.includes(data.method),
        ...transformers,
    )
    : this.filter(/* ... */);
```

Source: https://github.com/grammyjs/grammY/blob/v2/src/transform.ts#L78-L92

The public signature likely needs an array overload or a suitable single-or-array method type while preserving correct narrowing for the returned composer.

This was found during follow-up analysis of #940.

roziscoding:
Fixes #941

Includes work from #956

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/src/transform.ts
+++ b/src/transform.ts
@@ -76,20 +76,18 @@ export class TransformerComposer<
     }
 
     on<M extends D["method"]>(
-        method: M,
+        method: MaybeArray<M>,
         ...transformers: Array<Transformer<R, Extract<D, { method: M }>>>
     ): TransformerComposer<R, Extract<D, { method: M }>> {
-        return Array.isArray(method)
-            ? this.filter(
-                (data): data is Extract<D, { method: M }> =>
-                    method.includes(data.method),
-                ...transformers,
-            )
-            : this.filter(
-                (data): data is Extract<D, { method: M }> =>
-                    data.method === method,
-                ...transformers,
-            );
+        const methods: Array<D["method"]> = Array.isArray(method)
+            ? method
+            : [method];
+
+        return this.filter(
+            (data): data is Extract<D, { method: M }> =>
+                methods.includes(data.method),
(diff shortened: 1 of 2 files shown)


KnightNiwrem (approved):
Approved the code. Test not checked.

PR #959, opened 2026-08-17
fix: make polling startup cancellation-safe

roshnicdave:
Fixes cancellation races between start, stop, and shared bot initialization

-Scope polling controllers and stop confirmation to each run
-Cancel pending setup/retry work and ignore stale polling results after a restart
- Preserve setup failures and shared initialization behavior for concurrent callers

### DIFF
--- a/test/bot.test.ts
+++ b/test/bot.test.ts
@@ -463,6 +463,41 @@ describe("Bot error handling", () => {
             assertEquals(getUpdatesStub.calls.length, 3);
         });
 
+        it("should cancel retry delay when stopped", async () => {
+            using time = new FakeTime();
+            let callCount = 0;
+            using getUpdatesStub = stub(
+                bot.api,
+                "getUpdates",
+                () => {
+                    callCount++;
+                    if (callCount === 1) {
+                        return Promise.reject(
+                            new GrammyError(
+                                "Too Many Requests",
+                                {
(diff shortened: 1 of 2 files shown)


test/bot.test.ts
```diff
@@ -215,6 +239,48 @@ describe("Bot initialization", () => {
...
+        using _ = stub(bot.api, "getMe", () => {
+            callCount++;
+            if (callCount === 1) {
+                throw new GrammyError(
+                    "Too Many Requests",
+                    {
+                        ok: false,
+                        error_code: 429,
+                        description: "Too Many Requests",
+                        parameters: { retry_after: 0 },
```
KnightNiwrem: How is this test meaningfully different from the test immediately above it?

If the previous test has established that retry occurs - even in the absence of parameters.retry_after for HTTP 429 - what will we be trying to establish with this test with a parameters.retry_after value of 0?

roshnicdave: sorry, forgot to look over tests! can you review again 🙏

KnorpelSenf:
I have no idea which issue you ran into with your bot, and this seems to be just a slop contribution. I don't see why I should spend my free time on a review if you don't spend yours on the quality of your contributions. Closing.

Feel free to join the chat and convince us otherwise if you actually care about this.

PR #960, opened 2026-08-17
fix: correct retry backoff units

roshnicdave:
Fixes generic retry backoff delays being treated as seconds. Telegram `retry_after` values remain in seconds.

Tests: `deno task test`, `deno task check`, `deno fmt --check`, `deno lint`.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -673,7 +673,7 @@ async function withRetries<T>(
         if (delay) {
             // Do not sleep for the first retry
             if (lastDelay !== INITIAL_DELAY) {
-                await sleep(lastDelay, signal);
+                await sleep(lastDelay / 1000, signal);
             }
             const TWENTY_MINUTES = 20 * 60 * 1000; // ms
             lastDelay = Math.min(TWENTY_MINUTES, 2 * lastDelay);
--- a/test/bot.test.ts
+++ b/test/bot.test.ts
@@ -162,6 +162,32 @@ describe("Bot initialization", () => {
         assertEquals(callCount, 2);
     });
 
+    it("should back off in milliseconds after repeated HttpErrors", async () => {
+        using time = new FakeTime();
+        const bot = new Bot(token);
+        const controller = new AbortController();
+        let callCount = 0;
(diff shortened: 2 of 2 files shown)


frederico-kluser:
Ran into this same bug independently and opened #961 (issue, with a reproduction) and #962 (PR) before finding this one — my mistake, I searched issues but not pull requests. I have closed #962; this PR was first and is the one that should land.

Two small things from mine that might be worth folding in here, entirely up to you and @KnorpelSenf:

**1. The docstring ceiling no longer matches the constant.** `withRetries` says:

> Never waits longer than 1 hour before retrying.

Before the fix that was wrong in the other direction (the ceiling was effectively ~13.9 days). After the fix the real ceiling is 20 minutes, matching the named constant `TWENTY_MINUTES`. So the prose is now off by 3×. Either the docstring should say 20 minutes, or `TWENTY_MINUTES` was meant to be an hour and the constant should change — a maintainer call, but it is currently inconsistent either way.

**2. Why the suite never caught it**, which might be worth a line in the PR body: the two existing retry tests assert `callCount === 2`, i.e. the initial attempt plus the *first* retry — and the first retry is deliberately immediate, since `if (lastDelay !== INITIAL_DELAY)` is false on the first pass. So `sleep` was never reached by any test. Your new test is the first one that drives a second retry. (The `should retry on 429 errors` test does advance a fake clock by `30000` for `retry_after: 30`, which is consistent with `sleep` taking seconds and passes because that call site was already correct.)

#961 has a self-contained reproduction that needs no Telegram token and no network access to `api.telegram.org` — it points `client.apiRoot` at a local server that destroys every connection. Measured on 1.45.1:

```
t=   0.01s  method=getMe         gap=   0.00s     <- 1st retry, immediate
t= 100.02s  method=getMe         gap= 100.00s     <- 2nd retry: 100 seconds
```

Happy to help however is useful — reviewing, testing, or nothing at all.

KnorpelSenf:
This is an ugly fix. Instead of converting the units back and forth several times, we should instead refactor the code to use time units consistently everywhere.

Also, the tests are just slop. I will do this myself.

PR #962, opened 2026-08-20
fix: convert backoff delay to seconds when retrying

### LINKED ISSUE #961: withRetries passes a millisecond backoff to sleep(), which expects seconds (1000x too long)
## Summary

`withRetries` in `src/bot.ts` tracks its backoff delay in **milliseconds** but passes it to `sleep`, which takes **seconds**. Every backoff wait is therefore **1000× longer than intended**.

The second retry waits **100 seconds** instead of the documented 100 ms, and the `TWENTY_MINUTES` ceiling, interpreted as seconds, becomes roughly **13.9 days**.

This affects `bot.init()` / `getMe` at startup and the long polling loop, on any `HttpError` (network failure) or any 5xx from the Bot API.

## Where

`src/bot.ts`, in `withRetries`:

```ts
const INITIAL_DELAY = 50; // ms
let lastDelay = INITIAL_DELAY;
...
if (delay) {
    // Do not sleep for the first retry
    if (lastDelay !== INITIAL_DELAY) {
        await sleep(lastDelay, signal);          // <-- ms passed to a seconds API
    }
    const TWENTY_MINUTES = 20 * 60 * 1000; // ms
    lastDelay = Math.min(TWENTY_MINUTES, 2 * lastDelay);
}
```

and

```ts
/**
 * Returns a new promise that resolves after the specified number of seconds, or
 * rejects as soon as the given signal is aborted.
 */
async function sleep(seconds: number, signal?: AbortSignal) {
    ...
    handle = setTimeout(res, 1000 * seconds);
}
```

## Why this is a units bug and not intended behaviour

Three independent places in the code state the intent, and all three disagree with what happens:

1. Both constants carry an explicit `// ms` comment.
2. The `withRetries` docstring says: *"If second attempt fails, too, waits for **100 ms**, and then doubles this delay for every subsequent attempt. Never waits longer than **1 hour** before retrying."*
3. The `sleep` docstring says it takes **seconds** — and that is genuinely its contract, since the 429 branch calls `await sleep(retryAfter, signal)` where the Bot API's `retry_after` is specified in seconds. The same is true of `sleep(sleepSeconds)` in the polling loop above.

So `sleep` is correct; the two call sites that pass a millisecond value are not.

## Observed vs documented

| retry | actual wait | documented |
| ----- | ----------- | ---------- |
| 1st | immediate | immediate |
| 2nd | **100 s** | 100 ms |
| 3rd | 200 s | 200 ms |
| 12th | ~1.2 days | ~102 s |
| ceiling | **~13.9 days** | 1 hour |

## Reproduction

No Telegram credentials and no network access to `api.telegram.org` are needed: the script starts a local server on an ephemeral port that destroys every connection, which is what makes grammY raise `HttpError`.

```js
// repro.mjs  —  node repro.mjs
import http from "node:http";
import { Bot } from "grammy";

const t0 = Date.now();
let previous = null;

const server = http.createServer((req, res) => {
  const now = Date.now() - t0;
  const method = (req.url ?? "").split("/").pop();
  const gap = previous === null ? 0 : now - previous;
  previous = now;
  console.log(
    `t=${(now / 1000).toFixed(2).padStart(7)}s  method=${method.padEnd(14)}` +
      `gap=${(gap / 1000).toFixed(2).padStart(7)}s`,
  );
  res.destroy();
});

server.listen(0, "127.0.0.1", () => {
  const { port } = server.address();
  const bot = new Bot("123456:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA", {
    client: { apiRoot: `http://127.0.0.1:${port}` },
  });
  bot.catch(() => {});
  void bot.start();
});

setTimeout(() => process.exit(0), 135_000);
```

Output on `grammy@1.45.1`, Node v24.15.0:

```
t=   0.01s  method=deleteWebhook gap=   0.00s
t=   0.01s  method=getMe         gap=   0.00s     <- 1st retry, immediate (as documented)
t=   0.01s  method=deleteWebhook gap=   0.00s
t= 100.02s  method=getMe         gap= 100.00s     <- 2nd retry: 100 SECONDS
t= 100.02s  method=deleteWebhook gap=   0.00s
```

## Impact

The failure mode is quiet, which is what makes it expensive to diagnose. During the 100 s wait the process stays alive, `bot.start()` has not resolved or rejected, and nothing is written to stderr — the retry error goes to `debugErr`, which is silent unless `DEBUG` is set. From the outside, the bot has simply stopped responding.

A transient network blip at startup therefore costs at least 100 s of silence, and a longer outage escalates into waits of hours and then days, well past the point where an operator would assume the process is wedged and restart it.

## Why the current tests don't catch it

The existing retry tests (`should retry on HttpError during initialization`, `should retry on 5xx errors during initialization`) assert `callCount === 2`, i.e. the initial attempt plus the **first** retry. The first retry is deliberately immediate — `if (lastDelay !== INITIAL_DELAY)` is false on the first pass — so no `sleep` is ever reached. The delay only appears on the **second** retry, and no test exercises it.

The `should retry on 429 errors` test does advance a `FakeTime` clock by `30000` for `retry_after: 30`, which is consistent with `sleep` taking seconds, and passes because that call site is correct.

## Suggested fix

Convert at the call site, keeping the delay bookkeeping in milliseconds as the comments describe:

```ts
if (lastDelay !== INITIAL_DELAY) {
    // `sleep` expects seconds, but the backoff is tracked in
    // milliseconds, so it has to be converted here
    await sleep(lastDelay / 1000, signal);
}
```

One secondary point worth a maintainer's decision: after the units are fixed, the ceiling becomes 20 minutes, matching the named constant `TWENTY_MINUTES`, while the docstring says "1 hour". I have assumed the constant is authoritative and aligned the docstring to it, but if 1 hour was the intent then the constant should change instead.

I am opening a pull request with this fix and a regression test that fails before it and passes after.

---

Environment: `grammy@1.45.1` (current `latest` on npm, and the same code is present on `main`), Node v24.15.0, Linux x86_64.

frederico-kluser:
Fixes #961.

## The bug

`withRetries` tracks its backoff in milliseconds — both `INITIAL_DELAY` and `TWENTY_MINUTES` say `// ms` — but passed that value to `sleep`, which takes seconds. Every backoff wait was 1000× longer than intended: the second retry waited **100 seconds** instead of the documented 100 ms, and the ceiling became roughly **13.9 days** instead of the documented hour.

That `sleep` really does take seconds is settled by its neighbours: the 429 branch a few lines above passes the Bot API's `retry_after` to it directly, and the polling loop passes `sleepSeconds`. Both are correct, so the conversion belongs at this one call site rather than in `sleep`.

## The change

```diff
             if (lastDelay !== INITIAL_DELAY) {
-                await sleep(lastDelay, signal);
+                // `sleep` expects seconds, but the backoff is tracked in
+                // milliseconds, so it has to be converted here
+                await sleep(lastDelay / 1000, signal);
             }
```

Plus a regression test, and one docstring line.

## Why it went unnoticed

The two existing retry tests assert `callCount === 2` — the initial attempt plus the **first** retry. The first retry is deliberately immediate, since `if (lastDelay !== INITIAL_DELAY)` is false on the first pass, so the suite never reached `sleep` at all. The delay only appears on the second retry.

The new test drives a second retry under `FakeTime` and advances the clock by exactly 100 ms. It fails on `main` (`AssertionError: Values are not equal`, in 20 ms — it does not hang) and passes with this change.

## A question for you

Once the units are fixed, the ceiling is 20 minutes, matching the named constant, while the docstring said "1 hour". I assumed the constant is authoritative and aligned the prose to it. If an hour was the intent, then `TWENTY_MINUTES` should change instead and I am happy to switch it — just say which.

## Checks

`deno task dev` passes on this branch: `deno fmt` (59 files), `deno lint` (45 files), `deno task test` (**39 passed, 449 steps, 0 failed**), `deno task check`. `deno fmt` left everything but the two edited files untouched.

Reproduction script and measured output are in #961.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -627,8 +627,8 @@ a known bot info object.",
  *
  * Otherwise, if the first attempt at running the task fails, the task is
  * retried immediately. If second attempt fails, too, waits for 100 ms, and then
- * doubles this delay for every subsequent attempt. Never waits longer than 1
- * hour before retrying.
+ * doubles this delay for every subsequent attempt. Never waits longer than 20
+ * minutes before retrying.
  *
  * @param task Async task to perform
  * @param signal Optional `AbortSignal` to prevent further retries
@@ -673,7 +673,9 @@ async function withRetries<T>(
         if (delay) {
             // Do not sleep for the first retry
             if (lastDelay !== INITIAL_DELAY) {
-                await sleep(lastDelay, signal);
+                // `sleep` expects seconds, but the backoff is tracked in
+                // milliseconds, so it has to be converted here
+                await sleep(lastDelay / 1000, signal);
             }
             const TWENTY_MINUTES = 20 * 60 * 1000; // ms
             lastDelay = Math.min(TWENTY_MINUTES, 2 * lastDelay);
--- a/test/bot.test.ts
+++ b/test/bot.test.ts
@@ -162,6 +162,35 @@ describe("Bot initialization", () => {
         assertEquals(callCount, 2);
     });
 
+    it("should back off in milliseconds between retries", async () => {
+        using time = new FakeTime();
+        const bot = new Bot(token);
+        let callCount = 0;
+        using _ = stub(bot.api, "getMe", () => {
+            callCount++;
+            if (callCount <= 2) {
+                return Promise.reject(
+                    new HttpError("Network error", { error: "ECONNREFUSED" }),
+                );
+            }
+            return Promise.resolve(botInfo);
+        });
+
+        const initPromise = bot.init();
+        // The first retry is immediate, so no timer is involved yet
+        await time.runMicrotasks();
+        assertEquals(callCount, 2);
+
+        // The second retry waits for the initial delay, which is 100 ms
+        await time.tickAsync(100);
+        await time.nextAsync();
+        // Initial attempt + 2 retries
+        assertEquals(callCount, 3);
+
+        await initPromise;
+        assertEquals(bot.isInited(), true);
+    });
+
     it("should retry on 5xx errors during initialization", async () => {
         const bot = new Bot(token);
         let callCount = 0;


KnightNiwrem:
Is there any reason to create this PR when https://github.com/grammyjs/grammY/pull/960 already exist?

To be clear, I'm not saying that we shouldn't have multiple PRs that attempt to solve the same issue. But I would expect them to be at least qualitatively different in the approach they take - at least, to the extent that simply leaving a comment for discussion in the existing PR is insufficient.

frederico-kluser:
Closing this as a duplicate of #960, which proposes the same fix and was opened three days earlier. I missed it before opening this — my search covered issues but not pull requests, which is entirely on me. Apologies for the noise.

@roshnicdave got there first and their PR should be the one that lands.

I have left the two small points that this PR carried in a comment on #960, and #961 stays open with a runnable reproduction and the impact analysis, in case that is useful when reviewing.

PR #963, opened 2026-08-23
fix: increase robustness for future error handling changes

### LINKED ISSUE #961: withRetries passes a millisecond backoff to sleep(), which expects seconds (1000x too long)

KnorpelSenf:
As far as we know, this does not fix an actual bug. However, it's still nice to use correct time units even on dead code paths.

Fixes #961.

Closes #960.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -615,7 +615,7 @@ a known bot info object.",
         debugErr(
             `Call to getUpdates failed, retrying in ${sleepSeconds} seconds ...`,
         );
-        await sleep(sleepSeconds);
+        await sleep(1000 * sleepSeconds);
     }
 }
 
@@ -658,10 +658,10 @@ async function withRetries<T>(
                 delay = true;
                 strategy = "retry";
             } else if (error.error_code === 429) {
-                const retryAfter = error.parameters.retry_after;
-                if (typeof retryAfter === "number") {
+                const retryAfterSeconds = error.parameters.retry_after;
+                if (typeof retryAfterSeconds === "number") {
                     // ignore the backoff for sleep, then reset it
-                    await sleep(retryAfter, signal);
+                    await sleep(1000 * retryAfterSeconds, signal);
                     lastDelay = INITIAL_DELAY;
                 } else {
                     delay = true;
@@ -702,10 +702,10 @@ async function withRetries<T>(
 }
 
 /**
- * Returns a new promise that resolves after the specified number of seconds, or
- * rejects as soon as the given signal is aborted.
+ * Returns a new promise that resolves after the specified number of
+ * milliseconds, or rejects as soon as the given signal is aborted.
  */
-async function sleep(seconds: number, signal?: AbortSignal) {
+async function sleep(milliseconds: number, signal?: AbortSignal) {
     let handle: Parameters<typeof clearTimeout>[0];
     let reject: ((err: Error) => void) | undefined;
     function abort() {
@@ -720,7 +720,7 @@ async function sleep(seconds: number, signal?: AbortSignal) {
                 return;
             }
             signal?.addEventListener("abort", abort);
-            handle = setTimeout(res, 1000 * seconds);
+            handle = setTimeout(res, milliseconds);
         });
     } finally {
         signal?.removeEventListener("abort", abort);
--- a/test/bot.test.ts
+++ b/test/bot.test.ts
@@ -189,11 +189,12 @@ describe("Bot initialization", () => {
     });
 
     it("should handle 429 without retry_after parameter", async () => {
+        using time = new FakeTime();
         const bot = new Bot(token);
         let callCount = 0;
         using _ = stub(bot.api, "getMe", () => {
             callCount++;
-            if (callCount === 1) {
+            if (callCount <= 2) {
                 throw new GrammyError(
                     "Too Many Requests",
                     {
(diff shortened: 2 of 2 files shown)


KnightNiwrem (approved):
Not dead code paths. Video sent to DM instead of here - don't want to accidentally leak anything from my screen. Approved.

KnorpelSenf:
In fact, this did fix an actual bug for bots that are unable to reach TG in the first place.

PR #964, opened 2026-08-24
feat: support Bot API 10.3

KnorpelSenf:
https://core.telegram.org/bots/api#august-24-2026

PR #966, opened 2026-08-25
fix: preserve Bot API 10.3 argument semantics

KnightNiwrem:
> [!NOTE]

## Summary

Address two issues found while reviewing #964.

This change:

- preserves `force_reply` when cloning, transposing, reflowing, or copying `Keyboard` and `InlineKeyboard`
- excludes both `text` and `rich_message` from the remaining options for `editEphemeralMessageText`

## Why

Bot API 10.3 adds `force_reply` to reply and inline keyboard markup. The new value was not copied by the keyboard clone paths, so it was lost when a keyboard was cloned, transposed, reflowed, or passed to `from`.

`editEphemeralMessageText` accepts either text or a rich message through its dedicated argument. Its remaining-options type omitted `text`, but not `rich_message`, allowing the options object to override the selected payload or include both fields.

## Verification

- `deno fmt --check` on all changed files
- `npm run backport`
- focused runtime probe against the backported Node output: `force_reply` is preserved across the keyboard copy paths
- focused `deno check` against the current `@grammyjs/types` main branch

The full Deno test entrypoint could not load its remote JSR and Skypack dependencies in the local environment; CI can run the complete suite.

### DIFF
--- a/src/convenience/keyboard.ts
+++ b/src/convenience/keyboard.ts
@@ -636,6 +636,7 @@ export class Keyboard {
         clone.one_time_keyboard = this.one_time_keyboard;
         clone.resize_keyboard = this.resize_keyboard;
         clone.input_field_placeholder = this.input_field_placeholder;
+        clone.force_reply = this.force_reply;
         return clone;
     }
     /**
@@ -1258,7 +1259,7 @@ export class InlineKeyboard {
     toTransposed() {
         const original = this.inline_keyboard;
         const transposed = transpose(original);
-        return new InlineKeyboard(transposed);
+        return this.clone(transposed);
     }
     /**
      * Creates a new inline keyboard with the same buttons but reflowed into a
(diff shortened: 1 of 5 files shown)


src/convenience/keyboard.ts
```diff
@@ -1298,15 +1299,23 @@ export class InlineKeyboard {
...
+     *
+     * Optionally takes a new grid of buttons to replace the current buttons. If
+     * specified, only the options will be cloned, and the given buttons will be
+     * used instead.
      */
-    clone() {
-        return new InlineKeyboard(
-            this.inline_keyboard.map((row) => row.slice()),
+    clone(
+        inline_keyboard: InlineKeyboardButton[][] = this.inline_keyboard,
```
KnightNiwrem, line 1312: I am not typically a fan of changing function signatures. But here, it seems to be the cleanest way, if transpose and flowed wants to use clone. So I'm tentatively.. fine with it

KnorpelSenf: I agree with both takes.

KnorpelSenf (approved):
Good changes

PR #968, opened 2026-09-10
Pass unmatched calls through when branch has no fallback

### LINKED ISSUE #940: TransformerComposer.on crashes for nonmatching API methods

soroush5:
Fixes #940. (The repro in draft #939 should now pass with this change.)

`composer.on("sendMessage")` crashed with `TypeError: Cannot read properties of undefined (reading 'transformer')` for any nonmatching method. Root cause is in `branch()`: `filter()` calls it without a fallback, and it built `new TransformerComposer(falseTransformer)` with `falseTransformer === undefined` — a one-element chain whose `flatten` step blows up on `undefined`.

The fix uses an empty composer (identity pass-through) when no fallback is given, so unmatched calls flow to `previous` untouched.

How I verified: new `test/transform.test.ts` — a nonmatching `getMe` reaches `previous` exactly once (fails before with the exact reported TypeError, passes after), plus a matching-method test proving scoped transformers still run. Neighbors (`composer`, `filter`) green: 5 files / 78 steps, `deno lint` + `deno fmt --check` clean.


soroush5:
You're right, thanks for pointing it out — #956 was already open and covers this. Closing mine to keep the review queue clean; happy to help test or review #956 instead.

PR #972, opened 2026-09-17
fix: reject HTTP error responses in InputFile downloads

### LINKED ISSUE #971: InputFile uploads HTTP 4xx/5xx response bodies as file contents
> [!NOTE]
> This issue was primarily written by GPT-5.6 Sol xhigh in Codex CLI.

## Summary

In grammY v1.46.0, `InputFile` streams URL response bodies into Telegram uploads without checking the upstream HTTP status. A 404, 403, 429, or 5xx error page can therefore be delivered as the file, or produce a misleading Telegram media error.

This also affects `{ url }` sources and Deno's supported `Response` inputs.

## Minimal reproduction

Set `BOT_TOKEN` and `CHAT_ID` for a test bot and chat, then save this as `repro.ts`:

```ts
import {
  Api,
  InputFile,
} from "https://cdn.jsdelivr.net/gh/grammyjs/grammY@v1.46.0/src/mod.ts";

const api = new Api(Deno.env.get("BOT_TOKEN")!);
const chatId = Number(Deno.env.get("CHAT_ID"));
const server = Deno.serve(
  { hostname: "127.0.0.1", port: 8080 },
  () =>
    new Response(
      "<!doctype html><html><body>grammY fixture: upstream file unavailable</body></html>\n",
      { status: 404, headers: { "content-type": "text/html" } },
    ),
);

try {
  await api.sendDocument(
    chatId,
    new InputFile(new URL("http://127.0.0.1:8080/report.pdf")),
  );
} finally {
  await server.shutdown();
}
```

```sh
deno run --allow-import=cdn.jsdelivr.net,cdn.skypack.dev,jsr.io --allow-env=BOT_TOKEN,CHAT_ID --allow-net=127.0.0.1,api.telegram.org repro.ts
```

The call succeeds, and the delivered `report.pdf` contains the HTML error page. This reproduction sends a real Telegram message.

## Expected behavior

Reject sources for which `response.ok` is false with a generic `Error` whose message includes the HTTP status, before consuming the response body as file contents. `response.ok` is true for HTTP 200–299. Successful responses should retain the existing streaming and missing-body checks, including the current failure for a bodyless 204 response.

## Actual behavior

Live testing with grammY v1.46.0 and Deno 2.9.6 produced these results for the 83-byte HTTP 404 HTML fixture:

| Method          | Result                                                                                               |
| --------------- | ---------------------------------------------------------------------------------------------------- |
| `sendDocument`  | Succeeds; downloaded document contains the exact HTML bytes                                          |
| `sendPhoto`     | Rejects with `400 IMAGE_PROCESS_FAILED`                                                              |
| `sendVideo`     | Succeeds with a `video` field; downloaded `.mp4` contains the exact HTML bytes                       |
| `sendAnimation` | Succeeds with a `video` field for this `.mp4` fixture; downloaded file contains the exact HTML bytes |

Document, video, and animation uploads were each tested with URL and Deno `Response` inputs. For the video/animation cases, use `sendVideo` or `sendAnimation` with the same source and a `.mp4` filename. For the Response path, use `new InputFile(await fetch(url), filename)`.

All six accepted error-page files were downloaded using Telegram `getFile`; their bytes and SHA-256 hashes matched the source HTML exactly. Valid HTTP 200 document and photo controls also succeeded. No valid video/animation control was tested; the returned `video` field describes these particular `.mp4` fixtures.

Local tests also confirmed unchanged error-page uploads for HTTP 403, 404, 429, 500, and 503 through URL, `{ url }`, and Deno `Response` sources.

## Suspected cause

The URL helpers use `response.body` without inspecting `response.status`. `fetch()` resolves normally for HTTP error statuses, so the error body is treated as a successful download.

Sources at the audited commit `055a5a4`, matching the tested v1.46.0 files:

- [Deno URL helper](https://github.com/grammyjs/grammY/blob/055a5a440f04d0b9fd5fd75a6d14dac4c2b83553/src/types.deno.ts#L178) and [Response branch](https://github.com/grammyjs/grammY/blob/055a5a440f04d0b9fd5fd75a6d14dac4c2b83553/src/types.deno.ts#L156).
- [Node URL helper](https://github.com/grammyjs/grammY/blob/055a5a440f04d0b9fd5fd75a6d14dac4c2b83553/src/types.node.ts#L166).
- [Web URL helper](https://github.com/grammyjs/grammY/blob/055a5a440f04d0b9fd5fd75a6d14dac4c2b83553/src/types.web.ts#L136).

Node and web findings are based on source inspection; runtime testing used Deno. Neither Node nor web currently accepts `Response` inputs.

`response.ok` is implemented across all three fetch surfaces: Deno's [native Response API](https://docs.deno.com/api/web/fetch/#Response), the web [Fetch Standard Response](https://fetch.spec.whatwg.org/#dom-response-ok), and Node's `node-fetch` dependency ([v2.7.0 getter](https://github.com/node-fetch/node-fetch/blob/v2.7.0/src/response.js#L59-L61)). Each defines success as HTTP 200–299, so the same `!response.ok` criterion can be used across the runtime implementations.

A minimal fix would check `if (!response.ok)` in all three URL helpers and Deno's existing Response branch, release rejected bodies where possible, and throw a plain `Error`. Keep streaming and multipart ordering; no new error class, eager preparation, MIME checks, or format sniffing is needed. Focused regression tests should cover unsuccessful responses, unchanged successful streams, and the 2xx boundaries (including 299 versus 300). Redirects should continue to follow the existing fetch behavior; check the final response's `ok` value. A redirect that ends at an HTTP 200 login page still passes this status-only check.

The existing lazy streaming error path wraps source errors as `HttpError`, with the original error available in `.error`. This proposal preserves that path and does not require rejection before the Telegram fetch starts. One caveat is that auto-retry retries `HttpError` by default; retry-policy changes can be considered separately.

Original reporter origin: [report in the grammY Telegram group](https://t.me/grammyjs/381047).

KnightNiwrem:
> [!NOTE]
> The fix was written by Claude Fable 5.1 high on Claude Code CLI.
> The PR body and fix verification were done by GPT-5.6 Sol xhigh on Codex CLI.

## Summary

Reject unsuccessful responses in `InputFile` downloads using `!response.ok`, before their bodies can be uploaded as file contents. Only HTTP 200–299 responses pass this check.

Fixes #971.

This change:

- checks `response.ok` in the Deno, Node, and web URL helpers, and in Deno's existing `Response` input branch
- aborts unsuccessful URL downloads immediately
- rejects unsuccessful supplied `Response` objects without consuming or cancelling their bodies; callers remain responsible for consuming or cancelling those bodies
- throws a plain source `Error` containing the HTTP status through the existing lazy upload error path
- adds regression coverage for rejected 300/404/599 URL responses, upstream signal abortion, accepted 299 streams, and rejected 404 `Response` inputs

## Why

`fetch()` resolves normally for HTTP error responses. Previously, a URL returning 404 or 503 could therefore produce a successfully delivered document containing the HTML error page, or a misleading Telegram media error such as `IMAGE_PROCESS_FAILED`.

Checking `response.ok` consistently requires a successful HTTP response across all three fetch implementations. URL downloads retain the existing redirect behavior and check the final response's `ok` value. A redirect ending at an HTTP 200 login page still passes this check.

Aborting a URL fetch started by grammY releases the connection without draining a potentially huge or endless unsuccessful response body. For a supplied `Response`, grammY rejects the unsuccessful status and leaves the body untouched. The caller decides whether to consume or cancel that body.

Successful responses retain streaming and the existing missing-body checks, including the failure for a bodyless 204 response.

## Verification

The regression tests in `test/types.test.ts` cover the updated success criterion:

- URL responses with statuses 300, 404, and 599 reject with an error containing the HTTP status, and the signal passed to `fetch()` is aborted.
- A URL response with status 299 streams its bytes unchanged.
- A Deno `Response` with status 404 rejects with an error containing the HTTP status.

The current patch is at commit `ce623338c0bdfba9acbbbe7e1bddbcf8d4f82c81`; its source is available through [pinned jsDelivr imports](https://cdn.jsdelivr.net/gh/KnightNiwrem/grammY@ce623338c0bdfba9acbbbe7e1bddbcf8d4f82c81/src/mod.ts).

Earlier independent sandbox verification on Deno 2.9.6 exercised commit `5aa566824a4179a493bfb6177d700e4093d7d307`, before the updates to `response.ok` and handling of supplied `Response` bodies:

- HTTP 4xx/5xx failures through URL, `{ url }`, and Deno `Response` inputs rejected with `HttpError.error` containing the original plain status error. Multipart inspection confirmed zero file-body bytes were uploaded.
- Endless 404 URL downloads triggered server-side cancellation after one 1 KiB chunk.
- All 19 live Telegram checks passed: fifteen 404/503 error-source uploads aborted, and four valid HTTP 200 controls succeeded. Three control documents were downloaded and matched byte for byte and by SHA-256; the valid photo was delivered at 64 × 64.
- Sandbox formatting, type checks, and lint passed.

Those sandbox results apply to the earlier commit; the current patch has not been rerun in that sandbox. They do not establish cancellation of supplied `Response` bodies under the current patch, which leaves that responsibility to the caller. Runtime verification there covered Deno and the web `InputFile` implementation running in Deno, rather than actual Node or browser runtimes. Valid video/animation controls were not tested.

The fix preserves lazy multipart ordering: the Telegram fetch may start before an unsuccessful source is rejected. The existing `HttpError` wrapper also means auto-retry v2.0.2 retries these failures by default; the earlier sandbox run confirmed that `rethrowHttpErrors: true` propagates them immediately. Retry-policy changes are outside this fix.


MobinAskari (commented):
I tested out various URLs in different scenarios in my own project and it behaved exactly as I expected it to.

The fetch status check function implementation is kinda inconsistent. We have `isErrorStatus` in .deno.ts but we have a function in place when it comes to .node.ts and .web.ts. Not sure if it even matters since they do the same thing but I just wanted to point it out.

Also I used Node and installed it through npm+git, so I have no idea if it's behaving correctly on Deno, again it'll probably work just fine, I just wanted to notify.

This is more of a question for myself as I'm not really familiar with ins and outs of AbortController, so apologies for my ignorance, but isn't it used to cancel the request mid flight? if the request is already completed and we have the status, why are we calling abort at that point?

KnightNiwrem:
> I tested out various URLs in different scenarios in my own project and it behaved exactly as I expected it to.
> 
> The fetch status check function implementation is kinda inconsistent. We have `isErrorStatus` in .deno.ts but we have a function in place when it comes to .node.ts and .web.ts. Not sure if it even matters since they do the same thing but I just wanted to point it out.
> 
> Also I used Node and installed it through npm+git, so I have no idea if it's behaving correctly on Deno, again it'll probably work just fine, I just wanted to notify.
> 
> This is more of a question for myself as I'm not really familiar with ins and outs of AbortController, so apologies for my ignorance, but isn't it used to cancel the request mid flight? if the request is already completed and we have the status, why are we calling abort at that point?

When fetch has resolved with a Response, it does not mean that the connection has closed yet, or that the body bytes have all been downloaded to client. If that was the case, there would be no such thing as streaming - since all the body bytes would be in memory already.

Aborting a request, therefore, isn't just used to cancel a request mid-flight, but also to close the connection against the server, and signal that we don't want or care about the body bytes any more so it shouldn't bother sending any more, or tracking what is left to send.

MobinAskari (approved):
👍

